PR #2, opened 2025-12-11
First version of the OSI spec

sfc-gh-kbhatia:
Replaced with this PR 
https://github.com/open-semantic-interchange/OSI/pull/3

### DIFF
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -56,45 +56,16 @@ semantic_model:
 
     # Required: Collection of logical datasets(fact and dimension tables)
     # See Logical Dataset section below for detailed structure
-    logical_datasets:
-      - name: string
-        base_table: string
-        primary_key: string
-        unique_keys: []
-        description: string
-        fields: []
-        ai_context: {
-          string: string
-        }
-        custom_extensions:
-          - vendor_name: string
-            data: string
+    datasets: []
 
-    # Required: Defines how logical datasets or semantic models are connected
-    # Represents foreign key relationships between datasets
-    relationships:
-      - name: string
-        from: string
-        to: string
-        from_column: string
-        to_column: string
-        custom_extensions:
-          - vendor_name: string
-            data: string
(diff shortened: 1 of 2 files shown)


core-spec/spec.yaml
```diff
@@ -0,0 +1,222 @@
...
+  # Required: Unique identifier for the semantic model
+  - name: string
+
+    # Optional: Human-readable description of the semantic model
+    description: string
+
+    # Optional: Additional context for AI tools (e.g., custom prompts, instructions)
+    ai_context: string
+
+    # Required: Collection of logical datasets(fact and dimension tables)
```
hahmadi, line 57:
```suggestion
    # Required: Collection of logical datasets (fact and dimension tables)
```

core-spec/spec.yaml
```diff
@@ -0,0 +1,222 @@
...
+    # See Logical Dataset section below for detailed structure
+    datasets: []
+
+    # Optional: Defines how logical datasets are connected
+    # See Relationships section below for detailed structure
+    relationships:[]
+
+    # Optional: 
+    # These metrics can span one or more  logical datasets and use relationships
```
hahmadi, line 66:
```suggestion
    # These metrics can span one or more logical datasets and use relationships
```

core-spec/spec.yaml
```diff
@@ -0,0 +1,222 @@
...
+    # Allows vendors to add custom metadata without breaking core compatibility
+    custom_extensions:
+      - vendor_name: string  # Must be one of the values from 'vendors' enum above
+        data: string
+
+---
+# Logical Dataset Schema
+# Represents business entities or concepts (fact and dimension tables)
+# Fields are defined within the scope of a logical dataset
+datasets:
```
hahmadi, line 80: I'm wondering if additional information needs to be captured here to identify which system/deployment/region etc the dataset maps to. (something like a connection spec). Have you considered that?

PR #3, opened 2025-12-11
First version of the OSI spec 

khush-bhatia:
First version of the OSI spec. It covers the following:
  - Core semantic model structure
  - Support for datasets, relationships, fields, and metrics
  - Multi-dialect metric expressions
  - Vendor extensibility framework
  - Context for agents

### DIFF
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -38,7 +38,7 @@ semantic_model:
     # Optional: Additional context for AI tools (e.g., custom prompts, instructions)
     ai_context: string
 
-    # Required: Collection of logical datasets(fact and dimension tables)
+    # Required: Collection of logical datasets (fact and dimension tables)
     # See Logical Dataset section below for detailed structure
     datasets: []
 
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -73,7 +73,10 @@ datasets:
     # Can be a single column or a composite of multiple columns
     # This is the preferred unique identifier for this dataset and is used in relationships to determine many-to-one or one-to-one.
     # Examples:
+    # primary_key:
     #   - [customer_id]           # Simple primary key
+    #
+    # primary_key:
     #   - [order_id, line_number] # Composite primary key
     primary_key: []  # Array of column names (single or composite)
 
@@ -81,8 +84,13 @@ datasets:
     # Each unique key can be a single column or a composite of multiple columns
     # Used for determining relationship type of either many-to-one or one-to-one
     # Examples:
-    #   - [column1]              # Simple key on one column
-    #   - [column2, column3]     # Composite key on two columns
+    # unique_keys:
+    #   - [column1]
+    #   - [column2, column3]
+    #
+    # unique_keys:
+    #   - [column1, column2]      
+    #   - [column3, column4]     
     unique_keys:
       - []  # Array of column names (single or composite)
 
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -15,6 +15,7 @@ dialects:
   - "ANSI_SQL"              # Standard SQL dialect
   - "SNOWFLAKE"             # Snowflake
   - "MDX"                   # Multi-Dimensional Expressions
+  - "TABLEAU"               # Tableau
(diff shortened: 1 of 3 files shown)


core-spec/spec.yaml
```diff
@@ -0,0 +1,207 @@
...
+    # Required: Collection of logical datasets(fact and dimension tables)
+    # See Logical Dataset section below for detailed structure
+    datasets: []
+
+    # Optional: Defines how logical datasets are connected
+    # See Relationships section below for detailed structure
+    relationships:[]
+
+    # Optional: 
+    # These metrics can span one or more  logical datasets and use relationships
```
hahmadi:
```suggestion
    # These metrics can span one or more logical datasets and use relationships
```

core-spec/spec.yaml
```diff
@@ -0,0 +1,207 @@
...
+  # Required: Unique identifier for the semantic model
+  - name: string
+
+    # Optional: Human-readable description of the semantic model
+    description: string
+
+    # Optional: Additional context for AI tools (e.g., custom prompts, instructions)
+    ai_context: string
+
+    # Required: Collection of logical datasets(fact and dimension tables)
```
hahmadi:
```suggestion
    # Required: Collection of logical datasets (fact and dimension tables)
```

core-spec/spec.yaml
```diff
@@ -0,0 +1,207 @@
...
+    primary_key: []  # Array of column names (single or composite)
+
+    # Optional: Array of unique key definitions that uniquely identify rows in this dataset
+    # Each unique key can be a single column or a composite of multiple columns
+    # Used for determining relationship type of either many-to-one or one-to-one
+    # Examples:
+    #   - [column1]              # Simple key on one column
+    #   - [column2, column3]     # Composite key on two columns
+    unique_keys:
+      - []  # Array of column names (single or composite)
```
hahmadi, line 95: Shouldn't this be an array of array of column names?

core-spec/spec.yaml
```diff
@@ -0,0 +1,207 @@
...
+    #   - [customer_id]           # Simple primary key
+    #   - [order_id, line_number] # Composite primary key
+    primary_key: []  # Array of column names (single or composite)
+
+    # Optional: Array of unique key definitions that uniquely identify rows in this dataset
+    # Each unique key can be a single column or a composite of multiple columns
+    # Used for determining relationship type of either many-to-one or one-to-one
+    # Examples:
+    #   - [column1]              # Simple key on one column
+    #   - [column2, column3]     # Composite key on two columns
```
hahmadi: Can you add an example with two unique keys, each being composite?

core-spec/spec.yaml
khush-bhatia: It is. The first array is indicated by the `-`. I know yaml is confusing that way.

hahmadi: I understand. But the comment on line 80 as well as examples in the comment block need an update to clarify this.

tlento:
I didn't get space to read through this prior to merge but my one main comment is this spec is under-specified. 

For now it's not a big deal but there are a lot of opaque string and functionally untyped array types in here, which puts an enormous burden on all developers of converters to enumerate the possibilities of what those might contain.

We will need to go through and refine these before the spec is ready for publication or even for vendors to start building against it. If we are still going to use something like protobuf we'll need to resolve a lot of these issues in order to make that work well anyway, so I think that might be a good next step.

khush-bhatia:
> I didn't get space to read through this prior to merge but my one main comment is this spec is under-specified.
> 
> For now it's not a big deal but there are a lot of opaque string and functionally untyped array types in here, which puts an enormous burden on all developers of converters to enumerate the possibilities of what those might contain.
> 
> We will need to go through and refine these before the spec is ready for publication or even for vendors to start building against it. If we are still going to use something like protobuf we'll need to resolve a lot of these issues in order to make that work well anyway, so I think that might be a good next step.

I missed this comment before. Synced offline with Tom. This spec is a starting point, we will gradually move to a more typed spec as it evolves and yes we can maintain the spec in both proto and yaml format .

PR #26, opened 2026-01-16
Add Databricks to the list of vendors

christianeu-db:
- Add Databricks SQL as a dialect
- Add Databricks to the list of vendors with custom extensions
- Add an example (default catalog/schema)


khush-bhatia (commented):
Could you update the spec.yaml as well with same changes ?

christianeu-db:
> Could you update the spec.yaml as well with same changes ?

Thank you for the review - I've updated spec.yaml

khush-bhatia (approved):

PR #54, opened 2026-01-30
Add JSON Schema and Python validator

ralfbecher:
## Summary
- Add `core-spec/osi-schema.json` - JSON Schema (draft 2020-12) for validating OSI semantic model YAML files
- Add `validation/validate.py` - Python validator with extended checks beyond JSON Schema

## JSON Schema Features
- Validates structure, types, and enum values (Dialect, Vendor)
- Uses `additionalProperties: false` for strict validation
- References enum definitions for top-level `dialects`/`vendors` arrays

## Python Validator Features
Checks that JSON Schema cannot enforce:
- Unique dataset names per model
- Unique field names per dataset
- Unique metric/relationship names per model
- Valid relationship references (from/to datasets must exist)
- SQL syntax validation via sqlglot (ANSI_SQL, SNOWFLAKE, DATABRICKS)

## Usage
```bash
pip install pyyaml jsonschema sqlglot
python validation/validate.py examples/tpcds_semantic_model.yaml
```

For VS Code validation, add to YAML files:
```yaml
# yaml-language-server: $schema=../core-spec/osi-schema.json
```

## Test plan
- [x] JSON Schema validates against draft 2020-12 spec
- [x] Example file (tpcds_semantic_model.yaml) passes validation
- [x] Invalid inputs correctly rejected (tested duplicate names, invalid SQL, bad references)

### DIFF
--- a/core-spec/osi-schema.json
+++ b/core-spec/osi-schema.json
@@ -1,0 +1,339 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/osi-initiative/osi/core-spec/osi-schema.json",
+  "title": "OSI Core Metadata Specification",
+  "description": "JSON Schema for validating OSI (Open Semantic Interoperability) semantic model definitions",
+  "type": "object",
+  "properties": {
+    "dialects": {
+      "type": "array",
+      "description": "Supported expression language dialects (enumeration definition)",
+      "items": {
+        "$ref": "#/$defs/Dialect"
+      }
+    },
+    "vendors": {
+      "type": "array",
+      "description": "Supported vendors for custom extensions (enumeration definition)",
+      "items": {
+        "$ref": "#/$defs/Vendor"
+      }
+    },
+    "semantic_model": {
+      "type": "array",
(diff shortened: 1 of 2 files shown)


core-spec/osi-schema.json
```diff
@@ -0,0 +1,339 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/osi-initiative/osi/core-spec/osi-schema.json",
```
khush-bhatia: Could you please change the ID here to to map to the correct github URL

core-spec/osi-schema.json
```diff
@@ -0,0 +1,339 @@
...
+  "title": "OSI Core Metadata Specification",
```
khush-bhatia, line 4: Can you please add Version ? I realize it is missing in the yaml spec as well, I will add it there.

ralfbecher: Added version property

core-spec/osi-schema.json
ralfbecher: Updated ID

khush-bhatia (approved):

PR #81, opened 2026-03-06
Add verified queries (draft)

rlittle08:
Extends the spec for new section - `verified_queries`.

**What will this be used for?**
- For humans: documenting approved query patterns within a semantic model
- For AI: In tools like Cortex Analyst in Snowflake, will be directly used for query generation by AI tools

**Why does it warrant inclusion in this spec?**
- For both use cases listed above, example queries are universally helpful
- Having a namespace defined in the standard supports interoperable use, and allows data vendors to share proper query patterns with multiple downstream applications (AI or otherwise), in one place

### DIFF
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -16,7 +16,8 @@
 4. [Relationships](#relationships)
 5. [Fields](#fields)
 6. [Metrics](#metrics)
-7. [Examples](#examples)
+7. [Verified Queries](#verified-queries)
+8. [Examples](#examples)
 
 ---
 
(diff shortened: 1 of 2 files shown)


GabrielBarberini:
+1 from a downstream consumer migrating a proprietary cookbook (~25 named SQL templates). Two gaps:

**1. Typed parameters.** Without them, `verified_queries` is fixed SQL, not a reusable template. Suggested:

```yaml
- name: top_n_by_category
  question: "Top N items for {category}?"
  parameters:
    - name: category
      datatype: string          # leans on https://github.com/open-semantic-interchange/OSI/pull/113
      description: "category code"
      sample_values: ["A", "B", "C"]
    - name: days
      datatype: integer
      default: 30
    - name: n_limit
      datatype: integer
      default: 10
  expression:
    dialects:
      - dialect: ANSI_SQL
        expression: |
          SELECT ... LIMIT {n_limit}
```

**2. Placeholder convention.** Pin one (`{name}` / `:name` / `${name}`) so consumers can substitute deterministically.

GabrielBarberini:
Sharing here to give emphasis https://claude.com/blog/how-anthropic-enables-self-service-data-analytics-with-claude

```
Query corpus: historical SQL from dashboards, notebooks, and prior analyses. Intuitively, this should be high-value: it's a record of every question already answered correctly. In practice, we found that giving the agent raw retrieval access to thousands of prior queries moved accuracy by less than a point (we walk through that ablation in a later section below). Unstructured retrieval couldn't map a new question to the right precedent. What does work is distilling that corpus into structured per-domain reference docs and reusable analysis patterns described in skills. Treat the query history as raw material for curation, not as a source of truth the agent reads directly.
```

I guess by now everyone seem to agree having a canonic (question, SQL) corpus improves analytics models performance.

PR #90, opened 2026-03-20
Small syntax bugfix for `spec.yaml`

krisajenkins:
The `semantic_model.relationships` entry didn't have a space after the `:`. This is technically mandatory, and some parsers can't read the file without it.

### DIFF
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -46,7 +46,7 @@ semantic_model:
 
     # Optional: Defines how logical datasets are connected
     # See Relationships section below for detailed structure
-    relationships:[]
+    relationships: []
 
     # Optional: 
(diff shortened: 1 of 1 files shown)


khush-bhatia (approved):

khush-bhatia:
@jbonofre Could you review this as well ?

jbonofre:
@khush-bhatia absolutely. On it ! Thanks !

jbonofre (approved):
LGTM

Per the YAML spec, a space is required after the colon in a mapping.
`relationships:[]` is technically invalid YAML (some parsers treat it as the scalar string "relationships:[]" rather than a key with an empty list value). 

This one-character fix is  correct and necessary for strict YAML parsers.

Thanks !

PR #91, opened 2026-03-21
Change vendor_name from enum to free-form string

jochenchrist:
## Summary
- Change `vendor_name` in custom extensions from a closed enum to a free-form string, allowing any vendor or organization to define custom extensions without modifying the core spec
- Existing vendor names (COMMON, SNOWFLAKE, SALESFORCE, DBT, DATABRICKS) are preserved as `examples` in the JSON Schema and documented as well-known examples in the spec
- Updated `spec.yaml`, `osi-schema.json`, and `spec.md`

### DIFF
--- a/core-spec/osi-schema.json
+++ b/core-spec/osi-schema.json
@@ -19,7 +19,7 @@
     },
     "vendors": {
       "type": "array",
-      "description": "Supported vendors for custom extensions (enumeration definition)",
+      "description": "Well-known vendors for custom extensions (non-exhaustive list of examples)",
       "items": {
         "$ref": "#/$defs/Vendor"
       }
--- a/core-spec/osi-schema.json
+++ b/core-spec/osi-schema.json
@@ -42,8 +42,8 @@
     },
     "Vendor": {
       "type": "string",
-      "enum": ["COMMON", "SNOWFLAKE", "SALESFORCE", "DBT", "DATABRICKS", "GOODDATA"],
-      "description": "Supported vendors for custom extensions"
+      "examples": ["COMMON", "SNOWFLAKE", "SALESFORCE", "DBT", "DATABRICKS", "GOODDATA"],
+      "description": "Vendor name for custom extensions. Any string value is accepted."
     },
     "AIContext": {
       "description": "Additional context for AI tools",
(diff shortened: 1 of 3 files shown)


jbonofre (commented):
`osi-schema.json` should be updated on line 20-26 as it still references the array.
It also still mentions `Supported vendors for custom extensions (enumeration definition)` which is not accurate anymore.

In the `spec.md`, the custom extensions section (line 359) still show:
```
  custom_extensions:                                                                                                                                                                                                                                                           
    - vendor_name: string  # Must be from vendors enum 
```
which is not accurate with the change.
Also `CustomExtension` in schema still references `$ref: "#/$defs/Vendor". While the `Vendor` def itself is updated, the overall `vendors` top-level property and its description remain inconsistent.

If I understand the intent (relaxing vendors), and it makes to me, I suggest to update the vendors ref across the three files to be consistent.

Can you please do a complete pass to update all enum references ?

jochenchrist:
Thanks for the review! I've addressed the feedback:

- Updated the `vendors` array description in `osi-schema.json` from "enumeration definition" to a functional description
- Fixed the `spec.md` custom extensions schema comment to say "Free-form string identifying the vendor" instead of "Must be from vendors enum"

All remaining enum references should now be consistent with the change to a free-form string.

jbonofre:
@jochenchrist Many thanks ! I'm doing a new pass.

khush-bhatia:
The change makes sense, but would end up being a breaking change. 

Let's use git tag for versioning and then it would still be nice to have static shareable links for the different versions of the spec via build automation tooling ? @jbonofre  what do you think ?

jbonofre:
@khush-bhatia I agree. We have `version: 0.1.1` in the spec. I would say that any breaking change on the spec (like this one) should bump the version (according to semver).
We have branches per version, however, I would recommend to use only two active spec versions (so branches) because the maintenance is "burden" 😄 

I can work on the "CI" to verify and manage the versions.

Thoughts ?

khush-bhatia:
@jochenchrist Thanks for this PR. We have done the release work. And now this PR can be merged to main. Do you mind rebasing this PR to the latest main ?

jochenchrist:
Thanks @khush-bhatia! Rebased onto latest main. Resolved the conflict in osi-schema.json / spec.yaml by keeping GOODDATA in the vendor examples list. Ready to merge.

khush-bhatia (approved):

PR #93, opened 2026-03-23
Add Apache Spark converter module

jbonofre:
## Summary
- Adds a new `converters/spark/` Java/Maven module that reads OSI semantic model YAML files and generates PySpark code
- Generated code includes dataset loaders (`spark.table()` + computed columns via `F.expr()`), join helpers from relationship definitions, metric aggregation functions via `spark.sql()`, and a convenience `load_all_datasets()` function
- Supports multi-dialect expression selection (ANSI_SQL, SNOWFLAKE, DATABRICKS) with fallback

## Test plan
- [x] Unit tests pass (`mvn clean test` — 4 tests covering parsing, field extraction, code generation, and multi-dialect support)
- [x] End-to-end validation against `examples/tpcds_semantic_model.yaml` produces valid PySpark output
- [ ] Review generated PySpark code for correctness on an actual Spark cluster

### DIFF
--- a/converters/spark/src/main/java/org/osi/converter/spark/SparkCodeGenerator.java
+++ b/converters/spark/src/main/java/org/osi/converter/spark/SparkCodeGenerator.java
@@ -1,0 +1,246 @@
+package org.osi.converter.spark;
+
+import org.osi.converter.spark.model.OsiModel;
+import org.osi.converter.spark.model.OsiModel.*;
+
+import java.util.List;
+
+/**
+ * Generates PySpark code from a parsed {@link OsiModel}.
+ * <p>
+ * The generated code creates DataFrames, registers temp views,
+ * builds join helpers from relationships, and exposes metric functions.
+ */
+public class SparkCodeGenerator {
+
+    private final String dialect;
+
+    public SparkCodeGenerator() {
+        this("ANSI_SQL");
+    }
+
+    public SparkCodeGenerator(String dialect) {
+        this.dialect = dialect;
+    }
+
+    /**
+     * Generate the full PySpark module for the given OSI model.
+     */
+    public String generate(OsiModel model) {
+        StringBuilder sb = new StringBuilder();
+
+        for (SemanticModel sm : model.getSemanticModels()) {
+            generateHeader(sb, sm);
+            generateDatasetLoaders(sb, sm);
+            generateLoadAll(sb, sm);
+            generateJoinHelpers(sb, sm);
+            generateMetrics(sb, sm);
+            generateMain(sb, sm);
+        }
+
+        return sb.toString();
+    }
+
+    // -----------------------------------------------------------------------
+    // Header
+    // -----------------------------------------------------------------------
+
+    private void generateHeader(StringBuilder sb, SemanticModel sm) {
+        sb.append("\"\"\"\n");
+        sb.append("Auto-generated PySpark code from OSI semantic model: ").append(sm.getName()).append("\n");
+        if (sm.getDescription() != null) {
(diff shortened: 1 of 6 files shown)


jbonofre:
A very simple converter for Apache Spark, showing the kind of code we can have in the repo.

converters/spark/src/main/java/org/osi/converter/spark/model/OsiModel.java
```diff
@@ -0,0 +1,155 @@
+package org.osi.converter.spark.model;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Java representation of an OSI semantic model parsed from YAML.
+ */
+public class OsiModel {
```
khush-bhatia, line 10: The java representation code seems to be in two places in both spark and polaris modules.  How about we instead follow this directory structure 


core-spec/
│   ├── spec.yaml
│   ├── spec.json
│   └── spec.proto

And then use the proto version of the spec to dynamically generate the java/python/go bindings for the spec ? 

I think dbt was also suggesting doing that. Many applications can also directly consume the proto version of the spec. 
                
For the purpose for this PR, I think it is fine to add the java bindings, I would still keep the java bindings in one folder

converters/spark/src/main/java/org/osi/converter/spark/model/OsiModel.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private String version;
+    private List<SemanticModel> semanticModels = new ArrayList<>();
```
khush-bhatia, line 13: Also there is no parent OSI model that has list of semantic_models. 

The OSI spec represents the spec for one semantic_model.

converters/spark/src/main/java/org/osi/converter/spark/SparkCodeGenerator.java
```diff
@@ -0,0 +1,246 @@
...
+        sb.append("\ndef ").append(funcName).append("(spark: SparkSession) -> DataFrame:\n");
+        sb.append("    \"\"\"\n");
+        sb.append("    Metric: ").append(metric.getName()).append("\n");
+        if (metric.getDescription() != null) {
+            sb.append("    ").append(metric.getDescription()).append("\n");
+        }
+        sb.append("    Expression: ").append(expr).append("\n");
+        sb.append("    \"\"\"\n");
+        sb.append("    return spark.sql(\"SELECT ").append(escapeString(expr))
+                .append(" AS ").append(metric.getName()).append("\")\n\n");
```
khush-bhatia, line 195: Hey @jbonofre As far as I know spark does not have native support for querying a semantic model. 

And this code is generating some SQL , but I don't think that query will just work in spark. We need to align on the query semantics on OSI and then have a query rewriter implementation in spark. 

I suggest we keep CodeGen out of scope for now, and instead have the logic to load the semantic model to the spark catalog ?

khush-bhatia (commented):
Left some comments, thanks

converters/spark/src/main/java/org/osi/converter/spark/model/OsiModel.java
jbonofre: As the other converters are directly in the `converters` module (Python based for instance).

I'm happy to do a previous PR to reorg the modules with multi-languages binding.

converters/spark/src/main/java/org/osi/converter/spark/SparkCodeGenerator.java
jbonofre: That's reasonable. I will update accordingly.

PR #94, opened 2026-03-23
Add Apache Polaris converter module

jbonofre:
## Summary

- Adds a two-way converter between OSI semantic models and Apache Polaris catalogs under `converters/polaris/`
- **Import mode**: reads Polaris namespaces/tables via the Iceberg REST Catalog API and generates OSI YAML, mapping Iceberg schemas to OSI datasets, fields, primary keys, and temporal dimensions
- **Export mode**: reads OSI YAML and creates Polaris namespaces and Iceberg tables with type inference from field names and dimension metadata

## Test plan

- [x] Unit tests pass (9 tests covering parsing, YAML generation round-trip, Iceberg schema generation, type inference, and edge cases)
- [ ] Integration test against a running Polaris instance (import and export)
- [ ] Verify round-trip fidelity: Polaris → OSI → Polaris

### DIFF
--- a/converters/polaris/src/main/java/org/osi/converter/polaris/model/OsiModel.java
+++ b/converters/polaris/src/main/java/org/osi/converter/polaris/model/OsiModel.java
@@ -1,0 +1,180 @@
+package org.osi.converter.polaris.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Java representation of an OSI semantic model parsed from YAML.
+ */
+public class OsiModel {
+
+    private String version;
+    private List<SemanticModel> semanticModels = new ArrayList<>();
+
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
+    }
+
+    public List<SemanticModel> getSemanticModels() {
+        return semanticModels;
+    }
+
+    public void setSemanticModels(List<SemanticModel> semanticModels) {
+        this.semanticModels = semanticModels;
+    }
+
+    // -----------------------------------------------------------------------
+    // Nested model classes
+    // -----------------------------------------------------------------------
+
+    public static class SemanticModel {
+        private String name;
+        private String description;
+        private List<Dataset> datasets = new ArrayList<>();
+        private List<Relationship> relationships = new ArrayList<>();
+        private List<Metric> metrics = new ArrayList<>();
+
+        public String getName() { return name; }
+        public void setName(String name) { this.name = name; }
+
+        public String getDescription() { return description; }
+        public void setDescription(String description) { this.description = description; }
+
+        public List<Dataset> getDatasets() { return datasets; }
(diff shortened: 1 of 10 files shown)


jbonofre:
This is a use example:

```
  # Import: Polaris catalog → OSI YAML
  java -jar osi-polaris-converter.jar import \
    --url http://localhost:8181 --catalog my_catalog \                                                                                                                                                                                        
    --client-id <id> --client-secret <secret> \
    -o output.yaml                                                                                                                                                                                                                            
                                                            
  # Export: OSI YAML → Polaris catalog                                                                                                                                                                                                        
  java -jar osi-polaris-converter.jar export \              
    --url http://localhost:8181 --catalog my_catalog \
    --token <bearer-token> \                                                                                                                                                                                                                  
    model.yaml
```
                                                                                                                                                                                                                                              
 I implemented with the following logic:
 * Each Polaris namespace maps to an OSI semantic model                                                                                                                                                                                      
 * Each Iceberg table maps to an OSI dataset               
 * Iceberg schema fields map to OSI fields with ANSI_SQL expressions
 * Iceberg identifier-field-ids map to OSI primary_key                                                                                                                                                                                       
 * Temporal Iceberg types (timestamp, timestamptz, date) set dimension.is_time: true                                                                                                                                                         
 * Table properties are preserved in custom_extensions

converters/polaris/src/main/java/org/osi/converter/polaris/model/OsiModel.java
```diff
@@ -0,0 +1,180 @@
+package org.osi.converter.polaris.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Java representation of an OSI semantic model parsed from YAML.
+ */
+public class OsiModel {
```
flyrain, line 9: Thanks for adding this, @jbonofre.  Can we put it a common module instead of Polaris module? I think the  Java representation would be really useful for other catalogs as well.

jbonofre: @flyrain That's a good idea. I think it also related to a discussion we had before: what should be the "language layout" as we have converter in Python, Java, ...

Let me give a proposal.

flyrain: Awesome! Looking forward to it

jbonofre: I'm working on a PR to have `core-spec` module, that will create bindings for different languages (including java).

jbonofre:
@khush-bhatia I'm preparing a first step PR with `core-spec` and java binding.

khush-bhatia (approved):

sgandon (approved):
LGTM

jbonofre:
I created #121 for follow up.

PR #96, opened 2026-03-24
Add GoodData support: MAQL dialect, vendor registration, and bidirectional LDM converter

jaceksan:
GoodData joins the OSI initiative with full spec registration and a production-ready bidirectional convertor between our Logical Data Model and the OSI semantic model.

**What's included**

Spec changes — MAQL dialect and GOODDATA vendor added to schema, documentation, and validator (same treatment as MDX/TABLEAU).

Bidirectional convertor (convertors/gooddata/) — Python package with 28 tests:
- GoodData → OSI: Dual-dialect fields (ANSI_SQL + MAQL), relationships from references, date instances, GoodData-specific metadata via custom_extensions
- OSI → GoodData: Attribute/Fact classification from dimension metadata and MAQL patterns, date instance reconstruction, declarative API-ready output
- Round-trip verified: dataset IDs, attribute/fact counts, relationships, and date instances all survive GoodData → OSI → GoodData

**What's not included**

Metrics. MAQL is a context-aware metric language (like MDX) — dimensionality and filters are applied at report time, not baked into expressions. The current OSI metric model can't represent this. We'd like to collaborate on a vendor-agnostic context-aware metric standard that would unlock metric interchange for GoodData, MDX, dbt, and Tableau alike.


jbonofre (changes requested):
Do you mind to rename `convertors` to `converters` ?

Also, please don't include the "binary" `__pycache__` directories. It should be removed from the PR and included in the `.gitignore`.
I believe it's also better to not include `uv.lock`.

jaceksan:
Thanks for the review!

I've addressed points 1 and 2:
- Renamed `convertors` to `converters`
- Removed `__pycache__` directories from tracking and added `__pycache__/` to `.gitignore`

Regarding `uv.lock` — I'd prefer to keep it. The general best practice is to commit lock files for applications/tools (as opposed to libraries) to ensure reproducible builds. This is consistent with how other Python projects in our organization handle it. Happy to discuss further if you have concerns.

converters/gooddata/README.md
```diff
@@ -0,0 +1,61 @@
+# GoodData OSI Convertor
```
jbonofre:
```suggestion
# GoodData OSI Converter
```

converters/gooddata/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Bidirectional convertor between GoodData's declarative Logical Data Model (LDM)
```
jbonofre:
```suggestion
Bidirectional converter between GoodData's declarative Logical Data Model (LDM)
```

converters/gooddata/pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[project]
+name = "gooddata-osi"
+version = "0.1.0"
+description = "Bidirectional convertor between GoodData LDM and OSI semantic model"
```
jbonofre:
```suggestion
description = "Bidirectional converter between GoodData LDM and OSI semantic model"
```

converters/gooddata/src/gooddata_osi/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Bidirectional convertor between GoodData LDM and OSI semantic model."""
```
jbonofre:
```suggestion
"""Bidirectional converter between GoodData LDM and OSI semantic model."""
```

converters/gooddata/tests/conftest.py
```diff
@@ -0,0 +1,33 @@
+"""Shared pytest fixtures for gooddata-osi convertor tests."""
```
jbonofre:
```suggestion
"""Shared pytest fixtures for gooddata-osi converter tests."""
```

jaceksan:
@jbonofre, is anything else needed before the merge?

converters/gooddata/src/gooddata_osi/gooddata_to_osi.py
```diff
@@ -0,0 +1,232 @@
...
+    Returns:
+        OSI semantic model as a dict (ready to serialize to YAML).
+    """
+    osi_datasets = []
+    osi_relationships = []
+
+    # Map dataset id -> grain attribute ids (used as primary keys in OSI)
+    dataset_grain_map: dict[str, list[str]] = {}
+    for ds in model.ldm.datasets:
+        dataset_grain_map[ds.id] = [g.id for g in ds.grain]
```
khush-bhatia: This is populated but never read, was that intentional ?

converters/gooddata/src/gooddata_osi/gooddata_to_osi.py
```diff
@@ -0,0 +1,232 @@
...
+    source_columns: list[str],
+    multivalue: bool,
+) -> dict[str, Any]:
+    """Convert a GoodData reference to an OSI relationship."""
+    rel: dict[str, Any] = {
+        "name": f"{from_dataset}_to_{to_dataset}",
+        "from": from_dataset,
+        "to": to_dataset,
+        "from_columns": source_columns,
+        "to_columns": source_columns,
```
khush-bhatia: This does not seem correct. The from and to datasets are different but the columns are same ?

converters/gooddata/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+authors = [{name = "GoodData Corporation", email = "support@gooddata.com"}]
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "pyyaml>=6.0.1",
+    "jsonschema>=4.23.0",
+]
+
+[tool.uv]
+python-preference = "only-system"
```
khush-bhatia: This will be problematic as we add CI support. Can this be changed to managed or better yet omitted ?

core-spec/osi-schema.json
```diff
@@ -37,12 +37,12 @@
   "$defs": {
     "Dialect": {
       "type": "string",
-      "enum": ["ANSI_SQL", "SNOWFLAKE", "MDX", "TABLEAU", "DATABRICKS"],
+      "enum": ["ANSI_SQL", "SNOWFLAKE", "MDX", "TABLEAU", "DATABRICKS", "MAQL"],
```
khush-bhatia, line 40: Could you add this same change to the spec.yaml ?

converters/gooddata/src/gooddata_osi/gooddata_to_osi.py
jaceksan: Good catch — this map was indeed unused. It's now removed and replaced by an `attr_source_col_map` that the new `_convert_reference` actually uses to resolve target grain attribute IDs to physical source columns (the missing piece for the next comment).

converters/gooddata/src/gooddata_osi/gooddata_to_osi.py
jaceksan: You're right — this was wrong. The original code copied source columns to both sides, which only worked when the FK column and the target dataset's grain attribute ID happened to match by convention.

After cross-checking the GoodData backend (`libraries/metadata-model` `DeclarativeReference.kt`), the legacy `sourceColumns` field is marked deprecated and replaced by `sources[]`, where each source carries an explicit `target: GrainIdentifier`. We want OSI to track only the new format, so:

- `GdReference` now holds `sources: list[GdReferenceSource]` (column + target { id, type }); legacy `sourceColumns` input raises `ValueError`.
- `_convert_reference` resolves each source's `target.id` (an attribute id in the target dataset) to the physical column via the target's attribute id → `sourceColumn` map, so OSI `to_columns` now hold real join columns (e.g. `store_sales.ss_customer_sk` → `customer.c_customer_sk`).
- Date-instance targets (`type: date`) fall back to the source column since date dims have no physical grain columns.
- Fixtures updated to the `sources` shape.

converters/gooddata/pyproject.toml
jaceksan: Removed the `[tool.uv]` block entirely. uv's default (`managed`) will pick up project interpreters in CI without constraining to only-system.

core-spec/osi-schema.json
jaceksan: Done — added `MAQL` to `dialects` and `GOODDATA` to `vendors` in `core-spec/spec.yaml` so it matches `osi-schema.json`.

khush-bhatia (approved):
LGTM, thanks for the contribution.

jbonofre:
I'm doing a new pass.

converters/gooddata/pyproject.toml
```diff
@@ -0,0 +1,41 @@
+[project]
+name = "gooddata-osi"
+version = "0.1.0"
+description = "Bidirectional converter between GoodData LDM and OSI semantic model"
+authors = [{name = "GoodData Corporation", email = "support@gooddata.com"}]
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "pyyaml>=6.0.1",
+    "jsonschema>=4.23.0",
```
jbonofre: nit: I don't see any source importing jsonschema. I guess this is not necessary.

jbonofre (approved):
It looks good now. Thanks!

converters/gooddata/pyproject.toml
jaceksan: Good catch — removed in 6a06f34. It was added with the intent to validate OSI output against `core-spec/osi-schema.json` but never wired up. Dropping it for now keeps the dependency tree clean. I'll propose a shared `osi-validation` package on Slack so it can be adopted by all vendor converters consistently, rather than each re-implementing `jsonschema` glue.

PR #97, opened 2026-03-24
Add validation rules for semantic model names, relationship columns, extensions, and dialects

kevinwangz:
- Check for duplicate semantic model names across the file
- Validate that from_columns and to_columns have equal length in relationships
- Validate that custom extension 'data' fields contain valid JSON, checking all levels (model, dataset, field, metric, relationship)
- Detect duplicate dialect entries within a single expression's dialects array for both fields and metrics

### DIFF
--- a/validation/validate.py
+++ b/validation/validate.py
@@ -4,9 +4,10 @@ OSI Semantic Model Validator
 
 Validates OSI YAML files against:
 1. JSON Schema (structure, types, enums)
-2. Unique names (datasets, fields, metrics, relationships)
-3. Valid relationship references
-4. SQL syntax (using sqlglot)
+2. Unique names (semantic models, datasets, fields, metrics, relationships, dialects)
+3. Valid relationship references and column count consistency
+4. Custom extension JSON validity
+5. SQL syntax (using sqlglot)
 
 Usage:
     python validation/validate.py <yaml_file>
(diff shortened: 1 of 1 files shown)


khush-bhatia:
@kevinwangz Thanks for this PR. But I realized that the original validation script is missing tests. Do you mind bootstrapping the tests for this code ? Else, I am happy to add the tests for this.

validation/validate.py
```diff
@@ -116,6 +122,90 @@ def validate_references(data: dict) -> list[str]:
...
+            rel_name = rel.get("name", "<unnamed>")
+            extension_sources.append((f"relationship '{rel_name}'", rel.get("custom_extensions", [])))
+
+        for context, extensions in extension_sources:
+            for ext in extensions or []:
+                vendor = ext.get("vendor_name", "<unknown>")
+                ext_data = ext.get("data")
+                if ext_data is not None:
+                    try:
+                        json.loads(ext_data)
```
jbonofre, line 170: The `data` field from YAML will already be parsed as a Python object (`dist`/`list`/`int`), not a JSON string.

`json.loads()` expects a string. If someone writes:

```
custom_extensions:
  - vendor_name: acme
     data: '{"key": "value"}'              # string - works
  - vendor_name: acme
     data:                                            # YAML dict - TypeError from json.loads
         key: value
```

The `TypeError` is caught, but the error message would be misleading. It would say "invalid JSON" when really the data is valid structured data, just not a JSON string.

The validation should clarify what it actually expects: is `data` supposed to be JSON-encoded string, or a structured object ?

If either is acceptable, the check should only run `json.loads` when `ext_data` is a `str`.

jbonofre (commented):
Overall good, I just have a question about `json.loads` semantics.

PR #98, opened 2026-03-24
OSI->Snowflake YAML Converter Script

sfc-gh-tgao:
An initial python script to convert an OSI yaml to Snowflake yaml, still in development.


khush-bhatia:
@sfc-gh-tgao Thanks for this PR. Could you add tests for this script ? Also it would be good to have the code and the tests be in separate sub directories and the example output could live in the tests directory.

sfc-gh-tgao:
> @sfc-gh-tgao Thanks for this PR. Could you add tests for this script ? Also it would be good to have the code and the tests be in separate sub directories and the example output could live in the tests directory.

Thanks @khush-bhatia. Created sub-directories and added some unit tests.

converters/snowflake/src/osi_to_snowflake_yaml_converter.py
```diff
@@ -0,0 +1,444 @@
...
+    )
+
+
+def _extract_synonyms(ai_context):
+    """Extracts synonyms list from a structured ai_context object.
+
+    If ai_context is a dict with a 'synonyms' key, returns the synonyms list.
+    If ai_context is a plain string or None, returns None.
+    """
+    if isinstance(ai_context, dict):
```
jbonofre, line 388: The converter expects a `dict` here, but the OSI spec defines `ai_context` as `string`. 

Can we clarify here ? Are you proposing to update the spec ?

converters/snowflake/src/osi_to_snowflake_yaml_converter.py
```diff
@@ -0,0 +1,444 @@
...
+    # positives on table names like WITH_TABLE or SELECT_RESULTS.
+    upper = source_stripped.upper()
+    if upper.startswith(("SELECT ", "SELECT\n", "SELECT\t",
+                          "WITH ", "WITH\n", "WITH\t")):
+        return {"definition": source_stripped}
+
+    # TODO: Quoted identifiers (e.g., "my.db"."my schema"."my table") are not
+    # handled. Basic dot-splitting only.
+    parts = source_stripped.split(".")
+    if len(parts) == 3:
```
jbonofre, line 369: That's strictly correct regarding the OSI spec: the spec says `source` format should be `database_name.schema_name.table_name` or query (that you manage by testing if `source` starts with `SELECT`).

However, I believe some datasets may only have `schema.table`.

I would at a comment note about the "strict" three parts.

jbonofre (commented):
Overall good, I just have a question about `ai_context` (the difference between the spec and the converter).

converters/snowflake/src/osi_to_snowflake_yaml_converter.py
sfc-gh-tgao: Thanks, good catch. There is not a direct mapping for `ai_context` in snowflake yaml, so for now I update the code to append the string `ai_context` to `description`.
I will have follow up PRs to address those concepts that do not have direct mappings in snowflake, after discussing with the team.

converters/snowflake/src/osi_to_snowflake_yaml_converter.py
sfc-gh-tgao: Added a comment. We could relax this in the future.

khush-bhatia: We will solve this in the next version of the spec by allowing catalog type and catalog identifiers.

jbonofre (approved):
That's a solid contribution. Thanks !

I would maybe create an issue to follow up on `ai_context`.

khush-bhatia (approved):

khush-bhatia:
> That's a solid contribution. Thanks !
> 
> I would maybe create an issue to follow up on `ai_context`.

+1, we should track this issue for follow up.

sfc-gh-tgao:
> That's a solid contribution. Thanks !
> 
> I would maybe create an issue to follow up on `ai_context`.

Thanks. I created https://github.com/open-semantic-interchange/OSI/issues/111.

PR #113, opened 2026-04-24
Add datatype field to Field and Metric; reframe is_time as role marker

### LINKED ISSUE #84: Support field datatype rather than is_time
### Discussed in https://github.com/open-semantic-interchange/OSI/discussions/25

<div type='discussions-op-text'>

<sup>Originally posted by **jonmmease** January 16, 2026</sup>
I'd suggest that a logical `datatype` field be used in place of the proposed `is_time` field. For consumers of semantic models, the data type can be important when building logic on top of dimensions. A motivating example is a BI tool that supports its own expression language on top of dimensions to compute local derived fields or local measures for visualization. Having correct type information can be critical for working with dimensions properly.

I would suggest enumeration values of:
 - `number`
 - `string`
 - `boolean`
 - `time`
 - `timestamp`
 - `timestamptz`
 - `opaque`/`other` (for anything else)

For reference: LookML, Cube, and Hex require this information in their specs</div>

jonmmease:
**AI Disclosure:** I developed this PR and its description iteratively with Claude Code and Codex, and have reviewed the resulting changes fully.

## Motivation

Closes #84. Related to #17 and follows the earlier discussion in #25.

The issue originally framed this as “use `datatype` rather than `is_time`.” After reviewing existing semantic-layer implementations, the cleaner model is to represent logical data type and temporal role independently:
- `datatype` says what values a field or metric produces.
- `dimension.is_time` says whether a field plays a temporal-dimension role.

For example, `d_year`, `d_quarter_name`, and `d_month_name` may have integer data types while all serving temporal roles. Snowflake Semantic Views and LookML similarly separate scalar representation from temporal classification.

So, this PR adds `datatype` alongside `is_time`, then carries that metadata through the Python bindings and existing converter paths that have native type information.

### Core specification
Adds an optional top-level `datatype` property to both `Field` and `Metric`, backed by the following logical data types:

- `String`
- `Integer`
- `Decimal`
- `Float`
- `Boolean`
- `Date`
- `Time`
- `DateTime`
- `DateTimeTz`
- `Opaque`

The shared names align with the ontology specification’s built-in value types where possible. `Time`, `DateTimeTz`, and `Opaque` are additional types. I think these would make sense as ontology value types as well, but this PR doesn't modify the ontology’s layer.

The JSON Schema defines the authoritative `DataType` enum. Until the YAML specification is generated from JSON, `spec.yaml` mirrors the same enum explicitly.

### Temporal role semantics

`dimension.is_time` remains valid and is documented as independent of `datatype`.

For fields with a `dimension` block:

1. An explicit `is_time` value wins.
2. When `is_time` is unset, it defaults to `true` for `Date`, `Time`, `DateTime`, and `DateTimeTz`, and to `false` otherwise.

A field without dimension metadata remains a fact regardless of its datatype. This supports combinations like:
- `datatype: Integer` with `is_time: true` for a year grain.
- `datatype: DateTimeTz` with `is_time: false` for a timestamp that should not be exposed as a time dimension.
- `is_time: true` without `datatype` for legacy models that know the role but do not declare a scalar type.

### Python bindings

- Adds and exports `OSIDataType`.
- Adds optional `datatype` fields to `OSIField` and `OSIMetric`.
- Adds `OSIField.is_time_dimension()` to implement the effective temporal-role rule.
- Tests keep the Python enum synchronized with the core JSON Schema.

### Existing converters

I updated the existing converters to propagate data types where it looked unambiguous. Here tagging some folks who have worked on the various converters for visibility.

- @sfc-gh-tgao — Snowflake
- @jaceksan — GoodData
- @jbonofre — Polaris
- @deutschn — Salesforce


core-spec/spec.md
```diff
@@ -36,6 +36,22 @@ Supported SQL and expression language dialects for metrics and field definitions
...
 
+### Datatypes
+
+Logical data types for fields and metrics.
+
+| Datatype | Description |
+|----------|-------------|
+| `string` | Variable-length Unicode character data. |
+| `integer` | Signed integer with no scale. |
+| `number` | Real number (floating-point or decimal) with unspecified precision. |
```
KSDaemon: One concern here is that using a generic number type isn’t really a valid interoperability strategy. Different data warehouses interpret and implement it differently, so it’s not interchangeable in practice.

For example, you can’t take a model built against data in Snowflake and expect it to behave identically in Databricks just because both use a number type. As we’ve already discussed in other issues/threads, precision and scale vary across systems, and that directly affects results.

Without explicitly specifying those parameters, this approach is likely to introduce subtle bugs. If strict type checking is enforced and there’s no automatic conversion, things will simply break. Even with implicit or explicit casting at the physical layer, you can still end up with precision loss, which is not acceptable for many use cases.

So I think we need a different approach here. Either we make the type definition explicit (e.g., always specifying precision and scale for numeric types), or we introduce some abstraction that preserves correctness across backends instead of relying on a loosely defined number.

jonmmease: Thanks for the feedback @KSDaemon. The frame of reference I'm coming from is that these types need to be precise enough to generate SQL on top of (e.g. a BI tool building on top of the semantic definitions). In my experience, the SQL required for integers vs float/decimal can differ for warehouses that use truncation semantics for integers. I have not come across cases where I've needed to generate different SQL depending on whether the underlying type is a float or decimal.

But it is worth taking a step back to ask what else folks would rely on these logical data types for. Do you have any particular use cases in mind?

xavipereztr:
Hi @jonmmease ! 

Thanks for your work on this PR! Some comments on this PR based on comments and discussions we have opened in regards of including spatial semantics in the OSI spec:
- https://github.com/open-semantic-interchange/OSI/discussions/114
- https://github.com/open-semantic-interchange/OSI/issues/84 

**On `geometry` and `geography` as first-class datatype values**

We'd like to make the case for adding `geometry` and `geography` to the
`datatype` enum rather than leaving spatial columns to fall through to `other`.

Both are first-class native types in the leading cloud data warehouses:
- **Snowflake**: `GEOGRAPHY` and `GEOMETRY`
- **BigQuery**: `GEOGRAPHY` (WGS84 point/line/polygon)
- **Databricks**: `GEOMETRY` 
- **PostGIS**: `geometry` and `geography`

The `geometry` / `geography` distinction is also meaningful and standardized
(OGC/ISO SQL): `geography` operates on a spherical earth model with real-world
units, while `geometry` operates on a flat Cartesian plane. Consumers need to
know which they're dealing with to generate correct spatial SQL.

Leaving these as `other` + `custom_extensions` means every OSI implementation
dealing with spatial data falls back to vendor-specific metadata, something that at CARTO we believe it should not happen.

**On `dimension.spatial` as the spatial role marker**

The type/role split you've introduced gives spatial data a natural place to
live. We're working on a proposal (related to [#69](https://github.com/open-semantic-interchange/OSI/discussions/69)) 
that would add a `spatial` object to the `dimension` block as the spatial 
equivalent of `is_time` — a role marker that carries the metadata consuming 
tools need: geometry type, SRID, spatial index system and resolution, and 
geographic level. It would look like this:

```yaml
- name: geom
  datatype: geometry              # proposed new enum value
  dimension:
    spatial:                      # proposed role marker + metadata
      type: polygon
      srid: 4326
      geographic_level: "census_block_group"
      geographic_hierarchy:
        parent: "census_tract"
        children: []
  description: "Block group boundary polygon (WGS84)"

- name: h3
  datatype: string                # H3 index is stored as a string
  dimension:
    spatial:
      type: spatial_index
      index:
        system: h3
        resolution: 8
        rollup_resolutions: [7, 6, 5]
      geographic_level: "h3_cell_res8"
  description: "H3 cell index at resolution 8"
```

This maps exactly onto the pattern you're establishing: `datatype` for the
data type question, `dimension.spatial` for the role/semantics question.
We can open a separate discussion on the `spatial` descriptor, but wanted to
flag it here so the `geometry`/`geography` datatype values are on the radar
before the enum is locked in.

**On `number` precision** — +1 to @KSDaemon's concern. We raised the same
point on issue #84 and in a comment there. Worth resolving before this merges.

jonmmease:
Hi @xavipereztr, your descriptions of `geometry` and `geography` make good sense to me as additional data types, though I'd lean toward incorporating these as type additions in your proposal in https://github.com/open-semantic-interchange/OSI/discussions/69 rather than this PR.

I'm not opposed to adding a decimal type, but I'd be interested to articulate a case where a consumer of the semantic model would act differently for a float vs decimal column or expression type.

You mentioned in the other issue:

> So a value like 1.5 in a column declared as unqualified NUMBER/NUMERIC survives on BigQuery but is silently truncated to 2 on Snowflake and Databricks. Precision ceilings also diverge (29 vs 38 total digits).

If this value is in a column with a decimal type in BigQuery or Snowflake, it's either already truncated or the precision accommodates it. If I'm writing SQL generation logic on top of this dimension, are there cases where this needs to differ when the underlying type is decimal vs a float type?

Another factor would be, for translation purposes, whether any mainstream semantic layers already distinguish these numeric types (cube and lookml do not as far as I've seen). If so, that would also be good argument to separate decimal as a dedicated type.

Grounding this in a use case would also help me think through whether `decimal` by itself is useful without the specific `scale` and `precision` parameters, or whether these parameters would be necessary.

xavipereztr:
Hi @jonmmease - on the `geometry` and `geography `aspect, are you suggesting we create a PR based on #69 or shall we wait for further discussion with OSI members? 

On the `numeric` simplification, I guess that you are right that between `integer` vs `number ` is largely enough for most use-cases.

jonmmease:
I only meant that I think it makes sense to attach the additional `geometry`/`geography` data types to the `dimension. spatial_data` proposal in #69, and for these to be evaluated together through the OSI standards process (of which I'm not familiar with the details).

core-spec/spec.md
jonmmease: Following up. Are there cases beyond the generation of SQL for semantic queries that you have in mind to support here? I'm happy to add decimal as data type, but I'd like to understand a motivating scenario for how the spec would be used in such a way that `number` and `decimal` would be treated differently.

core-spec/spec.md
```diff
@@ -36,6 +36,22 @@ Supported SQL and expression language dialects for metrics and field definitions
...
+| `boolean` | Logical two-valued truth type. |
+| `date` | Calendar date with no time-of-day component. |
+| `time` | Time-of-day with no date component. |
+| `timestamp` | Instant-in-time without timezone offset (naive / local). |
```
jochenchrist: maybe the default `timestamp` should be with tz (this is the most common use for timestamps to avoid typical errors). then Add a `timestamp_ntz` for the special case when really no timezone is expected.

core-spec/spec.md
jochenchrist: Using JSON Schema data types (https://json-schema.org/understanding-json-schema/reference/type) as a minimum would be a good idea.

core-spec/spec.md
jonmmease: I don't have a strong opinion on it, but I tend to lean towards postgres naming. But also not opposed to omitting `timestamp` for clarity and only having `timestamp_tz` and `timestamp_ntz`

jonmmease:
Hey all, I'm not up to speed on the decision making process here, or who the decision makers are, but I'd like to do what I can to move this forward. I'm happy to make adjustments to the type system, but not sure we have consensus on this yet. 

Would it be possible for someone with decision making authority here to summarize the desired changes?

jonmmease:
closing in favor of https://github.com/apache/ossie-temp/pull/12

khush-bhatia:
Hey @jonmmease Sorry for the confusion here, now the ASF setup is fully done and the original repo is transferred to apache. With that all the open PRs are correct preserved. I am reopening this PR and reviewing now. The PR on the ossie-temp can be closed.

core-spec/spec.yaml
```diff
@@ -223,11 +221,16 @@ metrics:
     # Should explain what the metric measures and how it's used
     description: string
 
+    # Optional: Logical data type for this metric
+    # One of: string, integer, number, boolean, date, time, timestamp, timestamp_tz, other
+    # Use "other" + custom_extensions for vendor-specific types
+    datatype: string
```
khush-bhatia, line 262: Could you please define the datatype enum in the yaml spec as well ? 

I want to make the json spec the source of truth and have the yaml spec be generated via scripts, but until then we will need to preserve the spec in two places unforunately.

khush-bhatia (dismissed):
LGTM, could you please resolve the merge conflicts and one comment on left on keeping the yaml and json version of the spec in sync.

khush-bhatia:
The datatypes for  geometry/geography can be added in a follow up PR.

jonmmease:
Thanks @khush-bhatia! I brought the changes I made in the `ossie-temp` back here. Main changes are: 
 1) I adopted the type system from the ontology layer (with a few extras that we could consider adding to the ontology value types as well)
 2) I updated several of the converters to propagate types where it was relatively unambiguous how to do so
 3) Added types to core-spec/spec.yaml as well

khush-bhatia (dismissed):

khush-bhatia:
@jbonofre and @QMalcolm could one of you also review this PR since this involves changes to the spec ?

jbonofre:
@khush-bhatia of course! I will! Thanks!

jbonofre:
Can we resolve the conflicts?

core-spec/spec.yaml
jbonofre: Maybe worth to add a TODO comment to be sure we keep `datatypes` in sync with `osi-schema.json`.

jbonofre (dismissed):
LGTM, it's an addition and optional, so no "impact".
I proposed to add a TODO comment to be sure `datatypes` is keep in sync between the `spec.yaml`and the `osi-schema.json`.

We can merge as soon as the merge conflicts will be resolved.

jonmmease:
Done. Conflict resolved and comment added. Thanks @jbonofre!

jonmmease:
Ping to merge before we get more conflicts @jbonofre. thanks!

khush-bhatia (approved):

PR #116, opened 2026-04-29
Initial OSI <> dbt Semantic Layer converters

QMalcolm:
The initial implementation of the OSI <> dbt Semantic Layer converters. This is initially written in python as the simplest path forward required taking MetricFlow as a dependency, though that doesn't need to be the case longer term.

### DIFF
--- a/converters/dbt/src/osi_dbt/msi_to_osi.py
+++ b/converters/dbt/src/osi_dbt/msi_to_osi.py
@@ -104,7 +104,7 @@ class MSIToOSIConverter:
 
         return ConverterResult(
             output=OSIDocument(
-                version="0.1.1",
+                version="0.2.0.dev0",
                 dialects=[self._dialect],
                 semantic_model=[
                     OSISemanticModel(
--- a/converters/dbt/src/osi_dbt/msi_to_osi.py
+++ b/converters/dbt/src/osi_dbt/msi_to_osi.py
@@ -1,0 +1,449 @@
+import re
+from collections import defaultdict
+from dataclasses import dataclass
+from itertools import combinations
+from typing import Dict, List, Optional, Sequence, Tuple
+
+from osi import (
+    OSIDataset,
+    OSIDialect,
+    OSIDialectExpression,
+    OSIDimension,
+    OSIDocument,
+    OSIExpression,
+    OSIField,
+    OSIMetric,
+    OSIRelationship,
+    OSISemanticModel,
+)
+from osi_dbt.converter_issues import ConverterIssue, ConverterIssueType, ConverterResult
+from osi_dbt.filter_utils import _collect_filter_sql, _merge_filter_sqls
+
+from metricflow_semantic_interfaces.enum_extension import assert_values_exhausted
+from metricflow_semantic_interfaces.implementations.semantic_manifest import PydanticSemanticManifest
+from metricflow_semantic_interfaces.protocols.dimension import Dimension
+from metricflow_semantic_interfaces.protocols.entity import Entity
+from metricflow_semantic_interfaces.protocols.measure import (
+    Measure,
+    MeasureAggregationParameters,
+)
+from metricflow_semantic_interfaces.protocols.metric import Metric
+from metricflow_semantic_interfaces.protocols.semantic_model import SemanticModel
+from metricflow_semantic_interfaces.transformations.convert_count import ConvertCountMetricToSumRule
+from metricflow_semantic_interfaces.transformations.semantic_manifest_transformer import (
+    PydanticSemanticManifestTransformer,
+)
+from metricflow_semantic_interfaces.type_enums import (
+    AggregationType,
+    DimensionType,
+    EntityType,
+    MetricType,
+)
+
+
+@dataclass(frozen=True)
+class _EntityEntry:
+    dataset: str
+    col: str
+    entity_type: EntityType
+
+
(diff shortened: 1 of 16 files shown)


converters/dbt/src/osi_dbt/msi_to_osi.py
```diff
@@ -0,0 +1,449 @@
...
+                    ConverterIssue(issue_type=ConverterIssueType.CUMULATIVE_SEMANTICS_LOSS, element_name=metric.name)
+                )
+            expr = self._resolve_metric_expression(metric, metric_index, expression_cache)
+            osi_metrics.append(
+                OSIMetric(
+                    name=metric.name,
+                    expression=self._make_expression(expr),
+                    description=metric.description,
+                )
+            )
```
khush-bhatia, line 103: For cumulative metrics, looks like `continue` is missing ?

khush-bhatia (approved):
LGTM, thanks , just left one comment.

python/src/osi/models.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class OSIDialect(str, Enum):
+    """Supported SQL and expression language dialects."""
+
+    ANSI_SQL = "ANSI_SQL"
+    SNOWFLAKE = "SNOWFLAKE"
+    MDX = "MDX"
+    TABLEAU = "TABLEAU"
+    DATABRICKS = "DATABRICKS"
```
jbonofre, line 16: I guess we should add `MAQL` in the `enum` (as it's in the schema's `Dialect`).

python/src/osi/models.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+class OSIVendor(str, Enum):
+    """Vendors with supported custom extensions."""
+
+    COMMON = "COMMON"
+    SNOWFLAKE = "SNOWFLAKE"
+    SALESFORCE = "SALESFORCE"
+    DBT = "DBT"
+    DATABRICKS = "DATABRICKS"
```
jbonofre, line 26: I guess we should add `GOODDATA` here: it's in our examples but rejected here.

We should decide in this PR is we upgrade to 0.2.0.dev0 version, or stay with 0.1.0.

jbonofre (approved):

python/src/osi/models.py
QMalcolm: Getting MAQL added 🙂

converters/dbt/src/osi_dbt/msi_to_osi.py
QMalcolm: So for cumulative metrics, we don't skip them, we just drop the extra information about them that we aren't able to represent yet. However we can still produce the basic representation, and then emit a warning about what information was dropped.

python/src/osi/models.py
```diff
@@ -0,0 +1,163 @@
...
+
+    dialects: list[OSIDialectExpression]
+
+
+class OSIDimension(BaseModel):
+    """Dimension metadata on a field."""
+
+    model_config = ConfigDict(frozen=True)
+
+    is_time: bool
```
jbonofre: The spec describes `is_time` as optional, but here it's required.

I believe we should use `Optional[bool] = None` here.

WDYT?

converters/dbt/src/osi_dbt/osi_to_msi.py
```diff
@@ -0,0 +1,389 @@
...
+                semantic_models.append(self._convert_dataset(dataset, osi_sm))
+            metrics.extend(self._convert_metrics(osi_sm))
+
+        return ConverterResult(
+            output=PydanticSemanticManifest(
+                semantic_models=semantic_models,
+                metrics=metrics,
+                project_configuration=PydanticProjectConfiguration(),
+            ),
+            issues=[],
```
jbonofre, line 98: Correct me if I'm wrong, but `ConverterIssue` is never thrown here? Should we?

jbonofre (approved):
LGTM to overall. I left a couple of comments/questions, especially about `is_time` (to clarify).

python/src/osi/models.py
khush-bhatia: +1

khush-bhatia (approved):

python/src/osi/models.py
QMalcolm: Will make the change! Might be worth reviewing that in the spec later though 🤔 What does the third state give us if `None` is the same as `False`?

converters/dbt/src/osi_dbt/osi_to_msi.py
QMalcolm: Presently there are no converter issues going OSI to MSI, issues only exist going MSI to OSI currently. However, I wanted to use the same `ConverterResult` class to make future work where there are OSI -> MSI converter issues easier to handle without a breaking change. We _could_ add a
```python
    if result.issues:
        for issue in result.issues:
            verb = "was dropped" if issue.issue_type in _DROPPED_ISSUE_TYPES else "was converted with loss"
            reason = _ISSUE_REASON[issue.issue_type]
            print(f"[WARNING] {issue.issue_type.value}: {issue.element_name} {verb} during conversion because {reason}", file=sys.stderr)
```
 block to `_cmd_osi_to_msi` in `converters/dbt/src/osi_dbt/cli.py`, but that would be dead code at this time.

PR #118, opened 2026-04-30
Add OSI-Salesforce converter module

deutschn:
An initial implementation of the OSI <> Salesforce Semantic Model converters.

### DIFF
--- a/converters/salesforce/src/main/java/org/osi/converter/RelationshipMappingHandler.java
+++ b/converters/salesforce/src/main/java/org/osi/converter/RelationshipMappingHandler.java
@@ -1,0 +1,406 @@
+package org.osi.converter;
+
+import static org.osi.converter.ConverterConstants.*;
+import static org.osi.util.DataStructureUtils.*;
+
+import org.osi.converter.ConverterConstants.Level;
+import org.osi.converter.pipeline.PipelineStep;
+import java.util.*;
+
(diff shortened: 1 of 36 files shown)


converters/salesforce/src/main/java/org/osi/converter/RelationshipMappingHandler.java
```diff
@@ -0,0 +1,406 @@
...
+            List<Object> fromColumns = asList(fromColumnsObj);
+            List<Object> toColumns = asList(toColumnsObj);
+
+            if (fromColumns.isEmpty() || toColumns.isEmpty()) {
+                return;
+            }
+
+            if (fromColumns.size() != toColumns.size()) {
+                return;
+            }
```
khush-bhatia, line 183: All of these should "continue;" instead of return ? We want to skip the bad relationship but still continue handling the other relationships ?

khush-bhatia (approved):
Looks good, thanks. Left some comments.

converters/salesforce/src/main/java/org/osi/converter/RelationshipMappingHandler.java
deutschn: Right, thanks. My intention is to remove those checks in the next PR. They’re all enforced by the schema validation step.
Thank you.

PR #123, opened 2026-05-12
Add working groups page

jbonofre:
This PR introduces docs/working_groups.md page with lead, Slack channel, and Google Calendar (for corresponding meetings, as soon as it will be created).


jbonofre:
I can also include Google Groups, if there's any.

QMalcolm (approved):
Looks good to me! We can leave the google groups as a placeholder for now until we have links to add, and then open a separate PR to add them

PR #124, opened 2026-05-12
Support ontologies and schema mappings from the logical layer

kurtStirewalt:
This PR follows much discussion by the Ontology Working Group whose work is archived here:
https://docs.google.com/document/d/1i2usg8XD_BsI8hc1RDTwqaZaFhgTyTKLhl-5vCj3XXk/edit?tab=t.0#heading=h.ryh06jb7q2hv
Please note the document is a bit behind this PR and will be updated concurrently with its review.

### DIFF
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -1,6 +1,6 @@
 # OSI - Core Metadata Specification
 
-**Version:** 0.1.1
+**Version:** 0.1.2
 
 ## Goals
 
@@ -696,27 +696,24 @@ requires any item that has sales in some store to be offered in that store.
 ### Mappings
 
 Logical to conceptual schema mappings declare how to map field values to conceptual objects
-and links. The most basic kind of mapping is a value map, which has the following schema:
+and links. The most basic kind of mapping is a value map:
 
 | Field | Type | Required | Description |
 |--------------|----------|-----|-------|
 | `role`  | string   | Yes | Expression that indicates the role that the mapped objects will play |
 | `value` | string   | Yes | Expression that calculates the value to map to the role |
 
-These declare how to map a SQL expression that computes a value from one or more fields to a
+which declares how to map a SQL expression that computes a value from one or more fields to a
 role of some relationship, where the role must be played by a value type.
 
-Entity-type objects (entities) are opaque and must be referenced by their identifying
-relationships. These relationships are always binary, with one role played by the entity
-type and the other played by some other concept. In the common case where an entity type
-is identified by a single relationship whose other role is played by a value type, a value
-map suffices to declare the existence of entities of that type and also to look them up
-when mapping to links of a relationship. And if an entity type is identified by a pair
-of such relationships, a pair of value maps suffice for the same purpose.
+Entity-type objects (entities) are opaque and must be mapped to via their identifying
+relationships. These relationships are always binary, with the first role played by
+the entity type itself. In the common case where an entity type is identified by a
+single relationship whose second role is played by a value type, a value map suffices
+to declare the existence of entities of that type or to look them up when mapping to
+links of a relationship.
 
-An entity map is an array of structures, each of which is either a value map or a nested entity
-map that looks up an object to map to some role of an identifier relationship. These structures
-have the following schema:
+An entity map is an array of identifier maps:
 
 | Field | Type | Required | Description |
 |--------------|----------|-----|-------|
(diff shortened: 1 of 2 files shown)


core-spec/spec.md
```diff
@@ -412,6 +422,455 @@ custom_extensions:
...
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `concept` | string | Yes | Unique name of this concept |
+| `description` | string | No | Human-readable description |
+| `relationships` | list | No | Relationships where this concept plays the first role |
+| `extends` | list | No | Names of this concept's supertypes |
+| `derived_by` | list | No | Expressions that derive this concept's population |
+| `identify_by` | list | No | Names of relationships that uniquely reference objects of this concept |
+| `requires` | list | No | Expressions that constrain this concept's population |
+| `entities` | list | No | Mappings that determine this concept's objects (when an entity type) |
```
jochenchrist, line 452: I don't get the `entities` field.

From my understanding, a _concept_ is an entity (if it is an identifiable object).

The examples below reads like `properties` in the sense of a RDF Property, or a OWL DatatypeProperties/ObjectProperties.

If this is the intend, I would recommend to name it `properties`.

core-spec/spec.md
```diff
@@ -412,6 +422,455 @@ custom_extensions:
 
 ---
 
+## Ontology
```
jochenchrist, line 425: Let's discuss with @khush-bhatia if Ontology should be part of the core spec, or if we should split OSI in a semantic_model and ontology (and maybe others) specs.

core-spec/spec.md
```diff
@@ -412,6 +422,455 @@ custom_extensions:
...
+Concepts represent the types of things that have meaning in a business setting, e.g., person, company,
+or salary. Each concept is either an entity type or a value type. Ontologies implicitly include a
+value-type for each basic data type, like `Integer`, `Decimal`, and `String`, and an entity type
+called `Any`. Every other concept ontology extends (is a subtype of) one of these concepts.
+
+Concepts conform to the following schema:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `concept` | string | Yes | Unique name of this concept |
```
jochenchrist, line 445: You mentioned below, "for concepts of type entity"
-> I would introduce a type field here.

core-spec/spec.md
```diff
@@ -412,6 +422,455 @@ custom_extensions:
...
+| `description` | string | No | Human-readable description |
+| `relationships` | list | No | Relationships where this concept plays the first role |
+| `extends` | list | No | Names of this concept's supertypes |
+| `derived_by` | list | No | Expressions that derive this concept's population |
+| `identify_by` | list | No | Names of relationships that uniquely reference objects of this concept |
+| `requires` | list | No | Expressions that constrain this concept's population |
+| `entities` | list | No | Mappings that determine this concept's objects (when an entity type) |
+| `links` | list | No | Mappings from field values to relationship links |
```
jochenchrist, line 453: When I read "links" I was thinking of HREF links, but these links represent relationships.
TBH, I would drop them again for simplicity.

core-spec/spec.md
```diff
@@ -412,6 +422,455 @@ custom_extensions:
...
+Relationships relate objects of one or more concepts and declare how to verbalize links among
+those objects. Relationships have set (as opposed to bag) semantics, and links do not contain
+nulls.
+
+Each relationship that is declared under a concept conforms to the following schema:
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `name` | string | Yes | Part of the identifier for this relationship |
+| `description` | string | No | Human-readable description |
```
jochenchrist, line 477: I would also consider adding a `type` for relationships to express (`isA`, `hasProperty`, `belongsTo`...) to have a qualifier for the relationship.

core-spec/spec.md
khush-bhatia: I suggest that we move this to a separate spec file, we can call it ontology-spec.yaml.  There are multiple converters already added to the repo , see the `[converters](https://github.com/open-semantic-interchange/OSI/tree/main/converters)` from Salesforce, good-data, dbt(PR is ready), unity catalog metric(PR is ready) etc. All of these converters are for products that have a semantic model layer, but they don't have an ontology. So these converters shouldn't have to deal with the ontology part of the spec. 

Similarly as we add new converters for the ontology, these converters shouldn't have to think about the semantic model part of the spec.

I also think versioning of these two specs can happen independently if they are in separate files. A breaking change in ontology spec shouldn't require updating the versioning of the semantic model and all its converter code and vice versa.

jochenchrist: +1

core-spec/spec.md
kurtStirewalt: Yes, that's a very good idea. Will add that

core-spec/spec.md
kurtStirewalt: @jochenchrist , this field contains logical -> conceptual schema mappings that populate the entities (objects of entity type concepts). The contents of this field were originally in a separate top-level "mappings" section.

PR #132, opened 2026-05-20
Add basic legend transpiler to generate Legend models/mapping/store from OSI models

erwin-jeanlouis-gs:
Basic OSI to Legend transpiler


converters/legend/tests/test_osi_to_legend.py
```diff
@@ -0,0 +1,786 @@
+"""Tests for OSI to Legend converter."""
```
jbonofre, line 1: I have the following test failures:
-FAILED TestPureDslOutput::test_pure_table_declaration
- FAILED TestPureDslOutput::test_pure_primary_key_marking
- FAILED TestPureDslOutput::test_pure_composite_keys
- FAILED TestOntologySupport::test_entities_generate_mapping_block

The first three assert column declarations with a colon (`user_id: VARCHAR(256)`), but `Column.to_pure_declaration` at `src/legend_osi/models.py:23-26` emits no colon: `user_id VARCHAR(256)`. The asserts are wrong; please fix them to match the emitted (and correct) syntax.

`test_entities_generate_mapping_block` at `tests/test_osi_to_legend.py` opens `base_dir / "example.pure"` but `base_dir = Path(__file__).parent` resolves to `tests/`, and the file lives at `tests/example/example.pure`. Path must be `base_dir / 
  "example" / "example.pure"`. Additionally, the test's OSI input defines two datasets (store_sales, date_dim) and three concepts (StoreSales, StoreSalesNr, TicketNr), yet tests/example/example.pure contains an extra customer table and
  Customer/CustomerName concepts that the input doesn't produce — so the actual == expected comparison will still fail after the path fix. Either trim the expected fixture or extend the test input.

Can you please fix the tests?

converters/legend/pyproject.toml
```diff
@@ -0,0 +1,40 @@
+[project]
+name = "legend-osi"
+version = "0.1.0"
+description = "Converter from OSI semantic models to FINOS Legend Database representation"
+authors = [{name = "OSI Contributors"}]
+readme = "README.md"
```
jbonofre, line 6: I think this is not correct as there's no `README.md` in `converters/legend`.

Can you please fix that (remove `readme` or add `README.md` in the right directory)?

converters/legend/src/cli.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
```
jbonofre, line 1: I'm confused here: it should be in the `legend_osi` folder right?

jbonofre (changes requested):
I will provide more feedback on code (`osi_to_legend.py`) and the documentation (I believe we can group the MDs in one or two).

PR #136, opened 2026-05-27
Add bidirectional Honeydew ↔ OSI converter

baruchoxman:
## Summary

- Adds a new `converters/honeydew/` package with a bidirectional converter between OSI semantic model YAML and Honeydew workspace directories
- Supports both directions: `osi-to-honeydew` and `honeydew-to-osi` via a CLI and Python API
- Full round-trip fidelity: OSI-only fields (`ai_context`, `unique_keys`, non-Honeydew `custom_extensions`) are stored in Honeydew `metadata`; Honeydew-native fields (`owner`, `display_name`, `hidden`, `format_string`, `timegrain`, `labels`) are preserved in a `HONEYDEW` custom extension

## What's included

- `src/honeydew_osi_converter.py` — ~1000-line converter covering datasets, fields, calculated attributes, metrics, relationships, and workspace-level metadata
- `tests/test_honeydew_osi_converter.py` — 127 parametrized tests with full-dict output verification (no cherry-picked field assertions)
- `pyproject.toml` — package config, Python 3.12+ requirement, pytest setup
- `README.md` — mapping tables, usage, limitations

## Test plan

- [x] `python -m pytest tests/` passes (127 tests, ~1s)
- [x] `python src/honeydew_osi_converter.py osi-to-honeydew -i <file> -o <dir>` produces a valid Honeydew workspace
- [x] `python src/honeydew_osi_converter.py honeydew-to-osi -i <dir> -o <file>` produces a valid OSI YAML
- [x] Round-trip tested against the TPC-DS example in `examples/tpcds_semantic_model.yaml` — 5 datasets, 31 fields, 4 relationships, 5 metrics round-trip with zero semantic differences

### DIFF
--- a/converters/honeydew/src/honeydew_osi_converter.py
+++ b/converters/honeydew/src/honeydew_osi_converter.py
@@ -192,7 +192,7 @@ def _dataset_to_files(
             raise HoneydewConversionError(f"Field missing 'name' in dataset '{entity_name}'")
 
         expr = _pick_ansi_expression(field.get("expression"), field_name)
-        if expr is None:
+        if not expr:
             continue
 
         datatype = _osi_field_to_honeydew_datatype(field)
@@ -223,7 +223,10 @@ def _dataset_to_files(
             custom_extensions=field_ext or None,
         )
 
-        if _is_simple_identifier(expr):
+        hd_hint = _get_honeydew_extension(field)
+        force_calc = hd_hint.get("type") == "calculated_attribute"
+
+        if _is_simple_identifier(expr) and not force_calc:
             attr: dict[str, Any] = {"column": expr, "name": field_name, "datatype": datatype}
             if effective_desc:
                 attr["description"] = effective_desc
@@ -293,13 +296,18 @@ def _dataset_to_files(
 
         metric_ext = [e for e in (metric.get("custom_extensions") or []) if e.get("vendor_name") != HONEYDEW_VENDOR]
         metric_meta = _build_osi_metadata(
-            ai_context=metric_ai_ctx if isinstance(metric_ai_ctx, dict) else None,
+            ai_context=metric_ai_ctx,
             custom_extensions=metric_ext or None,
         )
         if metric_meta:
             metric_dict["metadata"] = [metric_meta]
 
-        files[f"{base}/metrics/{mname}.yml"] = _dump(metric_dict)
+        metric_path = f"{base}/metrics/{mname}.yml"
+        if metric_path in files:
+            warnings.warn(
+                f"Metric '{mname}' in entity '{entity_name}' is defined more than once; later definition wins"
+            )
+        files[metric_path] = _dump(metric_dict)
 
     return files
 
@@ -331,6 +339,10 @@ def _osi_relation_to_honeydew(rel: dict[str, Any]) -> dict[str, Any] | None:
             {"src_field": fc, "target_field": tc}
             for fc, tc in zip(from_cols, to_cols)
         ]
+    elif not from_cols:
+        hd_ext = _get_honeydew_extension(rel)
+        if hd_ext.get("connection_expr"):
+            honeydew_rel["connection_expr"] = {"sql": hd_ext["connection_expr"]}
     return honeydew_rel
 
 
@@ -364,6 +376,9 @@ def _pick_ansi_expression(expression: Any, field_name: str) -> str | None:
 
 
 def _osi_field_to_honeydew_datatype(field: dict[str, Any]) -> str:
+    hd_ext = _get_honeydew_extension(field)
+    if hd_ext.get("datatype"):
+        return hd_ext["datatype"]
     dimension = field.get("dimension")
     if isinstance(dimension, dict) and dimension.get("is_time"):
         return "timestamp"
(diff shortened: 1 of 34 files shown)


converters/honeydew/src/honeydew_osi_converter.py
```diff
@@ -0,0 +1,1018 @@
...
+            if dim is not None:
+                field["dimension"] = dim
+            if attr.get("description"):
+                field["description"] = str(attr["description"]).strip()
+
+            attr_osi_meta = _read_osi_metadata(attr)
+            attr_labels = attr.get("labels") or []
+
+            # Restore ai_context (structured form takes priority, else build from labels)
+            if attr_osi_meta.get("ai_context"):
```
jbonofre, line 727: I'm a bit confused here. I think it's not fully idempotent.

My understanding is when we do honeydew to osi convert, any attribute that has `labels` always get an `ai_context.synonyms` from those labels. Even when those labels were just round-tripped from OSI `label` field that has no `ai_context`.

Imagine, we start from OSI with:

```
  - name: status
    label: sales
    expression: { dialects: [{ dialect: ANSI_SQL, expression: status }] }
```

We convert to honeydew:

```
  - name: status
    column: status
    datatype: string
    labels: [sales]
```

Then we convert to OSI again:

```
  - name: status
    label: sales
    ai_context:
      synonyms: [sales]
    expression: { dialects: [{ dialect: ANSI_SQL, expression: status }] }
```

I think the converter should check where the `labels` are coming from.

converters/honeydew/src/honeydew_osi_converter.py
```diff
@@ -0,0 +1,1018 @@
...
+        "honeydew_extra": {},
+    }
+
+    entity_yml = os.path.join(entity_dir, f"{entity_name}.yml")
+    if os.path.exists(entity_yml):
+        with open(entity_yml) as f:
+            ey = yaml.safe_load(f) or {}
+        data["keys"] = _coerce_list(ey.get("keys"))
+        data["description"] = ey.get("description")
+        data["key_dataset"] = ey.get("key_dataset")
```
jbonofre, line 618: I believe honeydew entities can have multiple datasets. You clearly describe the limitation in the README.md (which is fine).

I would suggest to add a warning in the converter when the entity has more than one dataset.

jbonofre (approved):
I also suggest to add `HONEYDEW` in the list of supported vendors (in the `converters/index.md` and `core-spec`).

Overall good, but worth to clarify the comments.

converters/honeydew/src/honeydew_osi_converter.py
baruchoxman: Fixed. The OSI `label` value is now stored in the osi metadata section when a field has a label but no dict `ai_context`. On the Honeydew→OSI path, the presence of that metadata key suppresses synthetic `ai_context.synonyms` generation, so the round-trip is clean.

converters/honeydew/src/honeydew_osi_converter.py
baruchoxman: Fixed. Added a `warnings.warn` when an entity has more than one dataset file, so users are notified that only the primary dataset will be converted.

converters/honeydew/src/honeydew_osi_converter.py
```diff
@@ -0,0 +1,1037 @@
...
+        return None
+
+    rel_type = (rel.get("rel_type") or "many-to-one").lower()
+    connection = rel.get("connection") or []
+    connection_expr = rel.get("connection_expr")
+
+    if rel_type == "many-to-one":
+        from_entity, to_entity = entity_name, target
+        from_cols = [c.get("src_field", "") for c in connection]
+        to_cols = [c.get("target_field", "") for c in connection]
```
khush-bhatia, line 840: To maintain the metadata about the many-to-one relationship here, it would be good to make sure that the to_cols are set as unique_keys in the dataset.

converters/honeydew/src/honeydew_osi_converter.py
```diff
@@ -0,0 +1,1037 @@
...
+    elif rel_type == "one-to-many":
+        from_entity, to_entity = target, entity_name
+        from_cols = [c.get("target_field", "") for c in connection]
+        to_cols = [c.get("src_field", "") for c in connection]
+    else:
+        from_entity, to_entity = entity_name, target
+        from_cols = [c.get("src_field", "") for c in connection]
+        to_cols = [c.get("target_field", "") for c in connection]
```
khush-bhatia, line 848: Same comment as above.

khush-bhatia (approved):
Thanks, just one comment about the relationship type metadata mapping to unique keys.

converters/honeydew/src/honeydew_osi_converter.py
baruchoxman: Addressed in 9619e8b. Rather than reading the relation here, the converter now derives `unique_keys` from the target entity's `keys`: an entity's keys are emitted as both `primary_key` and a `unique_keys` entry. Since Honeydew validates that a many-to-one relation joins to the target entity's keys, the relation's `to_columns` are always covered by the target dataset's `unique_keys`.

To keep round-trips clean, on the OSI→Honeydew path a unique key identical to the primary key is not persisted back into Honeydew `metadata` (it's re-derived), so `Honeydew→OSI→Honeydew` is unchanged; `OSI→Honeydew→OSI` normalizes by surfacing the primary key as a unique key. Added `test_honeydew_to_osi_relation_target_columns_are_unique_keys` and documented it in the README.

converters/honeydew/src/honeydew_osi_converter.py
baruchoxman: Same fix in 9619e8b — `unique_keys` is now derived from the target entity's `keys` for every relation type (the converter normalizes all relations to many-to-one with the `to` side being the unique side), so this branch is covered too.

baruchoxman:
> Thanks, just one comment about the relationship type metadata mapping to unique keys.

Thank you for the review! Addressed the comments, please feel free to merge if it looks good.

PR #137, opened 2026-05-27
Remove top-level dialects and vendors from the json schema

khush-bhatia:
## Summary
- Remove unused top-level `dialects` and `vendors` properties from `core-spec/osi-schema.json`
- Root schema now contains only `version` and `semantic_model` (both required)
- `Dialect` and `Vendor` `$defs` are retained for use within `Expression` and custom extensions

## Why
- These top-level arrays were never referenced by `examples/tpcds_semantic_model.yaml` or by `validation/validate.py`
- `validate.py` reads dialects from each expression's per-dialect entries via `$defs/Expression`, not from the root
- Keeps the root schema minimal and focused

## Test plan
- [x] Ran `python3 validation/validate.py examples/tpcds_semantic_model.yaml` -> `Validation PASSED`
- [x] Converter tests

.... Generated with [Cortex Code](https://docs.snowflake.com/en/user-guide/cortex-code/cortex-code)

### DIFF
--- a/core-spec/osi-schema.json
+++ b/core-spec/osi-schema.json
@@ -10,20 +10,6 @@
       "const": "0.2.0.dev0",
       "description": "OSI specification version"
     },
-    "dialects": {
-      "type": "array",
-      "description": "Supported expression language dialects (enumeration definition)",
-      "items": {
-        "$ref": "#/$defs/Dialect"
-      }
-    },
-    "vendors": {
-      "type": "array",
-      "description": "List of vendor names used in custom extensions within this model",
-      "items": {
-        "$ref": "#/$defs/Vendor"
-      }
-    },
     "semantic_model": {
       "type": "array",
       "description": "Collection of semantic model definitions",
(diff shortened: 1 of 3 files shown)


jbonofre (approved):
LGTM.

However, I wonder if `core-spec/spec.yaml` and `core-spec/spec.md` should be updated in this PR.
I suggest to either update this PR or create an issue to track the update of `spec.yaml` and `spec.md`: the root-level `dialects`/`vendors` should be removed after the JSON schema has dropped them.

khush-bhatia:
> LGTM.
> 
> However, I wonder if `core-spec/spec.yaml` and `core-spec/spec.md` should be updated in this PR. I suggest to either update this PR or create an issue to track the update of `spec.yaml` and `spec.md`: the root-level `dialects`/`vendors` should be removed after the JSON schema has dropped them.

Fixed the comments and relevant sections in this PR. The  `core-spec/spec.yaml` and `core-spec/spec.md` did not have the array holder type to begin with. It was a mistake in the json spec only.

PR #144, opened 2026-06-02
Add OSI -> LookML (BigQuery) converter

1waterrj:
Adds a first-class converter that exports OSI semantic models to LookML, the semantic modeling language for Looker (Google Cloud's semantic layer over BigQuery), analogous to the existing OSI -> Snowflake Cortex Analyst converter.

- converters/bigquery/: offline OSI -> LookML converter, README, tests, and a BigQuery-dialect example fixture
  - datasets -> views (sql_table_name; backtick-quoted 3-part BigQuery refs; derived_table for subquery sources)
  - fields -> dimensions; time fields -> dimension_group
  - metrics -> measures; relationships -> explore/join (composite-key aware)
  - prefers the BIGQUERY dialect, falls back to ANSI_SQL
- validation/validate.py: map the BIGQUERY dialect to sqlglot's bigquery dialect so GoogleSQL expressions (e.g. SAFE_CAST) validate correctly
- converters/index.md: list BigQuery/LookML as a supported converter

26 passing tests; the converter handles the ANSI-only TPC-DS example as-is.


vinlee19:
Really looking forward to this! LookML is the most feature-complete and expressive semantic modeling language I’ve seen. Our internal semantic layer is also largely based on LookML, so first-class OSI-to-LookML support would be especially valuable to us.

PR #145, opened 2026-06-03
Ontology: Drop Float as a built-in concept

jochenchrist:
## Summary

`Float` and `Decimal` describe the same thing at the conceptual level: the most general floating-point/decimal value type. Carrying both as separate built-in concepts is redundant and invites ambiguity about which one to extend when modeling a non-integer numeric value type.

This PR drops `Float` as a built-in concept and keeps `Decimal` as the single built-in for non-integer numeric values.

## Changes

- `ontology/ontology.md`: remove the `Float` row from the built-in concepts table.
- `examples/flights.yaml`: update the runway `length` and `shapeLength` mappings (and their verbalizations) to use `Decimal` instead of `Float`, so the example still references only declared built-ins.

## Notes

No JSON schema or validation code enumerated the built-in concepts, so the doc table and the one example were the only references to `Float`.

### DIFF
--- a/examples/flights.yaml
+++ b/examples/flights.yaml
@@ -14,8 +14,8 @@ ontology:
         multiplicity: ManyToOne
       - name: shapelength
         roles:
-          - concept: Float
-        verbalizes: [ '{Example_Runway} shapeLength {Float}' ]
+          - concept: Decimal
+        verbalizes: [ '{Example_Runway} shapeLength {Decimal}' ]
         multiplicity: ManyToOne
       - name: airportid
         roles:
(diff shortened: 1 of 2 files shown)


khush-bhatia (approved):
LGTM , I am good with merging this , but I think we should follow up and converge on a common spec for datatype for both the ontology and the semantic model spec.

khush-bhatia:
@kurtStirewalt Could you review as well ?

kurtStirewalt:
I am not following the motivation behind this PR. Decimal is a fixed-point (not a floating-point) type, and I think we need to distinguish these different kinds of values even at the conceptual level. Languages like Object Role Modeling, Legend, and RAI's PyRel all distinguish floats from decimals, and xsd:decimal is a distinct RDF datatype from xsd:float.

Is there some conceptual modeling language that does not distinguish these that you are looking to support? If so, maybe the best thing to do would be to handle the conversion in a transpiler?

jochenchrist:
Let's close this issue in favor of a common type model with the semantic model spec.

PR #146, opened 2026-06-03
Ontology: Add custom_properties (e.g. for OWL import)

jochenchrist:
## Summary

Importing an ontology from an external format such as **OWL/RDF** currently loses any data that the core OSI spec does not model (labels, equivalence axioms, annotations, provenance, and so on) because every ontology object is locked down with `additionalProperties: false`. There is no place to put it, so a round-trip (OWL → OSI → OWL) is lossy.

This PR adds an open **`custom_properties`** object to the ontology root, concepts, relationships, and roles. Keys are free-form and values may be any JSON, so importers can preserve foreign data and round-trip it. The spec does not interpret or constrain the structure of `custom_properties`. The rest of each object keeps `additionalProperties: false`, so `custom_properties` is the single, explicit escape hatch rather than letting arbitrary keys leak in anywhere.

```yaml
ontology:
  - concept: ContributionMargin
    type: ValueType
    extends: [Decimal]
    custom_properties:
      abbreviation: CM
      formula: revenue - variable_costs
      unit: EUR
      better_when: higher
```

An OWL importer would instead key properties by the source predicate (`rdfs:label`, `owl:equivalentClass`); see `examples/foaf_owl_import.yaml`. The choice of keys belongs to the tool that writes them, not the spec.

## Design notes

The core spec (logical model) already has a `custom_extensions: [{vendor_name, data}]` mechanism where `data` is a stringified JSON blob. I deliberately did **not** reuse that here:

- OWL/RDF is a standard interchange format, not a single "vendor", so a `vendor_name` blob is an awkward fit.
- A structured map keeps imported annotations as real, queryable/validatable JSON instead of an opaque string.

If maintainers would prefer to unify on one extensibility mechanism spec-wide, I'm happy to switch this to the `custom_extensions` shape instead. Flagging the choice explicitly.

## Changes

- `ontology/ontology.json`: add a reusable `CustomProperties` `$def` and reference it from the root, `Concept`, `Relationship`, and `Role`.
- `ontology/ontology.md`: add the field to the four schema tables, a new "Custom properties" section with a metrics example, and a version-history note.
- `examples/foaf_owl_import.yaml`: a small FOAF/OWL import example exercising `custom_properties` at the root, concept, and relationship levels, with a `prefixes` map and QName keys.

## Validation

- `ontology/ontology.json` is well-formed JSON.
- The new example and the existing `examples/flights.yaml` both validate against the updated schema (Draft 2020-12).
- Confirmed unknown keys are still rejected everywhere except inside `custom_properties`.


kurtStirewalt (dismissed):
This is a nice addition!

examples/foaf_owl_import.yaml
```diff
@@ -0,0 +1,39 @@
...
+name: foaf
+description: >
+  Small ontology imported from the FOAF (Friend of a Friend) OWL vocabulary,
+  demonstrating how custom_properties preserves OWL/RDF data (IRIs, labels, and
+  axioms) that the core OSI spec does not model, so the source can be round-tripped.
+ontology:
+  - concept:
+      name: Person
+      type: EntityType
+      custom_properties:
```
jbonofre, line 11: I'm confused here. Is it really `custom_properties` or `custom_extensions`?

jbonofre (dismissed):
Overall good, I just wonder if we should use `custom_properties` or `custom_extensions`.

@jochenchrist @kurtStirewalt thoughts?

examples/foaf_owl_import.yaml
jochenchrist: The thing with current `custom_extensions` of the core spec is that it is not structured, and just a string, which makes it hard to edit. Open to find better names for this key of couse.

jochenchrist:
@jbonofre  @kurtStirewalt  
To push this PR, we would need a decision.

# Option A

Structured `custom_properties`:

Example:

```
  - concept:
      name: Person
      type: EntityType
      custom_properties:
        iri: "http://xmlns.com/foaf/0.1/Person"
        rdfs:label@en: "Person"
        rdfs:label@fr: "Personne"
        rdfs:comment@en: "A person."
        owl:equivalentClass: "https://schema.org/Person"
```

Consequences:
- Different to core-spec v0.0.1 `custom_extensions`
- I would introduce a PR to deprecate `custom_extensions` and introduce `custom_properties` in core spec as well.


# Option B

Same with `custom_extensions`

```
  - concept: Person
    type: EntityType
    custom_extensions:
      - vendor_name: OWL
        data: |
          {
            "iri": "http://xmlns.com/foaf/0.1/Person",
            "rdfs:label@en": "Person",
            "rdfs:label@fr": "Personne",
            "rdfs:comment@en": "A person.",
            "owl:equivalentClass": "https://schema.org/Person"
          }
```

Consequences:

- Following core-spec v0.0.1 version
- This data string (JSON in String...) is hard to read, serialize, and validate. But maybe good-enough? 


What do you think?

scotthenninger:
Option A would be my preference.

On language tags why not use the RDF convention of delimiting the string, e.g. 
`rdfs:comment “A person”@en`

ontology/ontology.json
```diff
@@ -55,6 +58,11 @@
   "required": ["version", "name", "ontology"],
   "additionalProperties": false,
   "$defs": {
+    "CustomProperties": {
+      "type": "object",
+      "description": "Open set of custom properties for carrying data that is not covered by the core ontology spec, for example when importing an ontology from an external format such as OWL/RDF. Keys are free-form; CURIEs such as 'owl:equivalentClass' or 'rdfs:label' are recommended when round-tripping semantic-web sources. Values may be any JSON.",
```
Ostrzyciel: CURIEs are recommended, but how do you plan to resolve them? Must the relevant prefixes be defined in the `prefixes` field? If not, this will be in practice an unresolvable mess.

Ostrzyciel: Consider using full IRIs, they are unambiguous.

ontology/ontology.md
```diff
@@ -379,6 +383,65 @@ ontology:
...
+#### Language-tagged literals
+
+RDF literals often carry a language tag (for example `rdfs:label` with values `"Person"@en` and
+`"Personne"@fr`). The recommended convention is to append the [BCP 47](https://www.rfc-editor.org/info/bcp47)
+language tag to the key after an `@`, giving keys of the form `prefix:local@lang`:
+
+```yaml
+custom_properties:
+  rdfs:label@en: "Person"
+  rdfs:label@fr: "Personne"
```
Ostrzyciel: This will be very messy to parse, in my opinion. Why not:

```
rdfs:label:
  en: "Person"
  fr: "Personne"
```

This would not require mixing in the language tag into the property IRI, which is some new invented syntax. The syntax I wrote above is already used in LinkML-Scala.

ontology/ontology.md
```diff
@@ -379,6 +383,65 @@ ontology:
...
+
+`@` is not a legal character in a CURIE local name, so `prefix:local@lang` parses unambiguously into the
+property and its language. This form holds at most one value per language per key. When a property needs
+multiple values in the same language, or carries a non-language datatype such as `xsd:date`, use the
+[JSON-LD](https://www.w3.org/TR/json-ld11/) value-object form instead, which is lossless:
+
+```yaml
+custom_properties:
+  rdfs:label:
+    - { "@value": "Person", "@language": "en" }
```
Ostrzyciel: in this case I would propose to instead use:

```
rdfs:label:
  en: 
    - "Person"
    - "Persona"
```

The syntax you propose here is borrowed from JSON-LD, which is currently not used anywhere in Ossie, so I see no reason to do it this way.

ontology/ontology.md
```diff
@@ -379,6 +383,65 @@ ontology:
...
+    - { "@value": "Persona", "@language": "en" }
+  dcterms:created:
+    "@value": "2020-01-01"
+    "@type": "xsd:date"
```
Ostrzyciel: This is a datatype literal, not a language-tagged literal.

This is getting very complex, very quickly, and we are right now designing a new YAML syntax for RDF – why?

If this feature is intended for custom extensions that are not part of Ossie, then this should not be standardized in Ossie. Otherwise, we will spend weeks discussing how to represent every corner case of RDF in YAML, which is not the goal of Ossie, I think.

I'd like to ask people in charge of the spec to think this over carefully. It's not a problem to dump ideas into a spec, but it will not be a useful spec.

Ostrzyciel: Why not just say that custom_properties can contain whatever the user wants, and that's it? Cut all these paragraphs?

Ostrzyciel (changes requested):

ontology/ontology.md
jochenchrist: > Why not just say that custom_properties can contain whatever the user wants, and that's it? Cut all these paragraphs?

Yes, let's go that way.
The discussion on lang-strings are acutally a separate topic. I will change this PR to be specific to custom properties.

dockstreet:
@jochenchrist 

Curious was this considered?

https://www.w3.org/TR/yaml-ld-10/

ontology/ontology.md
dockstreet: @jochenchrist @scotthenninger 
Curious was this considered? 

https://www.w3.org/TR/yaml-ld-10/

jochenchrist: https://www.w3.org/TR/yaml-ld-10/ might have been an alternative to the merged `prefixes:` style PR (#332). 

To focus this PR on custom properties, I will update it to make not any assumptions on the structure of the custom properties.

examples/foaf_owl_import.yaml
jochenchrist: See discusssion below with Option A or Option B.

ontology/ontology.json
jochenchrist: I changes this in the PR to not be opinionated on the structure and content of custom properties.

ontology/ontology.md
jochenchrist: I also changed the examples to be less opinoated.

ontology/ontology.md
jochenchrist: > Why not just say that custom_properties can contain whatever the user wants, and that's it? Cut all these paragraphs?

Agree.

MarioDeFelipe:
Thanks @jochenchrist  — this unblocks a concrete case for us. SAP CSN carries annotations that have no home in the ontology layer today (@Semantics.*, @ODM.entityName, multi-language labels); with custom_properties on concepts the converter can carry them through. 👍 on arbitrary JSON.

Two things I'd ask for in the spec text, because they decide whether this is a safe place or a leak:

A precedence rule. If something has a spec home, it goes in the spec home. In the example, formula reads like derived_by, and synonyms reads like a verbalization concern. If those can legally live in custom_properties, an emitter can express real semantics in a place, but the consumer might chose it or not, and if they are needed, must be explicit. I suggest one sentence — "**custom properties are for what the spec cannot express, not for what the emitter chose not to express"**

Enforcing a consumer obligation. We've seen a platform accept a document with a success response and silently discard parts of it (I wrote that up on #246). Custom properties are the most likely place for that to happen, because ignoring them is cheap and invisible. I'd ask that a conforming consumer either preserve custom properties on round-trip or declare that it drops them — the same represent / reject / declare-unsupported rule the rest of the spec is moving toward. Silent drop should **not** be conformant.

jbonofre:
I'm doing a new pass on this one 😄

Issue #52, opened 2026-01-29
Only allow one dialect per OSI document


sfc-gh-wpugh:
The current OSI spec deals with dialects in a very heavyweight way:
- Expressions allow for multiple implementations, since they can have different ones for each dialect
- Expressions are commonly used for Fields and Metrics.

This has a number of long term implementations that may make these untenable:
- Any new concept with an expression will need to add this as well
- We don't have any semantics around what to do if there are multiple dialects for an expression.  Which should get run?
- We don't have any way to ensure they all act the same.

This may ultimately make this standard untenable to implement and unreliable for using.  If there is drift across the different dialect implementations, then depending on which is chosen, the user may get different results.  Since, they are all setup with the same name this adds inherent fragility into the spec.

This issue proposes a different approach:
1. Each expression has only one implementation
2. There is a dialect at the document level
3. Semantic View Composability will support documents crossing dialects (providers can provide errors in the case they cannot support the included dialect).

This will make dialect optional.  If it is not filled in.  If OSI determines a common expression language, than the default will be that.  If not, then the provider should assume its own dialect.

PR #148, opened 2026-06-08
fix(converters): emit schema-conformant OSI documents (drop non-spec root dialects/vendors)

andreybavt:
## Problem
Output from the dbt→OSI converter fails OSI's own validator (`validation/validate.py`)
against the published core schema:

    [Schema] (root): Additional properties are not allowed ('dialects' was unexpected)

Because the CLI writes via `to_osi_yaml()`, every `dbt→OSI` conversion produces a
non-conformant document.

## Root cause
`OSIDocument` (`python/src/osi/models.py`) declared optional root-level
`dialects`/`vendors` fields that aren't in `core-spec/osi-schema.json`, whose root is
`additionalProperties: false` (only `version` + `semantic_model`). The dbt converter
set `dialects=[self._dialect]`, so `dialects` was emitted at the document root.

## Fix
- Remove the non-spec `dialects`/`vendors` fields from `OSIDocument`.
- Stop the dbt converter from populating a document-root dialect.

Per-expression dialect tagging (`OSIExpression.dialects`) is unchanged and remains the
schema-valid home for dialects, so **no information is lost**. Dialect *selection* still
flows end-to-end; the two affected converter tests now assert it on the per-expression path.

## Regression guard
Adds `converters/dbt/tests/test_schema_conformance.py`, which converts a representative
manifest and validates the emitted document (YAML and JSON, for `ANSI_SQL` and
`SNOWFLAKE`) against `core-spec/osi-schema.json`, reusing `validation/validate.py`.
CI now fails if a converter emits a non-conformant document root again.

## Testing
- `validate.py` on converter output: **fails before** (`'dialects' was unexpected`),
  **passes after**.
- Full dbt converter test suite green.

## Out of scope
Whether OSI should support a document-level (default) dialect is the open discussion in
#52 (one dialect per document) and #16 (default dialect at dataset level). This PR takes
**no position** and **makes no schema change** - it only aligns the reference model and
converter with the schema as published today.

### DIFF
--- a/converters/dbt/pyproject.toml
+++ b/converters/dbt/pyproject.toml
@@ -26,6 +26,7 @@ packages = ["src/osi_dbt"]
 
 [tool.uv]
 dev-dependencies = [
+    "jsonschema>=4.0",
     "pytest>=8.0",
     "syrupy>=4.0",
 ]
--- a/converters/dbt/src/osi_dbt/msi_to_osi.py
+++ b/converters/dbt/src/osi_dbt/msi_to_osi.py
@@ -105,7 +105,6 @@ class MSIToOSIConverter:
         return ConverterResult(
             output=OSIDocument(
                 version="0.2.0.dev0",
-                dialects=[self._dialect],
                 semantic_model=[
                     OSISemanticModel(
                         name=osi_model_name,
--- a/converters/dbt/tests/test_msi_to_osi.py
+++ b/converters/dbt/tests/test_msi_to_osi.py
@@ -285,7 +285,6 @@ class TestDialectConfiguration:
         )
         result = MSIToOSIConverter().convert(_manifest(semantic_models=[sm])).output
 
-        assert result.dialects == [OSIDialect.ANSI_SQL]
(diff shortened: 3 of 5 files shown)


andreybavt:
@khush-bhatia , let me know if this first PR follows the community guidelines. Happy to take bigger scope after this first one lands

khush-bhatia (approved):
Thanks for the fix.

QMalcolm (approved):
Looks good to me! Thank you for putting this together 🙂

khush-bhatia:
Thanks @andreybavt, We will merge this PR soon, we are waiting on some process to complete.

andreybavt:
Thanks @khush-bhatia and @QMalcolm ! Will continue with the other PRs then

khush-bhatia:
@andreybavt Because of the rename, there are some merge conflicts, could you please resolve the conflicts.

kayemkim:
Since this was opened, #297 added root-level `dialects`/`vendors` to the schema, so the premise here (that the schema rejects them) no longer holds on main. I have written the details up in #393 rather than here, since it also touches #306, the Sigma converter and the orionbelt tests.

PR #149, opened 2026-06-09
Update flights.yaml example following Summit presentation and fixed small omission in ontology spec

kurtStirewalt:
Updating examples/flights.yaml to reflect the full example we presented at Summit. Fixed 2 omissions in the ontology spec:
- requires constraints can be defined on relationships
- global requires constraints can be defined at the top level (to say things like the population a concept must be non-empty)

### DIFF
--- a/examples/flights.yaml
+++ b/examples/flights.yaml
@@ -1,4 +1,4 @@
-version: 0.1.1
+version: 0.2.0.dev0
 name: flights
 description: Ontology for flight-related concepts and relationships
 ontology:
--- a/examples/flights.yaml
+++ b/examples/flights.yaml
@@ -5,7 +5,7 @@ ontology:
     name: NrFeet
     description: "Unit of measure for distance in feet"
     type: ValueType
-    extends: [ Float ]
+    extends: [ Decimal ]
 - concept:
     name: NrPounds
     description: "Unit of measure for weight in pounds"
@@ -15,7 +15,7 @@ ontology:
     name: NrMiles
     description: "Unit of measure for distance in miles"
     type: ValueType
-    extends: [ Float ]
+    extends: [ Decimal ]
 - concept:
     name: NrMinutes
     description: "Unit of measure for time in minutes"
@@ -29,12 +29,12 @@ ontology:
 - concept:
     name: DegreesLatitude
     type: ValueType
-    extends: [ Float ]
+    extends: [ Decimal ]
     requires: [ DegreesLatitude <= 90, DegreesLatitude >= -90 ]
 - concept:
     name: DegreesLongitude
     type: ValueType
-    extends: [ Float ]
+    extends: [ Decimal ]
     requires: [ DegreesLongitude <= 180, DegreesLongitude >= -180 ]
 - concept:
     name: Polygon
@@ -143,7 +143,7 @@ ontology:
     roles:
     - concept: RunwayLength
     verbalizes:
-    - '{Runway} hase {RunwayLength}'
+    - '{Runway} has {RunwayLength}'
     multiplicity: ManyToOne
   - name: geometry
     roles:
@@ -191,7 +191,6 @@ ontology:
     name: Aircraft
     type: EntityType
     identify_by: [ tailnum ]
-    requires: [ tailnum ]
   relationships:
   - name: serial_number
     roles:
@@ -292,7 +291,7 @@ ontology:
     verbalizes:
     - '{Airport} has average arrival {Delay}'
     multiplicity: ManyToOne
-    derived_by: [ 'Decimal == AVG[Flight.arrival_delay WHERE Airport == Flight.route.destination GROUP BY Airport]' ]
+    derived_by: [ 'Delay == AVG[Flight.arrival_delay WHERE Airport == Flight.route.destination GROUP BY Airport]' ]
   - name: name
     roles:
     - concept: AirportName
--- a/examples/flights.yaml
+++ b/examples/flights.yaml
@@ -2,103 +2,100 @@ version: 0.1.1
 ontologies:
   - name: flights
     description: Ontology for flight-related concepts and relationships
-    components:
-      - name: Example_Runway
-        concept:
+    concepts:
+      - concept:
           name: Example_Runway
           type: EntityType
           identify_by: [ id ]
         relationships:
           - name: id
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_Runway} id {String}' ]
             multiplicity: ManyToOne
           - name: shapelength
             roles:
-              - player: Float
+              - concept: Float
             verbalizes: [ '{Example_Runway} shapeLength {Float}' ]
             multiplicity: ManyToOne
           - name: airportid
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_Runway} airportId {String}' ]
             multiplicity: ManyToOne
           - name: length
             roles:
-              - player: Float
+              - concept: Float
             verbalizes: [ '{Example_Runway} length {Float}' ]
             multiplicity: ManyToOne
           - name: geometry
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_Runway} geometry {String}' ]
             multiplicity: ManyToOne
           - name: designator
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_Runway} designator {String}' ]
             multiplicity: ManyToOne
           - name: airports
             roles:
-              - player: Example_Airport
+              - concept: Example_Airport
             verbalizes: [ '{Example_Runway} airports {Example_Airport}' ]
             multiplicity: ManyToOne
             derived_by: [ 'Example_Runway.airportid == Example_Airport.airportid' ]
-      - name: Example_RouteAlertComment
-        concept:
+      - concept:
           name: Example_RouteAlertComment
           type: EntityType
           identify_by: [ "id" ]
         relationships:
           - name: id
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_RouteAlertComment} id {String}' ]
             multiplicity: ManyToOne
           - name: routealertid
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_RouteAlertComment} routeAlertId {String}' ]
             multiplicity: ManyToOne
           - name: comment
             roles:
-              - player: String
+              - concept: String
             verbalizes: [ '{Example_RouteAlertComment} comment {String}' ]
             multiplicity: ManyToOne
           - name: routealert
             roles:
-              - player: Example_RouteAlert
+              - concept: Example_RouteAlert
             verbalizes: [ '{Example_RouteAlertComment} routeAlerts {Example_RouteAlert}' ]
             multiplicity: ManyToOne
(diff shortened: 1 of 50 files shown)


examples/flights.yaml
```diff
@@ -1,1092 +1,1109 @@
...
+    roles:
+    - concept: RunwayDesignator
+    verbalizes:
+    - '{Runway} uses {RunwayDesignator}'
+    multiplicity: ManyToOne
+  - name: length
+    roles:
+    - concept: RunwayLength
+    verbalizes:
+    - '{Runway} hase {RunwayLength}'
```
khush-bhatia: This should be "has" and not "hase" ? 

How are you thinking about validation for these so that we could catch these typos ?

examples/flights.yaml
```diff
@@ -1,1092 +1,1109 @@
...
+    roles:
+    - concept: RouteId
+    verbalizes:
+    - '{Route} is identified by {RouteId}'
+    multiplicity: OneToOne
+  - name: route_name
+    roles:
+    - concept: String
+    verbalizes:
+    - '{Route} has name- {String}'
```
khush-bhatia: Similar to the comment above, the hyphen after `name` looks like a typo ?

examples/flights.yaml
```diff
@@ -1,1092 +1,1109 @@
...
+    roles:
+    - concept: Manufacturer
+    verbalizes:
+    - '{Model} is manufactured by {Manufacturer}'
+    multiplicity: ManyToOne
+- concept:
+    name: Aircraft
+    type: EntityType
+    identify_by: [ tailnum ]
+    requires: [ tailnum ]
```
khush-bhatia: Are both identify_by and requires needed ? How does one reason about the need for both ?

examples/flights.yaml
```diff
@@ -1,1092 +1,1109 @@
...
+    verbalizes:
+    - '{Airport} centers at {DegreesLongitude}'
+    multiplicity: ManyToOne
+  - name: average_arrival_delay
+    roles:
+    - concept: Delay
+    verbalizes:
+    - '{Airport} has average arrival {Delay}'
+    multiplicity: ManyToOne
+    derived_by: [ 'Decimal == AVG[Flight.arrival_delay WHERE Airport == Flight.route.destination GROUP BY Airport]' ]
```
khush-bhatia: The expression above for `average_departure_delay` says `Delay=` .. but here it says `Decimal = ` ?

examples/flights.yaml
kurtStirewalt: Yes, this is a typo. Unfortunately, with verbalization strings there is not a good automated way to catch these kinds of things.

examples/flights.yaml
kurtStirewalt: This is actually not a typo, but I can see why it looks like one. Languages (like object role modeling) that support verbalization use these hyphens to group text around the roles of the relationships when verbalizing constraints. So, for instance, the verbalization of the many to one multiplicity constraint would read:

   "Each Route has at most one name String"

rather than:

   "Each Route has name at most one String"

Notice how by adding the hyphen after "name" that text floats with the second role (played by String) so that the verbalization of the constraint is more readable.

That said, I can see why folks who aren't familiar with this convention (which admittedly is only supported in a small number of modeling languages) might find it confusing.

What do you think we should do in this case? I will add a comment at the end of the line in the spec to clarify that it is not a typo. Please let me know if you think this suffices.

examples/flights.yaml
kurtStirewalt: Good question. The requirement on tailnum is implied by its use as an identifier. I'll remove it. Thanks

examples/flights.yaml
kurtStirewalt: Good catch. This is one that we should have caught in CI/CD on the RAI side. I'lll look into what failed there. But it's fixed now. Thanks

khush-bhatia (approved):

examples/flights.yaml
khush-bhatia: Yes a comment would be useful to indicate the difference. Thanks.

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
 version: 0.2.0.dev0
-name: flights
-description: Ontology for flight-related concepts and relationships
```
jbonofre, line 3: Maybe we can keep the `description` here?

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+    roles:
+    - concept: RouteId
+    verbalizes:
+    - '{Route} is identified by {RouteId}'
+    multiplicity: OneToOne
+  - name: route_name
+    roles:
+    - concept: String
+    verbalizes:
+    - '{Route} has name- {String}'    # The hyphen after "name" has significance when verbalizing constraints
```
jbonofre, line 551: I see the comment about hyphen after `name`. Thanks for that. I just wonder if we should not have details in the spec that we reference here (not necessarily in this PR, but a follow up).

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
 version: 0.2.0.dev0
-name: flights
-description: Ontology for flight-related concepts and relationships
+name: Flights
```
jbonofre, line 2: `ontology.json` added `requires` at top level. Maybe we should use it in this example.

Something like:

```
  version: 0.2.0.dev0
  name: Flights
  requires:
    - "COUNT[Airport] > 0"         # there must be at least one Airport
    - "COUNT[Carrier] > 0"         # there must be at least one Carrier
  ontology:
    - concept:
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+    roles:
+    - concept: Capacity
+    verbalizes:
+    - '{Aircraft} has {Capacity}'
+    multiplicity: ManyToOne
+  - name: carrier
+    roles:
+    - concept: Carrier
+    verbalizes:
+    - '{Aircraft} carrier {Carrier}'
```
jbonofre: In OSI, `verbalizes` is a natural-language template that reads the relationship out loud, with `{ConceptName}` placeholders for each role.
The "convention" is subject + verb + object.

I would suggest to use:

```
- '{Aircraft} belongs to {Carrier}'
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+    verbalizes:
+    - '{Flight} has arrival {Delay}'
+    multiplicity: ManyToOne
+  - name: cancelled
+    roles:
+    - concept: Boolean
+    verbalizes:
+    - '{Flight} cancelled {Boolean}'
+    multiplicity: ManyToOne
+  - name: cancellationcode
```
jbonofre:
```suggestion
  - name: cancellation_code
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+    roles:
+    - concept: String
+    verbalizes:
+    - '{Flight} cancellationCode {String}'
```
jbonofre:
```suggestion
    - '{Flight} has cancellation code {String}'
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+          expression: FLIGHT.scheduled_departure
+        relationship: scheduled_departure
+      - object_mapping:
+          concept: Boolean
+          expression: FLIGHT.cancelled
+        relationship: cancelled
+      - object_mapping:
+          concept: String
+          expression: FLIGHT.cancel_code
+        relationship: cancellationcode
```
jbonofre:
```suggestion
        relationship: cancellation_code
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+          expression: ROUTE.id
+      children:
+      - object_mapping:
+          concept: Distance
+          expression: ROUTE.distance
+        relationship: distance
+      - object_mapping:
+          concept: Integer
+          expression: ROUTE.dist_grp
+        relationship: distancegroup
```
jbonofre:
```suggestion
        relationship: distance_group
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+    - '{Route} has average arrival {Delay}'
+    multiplicity: ManyToOne
+    derived_by: [ 'Delay == AVG[Flight.arrival_delay WHERE Flight.route(Route) GROUP BY Route]' ]
+  - name: distance
+    roles:
+    - concept: Distance
+    verbalizes:
+    - '{Route} spans {Distance}'
+    multiplicity: ManyToOne
+  - name: distancegroup
```
jbonofre:
```suggestion
  - name: distance_group
```

examples/flights.yaml
```diff
@@ -1,1092 +1,1108 @@
...
+    roles:
+    - concept: Integer
+    verbalizes:
+    - '{Route} distanceGroup {Integer}'
```
jbonofre:
```suggestion
    - '{Route} has distance group {Integer}'
```

jbonofre (approved):
LGTM. I left some suggestions.

examples/flights.yaml
kurtStirewalt: Agreed. If you're ok with it, let's defer this treatment to a follow-up PR, where I will introduce a new section in the md file that discusses uses of verbalization (including verbalizing requires constraints with examples) and introduces this convention. Thanks for the suggestion.

examples/flights.yaml
kurtStirewalt: Yes, thanks. I missed a few of these when beautifying the example (that was previously generated from another artifact). I'll make another manual pass over the entire example to make sure that all of these verbalize properly.

examples/flights.yaml
kurtStirewalt: Yes, good example. I'll include it.

examples/flights.yaml
kurtStirewalt: Added more examples with comments. It should be much clearer now. Will also treat verbalizations more formally in a subsequent PR, per @jbonofre 's recommendation.

examples/flights.yaml
kurtStirewalt: Added. Thanks

examples/flights.yaml
kurtStirewalt: @jbonofre , I took advantage of an opportunity to make this example clearer by introducing a CancelationReason concept (which is typical in other versions of the flight ontology). It fixes this issue in a slightly different way than just changing the name of the relationship, but it's an improvement. Thanks

examples/flights.yaml
kurtStirewalt: Introduced CancelationCode as a first-class concept. The verbalization is now much clearer. Thanks

examples/flights.yaml
kurtStirewalt: Fixed

examples/flights.yaml
kurtStirewalt: Introduced DistanceGroup as a first class concept with a requires constraint to enrich its semantics.

examples/flights.yaml
kurtStirewalt: Fixed

examples/flights.yaml
kurtStirewalt: Introduced DistanceGroup as a first class concept. The verbalization now reads more clearly.

PR #150, opened 2026-06-10
Expression language Doc

sfc-gh-wpugh:
This is an MD file version of [OSI Proposal: Expression Language](https://docs.google.com/document/d/1Jn98kHWsnbvo1MycQBAnWirGTwRmw-DWaen7coNCYWA/edit?usp=sharing) which the Expression Language sub-group put together.

It is a sub-set of ANSI SQL with clarifications on places that ANSII does not define, so that OSI documents are portable.  The dialect name is OSI_SQL_2026.


core-spec/expression_language.md
```diff
@@ -0,0 +1,788 @@
...
+
+**Working Group**
+
+| Lead(s) | Participants |
+| :---- | :---- |
+| Will Pugh, Snowflake Khushboo Bhatia, Snowflake  | LLyod Tabb, Malloy Dianne Wood, Atscale Lior Ebel, Salesforce   Quigley Malcolm, dbt Labs Kurt, Relational AI Justin Talbot, Databricks Pavel Tiunov, Cube Damian Waldron, Thoughtspot Oliver Laslett, Lightdash Martin Traverso, Starburst JB Onofré, The ASF Raul Beiroa, Denodo |
+
+## Overview
+
+![][image1]
```
QMalcolm: This image isn't rendering on github when I view this markdown file ([view page](https://github.com/sfc-gh-wpugh/OSI/blob/e5f3e0ca885758f3591dede6d91d7979b249490a/core-spec/expression_language.md)). I'm trying to think how I did this in the past. I believe instead of embedding the encoding at the bottom of the markdown I instead added the photo as a separate file, and then added the relative link to it in the markdown?

core-spec/expression_language.md
```diff
@@ -0,0 +1,788 @@
...
+
+There are two layers in OSI that need an expression language:
+
+* **Ontology layer.**  This layer maps onto the ontology layer which sits above the logical layer.  It maps more closely to modelling languages like OWL, [(Py)Rel](https://docs.relational.ai) from RelationalAI, and [Legend](https://legend.finos.org) from Goldman Sachs  
+* **Logical layer.**  This layer maps directly to the databases and physical layer.  It maps closely to traditional BI semantic models.
+
+This proposal is only targeted at the Logical Layer.  It would be nice if the Ontological layer could re-use the same expression language, but that will be treated as a separate proposal.
+
+This document defines the SQL expression language subset that OSI-compliant implementations MUST support. The goal is to provide a portable expression language that works across all OSI implementations while allowing vendors to expose richer database-specific functionality through dialect extensions.  In particular, it is meant for expressions at the logical layer.  This means metrics, fields, filters, etc  In the future, expressions such as arbitrary join expressions should also use this expression language.
```
QMalcolm: I think in going forward with this decision we should drop database-specific dialects. In my mind OSI/Ossie is meant to to be a portable standard. The OSI/Ossie SQL expression language essentially enforces that. Allowing for dialects inherently breaks that. Maybe the escape hatch of dialects _is_ necessary, but I think it also fundamentally breaks the contract. However, maybe the various opinions on dialect are controversial enough that what I'm talking about needs it's own proposal 😅

QMalcolm (commented):
Thank you for putting this together ❤️ Still working through it, adding some initial comments as a checkpoint

core-spec/expression_language.md
sfc-gh-wpugh: Yeah.  I agree with this, but would break it into a separate PR.  That may be a more controversial opinion.

My current strategy is to:
1.  define the expression language
2.  then make it default
3.  Try to remove the ability to set dialect at the expression language.  Basically only allow it at the root.  I had a PR a while ago that I proposed this:  [POC for a switch to having a single dialect for the whole file](https://github.com/open-semantic-interchange/OSI/pull/60/changes). 
4. Reduce down to only allowing the OSI expression language

I would hope 1 & 2 are not controversial.  I'm interested in whether 3 is

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+1. **Portability**: Core functions work identically across all implementations  
+2. **Familiarity**: Based on widely-adopted SQL syntax and semantics  
+3. **Analytical Focus**: Prioritizes functions commonly used in BI and analytics  
+4. **Extensibility**: Vendor dialects can extend beyond the core
+
+### Changes to YAML
+
+1) Create a new dialect in the OSI spec: OSI\_SQL\_2026, which refers to this language specification.   
+2) Make OSI\_SQL\_2026 the default dialect if one is not chosen.
```
khush-bhatia: I don't think we should add a new dialect type. At the end, this is still ANSI SQL, why give it a new name ?

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| Arithmetic operators | `+`, `-`, `*`, `/`, `%` (modulo) |
+| Comparison operators | `=`, `<>`, `!=`, `<`, `>`, `<=`, `>=` |
+| Logical operators | `AND`, `OR`, `NOT` |
+| `BETWEEN` | `x BETWEEN a AND b` |
+| `IN` / `NOT IN` | `x IN (a, b, c)` This only supports lists of values, not subqueries. |
+| `LIKE` / `ILIKE` | Pattern matching |
+| `IS NULL` / `IS NOT NULL` | Null checks |
+| `CASE WHEN` | Conditional logic |
+| Aggregate functions | See [Aggregation Functions](https://docs.google.com/document/d/1nvt-vOV8TRKDOlF8C2OkThBqGF73eVZLSC1wryUBPM4/edit#aggregation-functions) |
+| Window functions | See [Window Functions](https://docs.google.com/document/d/1nvt-vOV8TRKDOlF8C2OkThBqGF73eVZLSC1wryUBPM4/edit#window-functions) |
```
khush-bhatia: The doc linked here isn't publicly accessible.

core-spec/expression_language.md
khush-bhatia: Agreed on 1, 2. For 3 I see the value in adding a default dialect but I don't think we should remove the flexibility of being able to capture multiple dialects at expression level. Specially if someone is trying to interoperate between a database specific SQL functions and let's say a  PowerBI  dashboard. There is an argument here that multi-dialect support is a feature of OSI. We could also add translators from DAX formulas to SQL expressions to the OSI codebase to unlock that interoperability. 

In the same vein, I am not sure if 4 is a good idea. I do see value in supporting the multiple database dialects and richer set of SQL features provided by the each of them.

khush-bhatia: To add here, I do believe in standardization of the expression language.  It will be good to brainstorm how we can move towards OSI default dialect covering majority of the cases but still allow flexibility of multiple dialects for the remaining cases.

core-spec/expression_language.md
sfc-gh-wpugh: No.  We are defining a subset and declaring parts that are not defined in ANSI.  

We need to have a name for it.

core-spec/expression_language.md
sfc-gh-wpugh: The problem with having multiple expressions for a single field (through different dialect expressions) is that you have no idea if they even return the same thing.

Then, your portability story is not so great.

If we want to have something like a way to stash the original expression, we should have a field that is something like "original_expression" or "upstream_expression".

core-spec/expression_language.md
sfc-gh-wpugh: Ah.  Yes.  That was one of the old docs.  Getting rid of the reference.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+Accessing an object’s field MUST be qualified with the name of the object in order to reference the field.  E.g. store\_sales.id would reference the ID field in the STORE\_SALES object.
+
+##### Dataset / Object
+
+The object scope is unique to the object the fields are defined in.  Currently, the only objects that have nested fields are Datasets.  They have a fields section to define new fields.
+
+Fields may be defined at the dataset level.  Their identifiers MUST be unique within the dataset, but can have the same name as identifiers in other datasets, or in the global scope.
+
+**Object fields can access logical or physical fields** within the object’s scope without requiring qualification.  The fields may also access global fields as well, which means that shadowing can occur here.  To handle these in a predictable way, names will be resolved with the following rules:
```
khush-bhatia: > The fields may also access global fields as well, which means that shadowing can occur here

This is hard to reason. Is there a use case where this is needed ? 

My preference is that dataset fields shouldn't be able to access global fields.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| `COUNT(x)` | `COUNT(x)` | `COUNT(X)` | `COUNT(x)` |
+| `COUNT(DISTINCT x)` | `COUNTD(x)` | `COUNT_DISTINCT(X)` | `DISTINCTCOUNT(x)` |
+| `AVG(x)` | `AVG(x)` | `AVG(X)` | `AVERAGE(x)` |
+| `MIN(x)` | `MIN(x)` | `MIN(X)` | `MIN(x)` |
+| `MAX(x)` | `MAX(x)` | `MAX(X)` | `MAX(x)` |
+| `STDDEV(x)` | `STDEV(x)` | `STDDEV(X)` | `STDEV.S(x)` |
+| `STDDEV_POP(x)` | `STDEVP(x)` | `STDDEV(X)` | `STDEV.P(x)` |
+| `VARIANCE(x)` | `VAR(x)` | `VARIANCE(X)` | `VAR.S(x)` |
+| `MEDIAN(x)` | `MEDIAN(x)` | `MEDIAN(X)` | `MEDIAN(x)` |
+| `PERCENTILE_CONT(x, 0.75)` | `PERCENTILE(x, 0.75)` | `PERCENTILE(X, 75)` | `PERCENTILE.INC(x, 0.75)` |
```
khush-bhatia, line 678: The function signature in the OSI dialect is defined as
`
| `PERCENTILE_CONT` | `PERCENTILE_CONT(p) WITHIN GROUP (ORDER BY expr)` | Continuous percentile (interpolated) | Holistic |
`

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+### Window Function Mapping
+
+| OSI Standard | Tableau | Looker Studio | DAX |
+| :---- | :---- | :---- | :---- |
+| `ROW_NUMBER() OVER(...)` | `INDEX()` | N/A | `RANKX(...)` with DENSE |
+| `RANK() OVER(...)` | `RANK(expr)` | N/A | `RANKX(...)` |
+| `SUM(...) OVER(PARTITION BY...)` | `{FIXED [...]: SUM(...)}` | N/A (blending only) | Context-dependent |
+| `LAG(x, 1) OVER(ORDER BY...)` | `LOOKUP(x, -1)` | N/A | `CALCULATE(x, PREVIOUSDAY(...))` |
+| `RUNNING_SUM(...)` | `RUNNING_SUM(SUM(...))` | N/A | `CALCULATE(SUM(...), FILTER(...))` |
```
khush-bhatia, line 724: RUNNING_SUM() isn't defined above

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| :---- | :---- | :---- |
+| `IF` | `IF(condition, true_result, false_result)` | Ternary conditional |
+| `IFF` | `IFF(condition, true_result, false_result)` | Alias for IF |
+| `NULLIF` | `NULLIF(expr1, expr2)` | Returns NULL if equal |
+| `COALESCE` | `COALESCE(expr1, expr2, ...)` | First non-null value |
+| `IFNULL` | `IFNULL(expr, default)` | Alias for COALESCE with 2 args |
+| `NVL` | `NVL(expr, default)` | Alias for COALESCE with 2 args |
+| `NVL2` | `NVL2(expr, not_null_result, null_result)` | Different results for null/not-null |
+| `ZeroIfNull` | `ZEROIFNULL(expr)` | Returns 0 if null |
+| `NullIfZero` | `NULLIFZERO(expr)` | Returns NULL if zero |
```
khush-bhatia, line 509: is the case difference here for ZeroIfNull and NullIfZero intentional ?

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
+# OSI Proposal: Expression Language
+
+**Current Status:** Proposed Final 
+**Last Updated:** 4 May Feb 2026
```
khush-bhatia: Remove Feb ?

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| Scalar functions | See function categories below                                                                                                                                                                                                                                                         |
+| Parentheses | Expression grouping                                                                                                                                                                                                                                                                   |
+| Bind parameters | `:parameter_name` syntax                                                                                                                                                                                                                                                              |
```
khush-bhatia: Shouldn't we use ANSI SQL syntax here which is `?`

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+Implementations MAY support additional functions through dialect extensions. These should be documented as dialect-specific.
+
+---
+
+## Version History
+
+| Version | Date | Changes |
+| :---- | :---- | :---- |
+| 0.1 | 2026-05-04 | Initial draft |
```
khush-bhatia: The current version on the repo is `0.2.0.dev0`

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+All identifiers MUST be valid names and follow ANSI SQL naming, with the size limitation of 128 characters for identifiers.  Many databases support longer identifiers, however, this number is safe for a broad number of vendors.
+
+Regular identifiers (unquoted) should be case insensitive.    For example, an identifier id is regular, so it would match with Id or iD.  Comparing quoted and non-quoted identifiers is DB specific, so for best portability it is best to use simple identifiers.
+
+The quote character for the OSI dialect will follow ANSI SQL and support the double quote character (“).  This means that if an expression is in a field expression or as an identifier in the YAML, this will be the expected quoting.  However, there are some databases that use other escape characters.  Working with these have the option of either creating expressions using their dialect or having the OSI document written in the OSI dialect, but then having the SQL Interface queried in the local dialect.  The SQL Interface will be defined in a different document. 
+
+#### Comparison Table
+
+| You type this in SQL | Database sees it as... | Will it match a column created as id? |
```
justint-db: Let's remove the "Database sees it as..." column, which is defining an internal database implementation detail. The only thing that matters to specify here is which identifiers belong to an equivalence class.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| :---- | :---- | :---- |
+| id | ID | **Yes** (Standard behavior) |
+| Id | ID | **Yes** (Standard behavior) |
+| "ID" | ID | **Yes** (Force-matched to normalized case) |
+| "id" | id | **No** (Database is looking for lowercase) |
+
+Sometimes, we may refer to a **normalized identifier**.  This is a form the identifiers can be put in, so they can be matched easily and matches can be made with case-sensitive, exact matching.  For **normalized identifiers**:
+
+* Regular identifiers are upper cased  
+* Quoted identifiers have their quotes stripped and any escaped characters are unescaped
```
justint-db, line 74: Same, how the database choses to normalize is an implementation detail that shouldn't be specified by OSI.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+#### Name Spaces
```
justint-db, line 76: I wonder if this whole section should be pulled out of this doc and moved to the foundational semantics piece. This seems to assume a large amount of context about the broader OSI spec beyond the expression language itself.

We could maybe get by in this doc by just saying that fields may be multi-part and the interpretation of a field reference depends on where it appears in the OSI spec (to be defined later).

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+Physical fields are ONLY accessible from Dataset fields. 
+
+There is no way to create Physical fields.
+
+## SQL Language Subset
+
+### Supported SQL Constructs
+
+OSI expressions support the following SQL constructs within metric and filter expressions:
```
justint-db: Are there other expressions beyond "metric and filter expressions"? If so, what language will those expressions use?

I would have expected that this language could also be used for dimensions. And I have a fairly strong expectation that there isn't another expression language than this one.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+### Operator Precedence
+
+Standard SQL operator precedence applies (highest to lowest):
+
+1. Parentheses `()`  
+2. Unary operators: `+`, `-`, `NOT`  
+3. Multiplication/Division: `*`, `/`, `%`  
+4. Addition/Subtraction: `+`, `-`  
+5. Comparison: `=`, `<>`, `<`, `>`, `<=`, `>=`, `LIKE`, `IN`, `BETWEEN`, `IS NULL`  
+6. `NOT`
```
justint-db: NOT also appears above in (2). Is there a non-unary NOT?

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+5. Comparison: `=`, `<>`, `<`, `>`, `<=`, `>=`, `LIKE`, `IN`, `BETWEEN`, `IS NULL`  
+6. `NOT`  
+7. `AND`  
+8. `OR`
+
+---
+
+## Aggregation Functions
+
+All aggregation functions operate on the effective grain of the metric.
```
justint-db: Cut this and move to foundational semantics doc, since this doc does not define "effective grain".

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| `APPROX_COUNT_DISTINCT` | ✅ | ✅ | ✅ | ❌ (extension) |
+| `APPROX_PERCENTILE` | ✅ | ✅ `APPROX_QUANTILES` | ✅ | ❌ |
+
+**Note**: BigQuery uses `APPROX_QUANTILES(expr, num_buckets)` which returns an array. To get a specific percentile: `APPROX_QUANTILES(amount, 100)[OFFSET(50)]` for median.
+
+**When to use approximate functions:**
+
+- Large datasets (millions+ rows) where exact results aren't critical  
+- Interactive dashboards where response time matters  
+- Exploratory analysis where directional accuracy is sufficient
```
justint-db: Cut? Seems irrelevant to defining the language.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+---
+
+### Conditional Aggregations (REQUIRED)
```
justint-db, line 198: If we're intentionally not requiring or recommending

`SUM(x) FILTER(WHERE ...)`

We should note that here.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+For multi-stage aggregation (see [OSI Analytical Context Extension](https://docs.google.com/document/d/1MKNySGmEv_C6CzBZ7um9Ym3_mMvmOolpDuwPvRzQ1bo/edit?usp=sharing)):
+
+| Category | Functions |
+| :---- | :---- |
+| **Distributive** | SUM, COUNT, MIN, MAX |
+| **Algebraic** | AVG, STDDEV, VARIANCE |
+| **Holistic** | MEDIAN, PERCENTILE, COUNT DISTINCT |
+| **Sketch-based** | APPROX\_COUNT\_DISTINCT, APPROX\_PERCENTILE |
+
+Approximate functions are naturally suited for multi-stage aggregation because their sketch data structures are designed to be mergeable.
```
justint-db: Cut? This seems like it would be good content for a "how to implement" addendum. Not in the main language spec.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| `RADIANS` | `RADIANS(degrees)` | Convert degrees to radians |
+| `DEGREES` | `DEGREES(radians)` | Convert radians to degrees |
+| `PI` | `PI()` | Value of π |
+
+### Comparison Functions (REQUIRED)
+
+| Function | Syntax | Description |
+| :---- | :---- | :---- |
+| `GREATEST` | `GREATEST(x, y, ...)` | Maximum of arguments |
+| `LEAST` | `LEAST(x, y, ...)` | Minimum of arguments |
```
justint-db, line 473: What is the desired null handling on these? I'm not sure what the sql standard is.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| **Distributive** | SUM, COUNT, MIN, MAX |
+| **Algebraic** | AVG, STDDEV, VARIANCE |
+| **Holistic** | MEDIAN, PERCENTILE, COUNT DISTINCT |
+| **Sketch-based** | APPROX\_COUNT\_DISTINCT, APPROX\_PERCENTILE |
+
+Approximate functions are naturally suited for multi-stage aggregation because their sketch data structures are designed to be mergeable.
+
+---
+
+## Date/Time Functions
```
justint-db, line 226: Flagging that Databricks has a number of deviations from the date/time specifications below. Data/time handling is going to be an area of substantial differences across databases.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| `BOOLEAN` | `TRUE`, `FALSE` | Boolean literals |
+| `NOT` | `NOT expr` | Logical negation |
+| `AND` | `expr1 AND expr2` | Logical AND |
+| `OR` | `expr1 OR expr2` | Logical OR |
+
+---
+
+## Window Functions
+
+Window functions operate over a window frame defined by `OVER()`. This should act consistently with window functions in ANSI SQL.
```
justint-db, line 524: We should probably flag that there may be some interaction with the query interface and/or foundational model here.

- Unlike most expressions, window functions in SQL can only appear in certain places in the query syntax. What does that mean for consistency?
- Window functions in SQL cannot change their granularity, but we almost certainly want that. So our window functions aren't totally consistent with sql's.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+| :---- | :---- | :---- |
+| `LAG` | `LAG(expr, offset, default) OVER (...)` | Value from previous row |
+| `LEAD` | `LEAD(expr, offset, default) OVER (...)` | Value from next row |
+| `FIRST_VALUE` | `FIRST_VALUE(expr) OVER (...)` | First value in window |
+| `LAST_VALUE` | `LAST_VALUE(expr) OVER (...)` | Last value in window |
+| `NTH_VALUE` | `NTH_VALUE(expr, n) OVER (...)` | Nth value in window |
+
+### Window Aggregations (REQUIRED)
+
+All standard aggregation functions can be used as window functions:
```
justint-db, line 566: Databricks does not support using COUNT DISTINCT as a window function.

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+-- Percent of total
+amount / SUM(amount) OVER () * 100
+```
+
+---
+
+## Type Conversion Functions
+
+### CAST (REQUIRED)
```
justint-db, line 586: Databricks does not support a number of these casts. This is a place where I think there will also be considerable database variation (both in what casts are supported and in exactly how those casts operate).

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+-- Null-safe comparison (treats NULLs as equal)
+a IS NOT DISTINCT FROM b    -- TRUE if both are NULL
+a IS DISTINCT FROM b        -- TRUE if one is NULL and other isn't
+```
+
+---
+
+## Dialect Extensions
+
+OSI implementations MAY support additional functions through dialect-specific extensions. When using dialect extensions, the expression must specify the dialect.
```
justint-db: Is there a practical way to enforce "the expression must specify the dialect"?

Should consumers of an expression reject it if they see a non-standard expression without a dialect specified?

core-spec/expression_language.md
```diff
@@ -0,0 +1,786 @@
...
+
+The OSI dialect should always be supported.  Other dialects MAY be ignored.  There is no guarantee that all different dialects for an expression will act the same, so implementations should be consistent with their dialect handling.  This means that if an OSI model has an expression written in two dialects, the implementation should deterministically choose which dialect to use.
```
justint-db: This seems overly complicated. I think, if we decide to keep multiple dialects at the expression level, that we should define that this feature is meant to be used for multiple expressions *that mean the same thing*. And that consumers are free to pick whichever one they want to use, since they all mean the same thing.

core-spec/expression_language.md
justint-db: I think we should pull parameterization out into its own doc. It has considerable implications for other parts of OSI:

- How do consumers find out which parameters are available? Can they get metadata like the the names/types/descriptions of the parameters?
- How do consumers query models with parameters?
- How do databases/catalogs store models with parameters? (Most databases don't allow creating assets with unresolved parameters.)

core-spec/expression_language.md
sfc-gh-wpugh: Updating.

core-spec/expression_language.md
sfc-gh-wpugh: Changing wording to be equivalnce.

core-spec/expression_language.md
sfc-gh-wpugh: Yes.  This is added because the current OSI spec allows for multiple dialects for the same expression.  

The approach I was going to do was to give that case a defined behaviour (which is that we cannot garuntee they do the same thing), and the advocate that we get rid of having multiple dialects at the expression level.  I think having a dialect at the file level makes more sense.

However, part of what I'm trying to do is backfill correct semantics for the spec we have.

core-spec/expression_language.md
sfc-gh-wpugh: Yeah.  This makes sense.  I don't think we have them in the core file format yet, so we should not have them here.

core-spec/expression_language.md
sfc-gh-wpugh: I think I got rid of the ones that Databricks does not support.

core-spec/expression_language.md
```diff
@@ -0,0 +1,773 @@
...
+
+Implementations MAY support additional functions through dialect extensions. These should be documented as dialect-specific.
+
+---
+
+## Version History
+
+| Version   | Date       | Changes |
+|:----------|:-----------| :---- |
+| 0.2.0.dev | 2026-06-44 | Initial draft |
```
QMalcolm: 😅 I'm guessing this was supposed to be
```suggestion
| 0.2.0.dev | 2026-06-04 | Initial draft |
```

core-spec/expression_language.md
```diff
@@ -0,0 +1,773 @@
...
+Pattern wildcards for `LIKE`:
+
+- `%` \- Match any sequence of characters  
+- `_` \- Match any single character
+
+### Regular Expressions (RECOMMENDED)
+
+| Function | Syntax | Description |
+| :---- | :---- | :---- |
+| `REGEXP_LIKE` | `REGEXP_LIKE(str, pattern)` | Test if pattern matches |
```
QMalcolm: `REGEXP_LIKE` is in both  `Pattern Matching (REQUIRED)` and `Regular Expressions (RECOMMENDED)` tables. Given it's in both, then I assume the strong constraint holds (e.g. It's `REQUIRED`). Perhaps we should drop it from the `RECOMMENDED` table, or add a note in the `RECOMMENDED` table that that specific regular expression function is actually required?

core-spec/expression_language.md
```diff
@@ -0,0 +1,773 @@
...
+### 
+
+### Not Supported in Expressions
+
+| Construct | Reason |
+| :---- | :---- |
+| `SELECT` / `FROM` / `JOIN` | Handled by semantic layer |
+| `GROUP BY` | Controlled by grain specification |
+| `WHERE` | Use filter property instead |
+| Subqueries | Use field references instead, or EXISTS\_IN() for filtering based on a subquery. |
```
QMalcolm, line 112: Hmmmm I cannot find `EXISTS_IN()` defined in this document. I think we need to either add it's definition, or remove it as a suitable substitute

QMalcolm (commented):
Looking good! I think we have some consistency items to rectify, and a few clarifying details to add, but overall I think we're getting close 🚀

core-spec/expression_language.md
```diff
@@ -0,0 +1,773 @@
...
+| `VARIANCE` | `VARIANCE(expr)` | Sample variance | Algebraic |
+| `VAR_POP` | `VAR_POP(expr)` | Population variance | Algebraic |
+| `VAR_SAMP` | `VAR_SAMP(expr)` | Sample variance (alias for VARIANCE) | Algebraic |
+
+### Percentile Functions (REQUIRED)
+
+| Function | Syntax | Description | Decomposability |
+| :---- | :---- | :---- | :---- |
+| `MEDIAN` | `MEDIAN(expr)` | Median value (50th percentile) | Holistic |
+| `PERCENTILE_CONT` | `PERCENTILE_CONT(p) WITHIN GROUP (ORDER BY expr)` | Continuous percentile (interpolated) | Holistic |
```
onkardahale, line 162: The [Foundation proposal doc ](https://docs.google.com/document/d/1bEuIb2eUpwesi6cIZ5FOkXMQDh7KAPIjZsWchmLgOAk) lists "ordered-set aggregates with WITHIN GROUP as deferred (section 1, 3 table) but this table requires WITHIN GROUP syntax for percentiles. My guess is the deferral only covers the windowed form, since ordered-set aggregates aren't window functions to begin with. If that's right, this doc is fine and the Foundation wording just needs a qualifier. Flagging in case it was meant to be broader.

core-spec/expression_language.md
```diff
@@ -0,0 +1,773 @@
...
+### Window Aggregations (REQUIRED)
+
+All standard aggregation functions can be used as window functions:
+
+```sql
+-- Running total
+SUM(amount) OVER (ORDER BY order_date)
+
+-- Running average
+AVG(amount) OVER (ORDER BY order_date ROWS BETWEEN 6 PRECEDING AND CURRENT ROW)
```
onkardahale, line 573: Line 585 uses 6 PRECEDING AND CURRENT ROW, but that's not one of the three frame patterns you list on 551-553. Either the list is missing a case or the example is wrong. (The [Foundation proposal doc's 6.10.6 ](https://docs.google.com/document/d/1bEuIb2eUpwesi6cIZ5FOkXMQDh7KAPIjZsWchmLgOAk/edit?tab=t.0#heading=h.sdqgv9m5g8p1) just allows any integer bounds, which would cover this

core-spec/expression_language.md
sfc-gh-wpugh: Yes.  But we do need to cover what will be equivalent when matching terms within an OSI document.  Otherwise, we can't provide portabiliyt.

core-spec/expression_language.md
sfc-gh-wpugh: Moved to Semantics doc.  So we can punt that discussion.

core-spec/expression_language.md
sfc-gh-wpugh: Yeah.  Will do.

core-spec/expression_language.md
sfc-gh-wpugh: Yup.  Updated the language to be "any expression"

core-spec/expression_language.md
sfc-gh-wpugh: I think we pushed that out of this initial draft.  Although, I agree we need it.

core-spec/expression_language.md
sfc-gh-wpugh: Oops. Fixed.

core-spec/expression_language.md
sfc-gh-wpugh: Sure.

core-spec/expression_language.md
sfc-gh-wpugh: Yeah.  We should fix that up in the Foundation proposal.  good Catch.

core-spec/expression_language.md
sfc-gh-wpugh: Yup.

core-spec/expression_language.md
sfc-gh-wpugh: Is there a subset that are easy to agree on?

core-spec/expression_language.md
sfc-gh-wpugh: Done

khush-bhatia (dismissed):
LGTM, could you please do a s/OSI/Ossie everywhere ? Including the image... 

This is a great starting point. We can continue to evolve this spec further.

QMalcolm (approved):
Thank you for doing this work, I know it's been a long one. It looks good to me, lets move forward 🙂

khush-bhatia (approved):

jbonofre:
I was about to review this one. LGTM.

PR #153, opened 2026-06-12
Add OrionBelt (OBML) converter

ralfbecher:
## Summary

Adds a bidirectional converter between OSI and OrionBelt's OBML (OrionBelt Markup Language) under `converters/orionbelt/`, following the hub-and-spoke guidance in `converters/index.md`.

- **OBML to OSI**: maps datasets, fields, relationships, and metrics to the OSI core spec. OrionBelt-proprietary constructs (OBML-only filters, settings, owner, refresh, type info, named secondary join paths, many-to-many cardinality) are emitted under the `ORIONBELT` vendor in `custom_extensions` so nothing is dropped.
- **OSI to OBML**: restores OSI-native fields that OBML can't represent (unique keys, field labels, leftover `ai_context`) via an `OSI` vendor stash, so a full OSI to OBML to OSI roundtrip is lossless. Third-party vendor extensions (`SNOWFLAKE`, `DBT`, `SALESFORCE`, `GOODDATA`, ...) are preserved verbatim at every level.
- Also supports the OSI **ontology** export direction.
- ANSI_SQL dialect for field and metric expressions, with the documented fallback behaviour.

## Layout

Mirrors the existing Python converters (`gooddata`, `dbt`):

- `src/osi_orionbelt/` (converter, CLI, vendored OSI/OBML/ontology JSON schemas for self-contained builds)
- `tests/` with 120 tests, including a TPC-DS baseline against `examples/tpcds_semantic_model.yaml` (byte-identical fixture)
- `pyproject.toml`, `README.md`, and OBML <-> OSI mapping analysis docs

```bash
cd converters/orionbelt
uv sync && uv run pytest   # 120 passed
```

CLI (`osi-orionbelt`) exposes `obml-to-osi` and `osi-to-obml` subcommands, mirroring `osi-dbt`.

## Spec impact

None. `vendor_name` is free-form in the core spec, so the `ORIONBELT` vendor tag needs no schema or specification change. This is a converter-only (non-spec) contribution.

## Notes

The canonical converter source is maintained in the [orionbelt-semantic-layer](https://github.com/ralfbecher/orionbelt-semantic-layer) repository (`packages/osi-orionbelt`) and published to PyPI as `osi-orionbelt` (Apache-2.0). Documented limitations (many-to-many joins, named secondary join paths, ontology-layer value concepts) are carried losslessly in `custom_extensions` and listed in the README.


converters/orionbelt/src/osi_orionbelt/converter.py
```diff
@@ -0,0 +1,2772 @@
...
+        """
+        name = field["name"]
+
+        # Get expression (prefer ANSI_SQL dialect)
+        expr_obj = field.get("expression", {})
+        code = name  # fallback
+        if isinstance(expr_obj, dict):
+            dialects = expr_obj.get("dialects", [])
+            for d in dialects:
+                if d.get("dialect") == "ANSI_SQL":
```
jbonofre: Any OSI metric without `ANSI_SQL` dialect is skipped with a warning. Since the converter only emits `ANSI_SQL` on the reverse path, this is fine for self-generated input. However, any third-party OSI model (let's say `SNOWFLAKE`) dialect will lose metric silently on import.
Maybe we should have a clear warning at import.

converters/orionbelt/src/osi_orionbelt/converter.py
```diff
@@ -0,0 +1,2772 @@
...
+                if osi_description:
+                    measure_def["description"] = osi_description
+                if osi_synonyms:
+                    measure_def["synonyms"] = osi_synonyms
+                self._apply_obml_measure_extras(measure_def, obml_extras)
+                measures[name] = measure_def
+                continue
+
+            # Try expression-based: AGG(expr with dataset.column refs)
+            parsed_expr = self._parse_expr_agg(expr_text)
```
jbonofre: Metrics with expressions that don't match either simple-agg, expr-agg, or complex-decompose patterns are warning about then dropped entirely.
I would clearly add a comment about that.

jbonofre (approved):
LGTM, I just left some comments about the fact we are loosing part of the model during convertion (whereas the README implies not 😄 ).

ralfbecher:
Thanks for the careful review, @jbonofre - you were right, and the "we lose part of the model on import" catch was a good one. The root cause was a test blind spot: our corpus only ever round-tripped our own output, which always emits `ANSI_SQL` and only expressions our parser can re-read, so the two import drop paths you spotted were never exercised. A genuinely third-party OSI model (e.g. Snowflake authored, or a non-SQL dialect) hit them and lost metrics with just a warning.

Pushed a fix to this branch that turns both drop paths into preserve-and-warn, plus a bit more:

- **No silent loss.** A metric with no OBML representation (non-SQL dialect only, or an expression that does not decompose into OBML measures/metrics) is no longer dropped. The original OSI metric is preserved verbatim under an `OSI`-vendor model-level `custom_extension` (`obml_unconverted_metrics`) and re-emitted on OBML to OSI, so the OSI to OBML to OSI round trip stays lossless. Each one raises a loud `LOSSY:` warning, since the metric is not queryable through OBML.
- **Clear warning at import.** Exactly your suggestion: the `LOSSY:` warnings surface through the converter's `warnings` list (and the OrionBelt REST API's `conversion_warnings`), so a SNOWFLAKE/MAQL-only model no longer goes quiet.
- **Dialect catching.** Rather than only reading `ANSI_SQL`, the import now reads the first available SQL dialect in the order `ANSI_SQL`, `SNOWFLAKE`, `DATABRICKS` (their aggregations are ANSI-compatible). Non-SQL dialects (`MDX`, `TABLEAU`, `MAQL`) are not parsed as SQL. So a Snowflake- or Databricks-authored aggregation now converts to a real measure instead of being lost.

Two follow-on edge cases got hardened while in here: preserved metrics now contribute their dialects/vendors to the top-level v0.2 `dialects` / `vendors` informational arrays, and a preserved metric whose name a real converted metric later owns is skipped on export so the output has no duplicate metric names (passes `validate_osi`). `convert()` is also now idempotent on both directions.

Docs updated to scope the fidelity claim, and new regression tests cover both former drop paths, the round trip, the metadata arrays, the name collision, and idempotency (134 passing). Thanks again!

ralfbecher:
Follow-up: aligned this branch with #148 (drop non-spec root `dialects`/`vendors`).

While addressing the metric-loss feedback I noticed the OrionBelt converter was emitting root-level `dialects` and `vendors`, which the published core schema forbids (root is `additionalProperties: false`, only `version` + `semantic_model`) - the same non-conformance #148 fixes for the dbt converter. Our vendored schema copy had wrongly allowed them, so it went unnoticed.

The converter now emits neither root array. Dialects stay per-expression in `expression.dialects[]` and vendors per-entity in `custom_extensions`, so no information is lost and a preserved metric still round-trips with its own dialect/vendor intact. The vendored schema is updated to match the published core, and a guard test asserts the root arrays are now rejected.

Note this supersedes the first inline finding above (advertise restored metrics in the top-level `dialects`/`vendors`): per the published schema those arrays should not exist, so the right fix is to drop them rather than populate them. The metadata lives in the schema-valid per-expression / per-entity homes instead.

This also keeps us pointed at the single-document-dialect direction in #52. 135 tests passing.

ralfbecher:
Hi @jbonofre, gentle nudge on this one. Since your approval the diff has grown a fair bit in response to review, so I'd value a fresh look before merge:

- **Metric loss fix** (preserve-and-warn): metrics with no OBML representation are now preserved verbatim in an `OSI`-vendor `custom_extension` and re-emitted, so the OSI to OBML to OSI round trip stays lossless, with loud `LOSSY:` warnings instead of silent drops.
- **Schema alignment with #148**: dropped the non-spec root `dialects`/`vendors`; dialects stay per-expression, vendors per-entity. Vendored schema updated to match the published core, with a guard test.
- **Re-vendor / refactor** (06-23): refreshed the vendored `obml-schema.json` to camelCase-only and split the converter monolith into focused modules.

The PR is `MERGEABLE`/`CLEAN` with all tests passing. If the module split feels like too much to re-review in one pass, I'm happy to land the converter as you originally approved it and move that refactor into a follow-up PR, whichever is easier for you.

@khush-bhatia flagging for merge once the re-approval lands. Thanks both!

ralfbecher:
Rebased on latest `main` and resolved the merge conflict (it was only in `ROADMAP.md`, from the OSI -> Ossie rename). Kept the Ossie wording for the existing converters and preserved the OrionBelt entry.

I also renamed the converter to match the incubation naming used by the other converters (dbt, gooddata): package `ossie_orionbelt`, distribution `apache-ossie-orionbelt`, CLI `ossie-orionbelt`. Consistent with upstream, I left `osi` in the spec-format file and class names (e.g. `osi_to_obml.py`, `OSItoOBML`). All 145 converter tests pass and the PR is mergeable again.

converters/orionbelt/src/ossie_orionbelt/schemas/osi-schema.json
```diff
@@ -0,0 +1,330 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/open-semantic-interchange/OSI/core-spec/osi-schema.json",
+  "title": "OSI Core Metadata Specification",
+  "description": "JSON Schema for validating OSI (Open Semantic Interoperability) semantic model definitions",
```
khush-bhatia: Why make a copy of the schema ? This becomes one more file that needs to be updated for any changes to the spec.


For existing converters,  we are planning to consolidate the java/python and other language specific bindings to be generated from the spec and remove the duplication across different language bindings.

converters/orionbelt/src/ossie_orionbelt/_common.py
```diff
@@ -0,0 +1,88 @@
...
+# brackets). The leading lookbehind prevents matching the tail of a longer path
+# (``a.b.c``) or a mid-token boundary; the bare form must start with a letter or
+# underscore so numeric literals (``1.5``) are never treated as references.
+_COLUMN_REF_RE = re.compile(
+    r'(?<![\w."`\]])'
+    r'(?P<ds>[A-Za-z_]\w*|"[^"]+"|`[^`]+`|\[[^\]]+\])'
+    r"\s*\.\s*"
+    r'(?P<col>[A-Za-z_]\w*|"[^"]+"|`[^`]+`|\[[^\]]+\])'
+)
+_OSI_KNOWN_VENDORS = (
```
khush-bhatia: Where is this used ? Trying to understand the need for maintaining a list of known vendors within the scope of a converter

converters/orionbelt/src/ossie_orionbelt/schemas/osi-schema.json
ralfbecher: Good point, thanks. Rather than keep a copy, the converter now resolves the OSI core schema via a search path (`validation._osi_schema_path`): a vendored copy if present, otherwise `core-spec/osi-schema.json` from the enclosing checkout. I've deleted the vendored `osi-schema.json`, so OSI-core validation now links the single canonical `core-spec/` copy with nothing to keep in sync. I also dropped the vendored `osi-ontology-schema.json` and made `validate_osi_ontology` semantic-only, since ossie ships no ontology schema and the converter emits ontology documents but never consumes them. This lines up with your plan to generate the bindings from the spec. Pushed.

converters/orionbelt/src/ossie_orionbelt/_common.py
ralfbecher: You're right, that list didn't belong here. It was dead: only re-exported, never used. The converter preserves any vendor outside its own internal set (`_INTERNAL_VENDORS`) verbatim in both directions, so there's nothing to enumerate. Removed both `_OSI_KNOWN_VENDORS` and the equally unused `_OSI_KNOWN_DIALECTS`. Pushed.

khush-bhatia (approved):

PR #151, opened 2026-06-12
feat(cli): scaffold ossie CLI

QMalcolm:
## Summary

This is the initial scaffold for the ossie CLI 🙂 None of the commands _do_ anything currently, all the currently planned commands are stubbed out for future PRs to implement. In addition the the basic scaffolding, the PR adds basic tests and CI tooling.

https://github.com/user-attachments/assets/a8ae073e-6efb-41bd-b3c6-c3d846415ebc

PR #154, opened 2026-06-15
feat(cli): define plugin objects and basic plugin listing ability

QMalcolm:
> **Stacked on** #151

## Summary

Define the basic structure of a plugin (plugin.yaml). In addition, add ability to list installed plugins. Note, this _does not_ add the ability to install plugins, their associated setup, or convert invocations. This is simply the initial "how should a plugin be defined and found when it exists".

https://github.com/user-attachments/assets/798f002e-8611-4a7d-a7a4-ce587919084f

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -1,10 +1,13 @@
 BINARY_NAME := ossie
 BUILD_DIR   := dist
 
-.PHONY: build test lint release-dry-run clean
+.PHONY: build install test lint release-dry-run clean
 
 build:
 	go build -o $(BUILD_DIR)/$(BINARY_NAME) .
+
+install:
+	go build -o $(shell go env GOPATH)/bin/$(BINARY_NAME) .
 
 test:
 	go test ./...
--- a/cli/cmd/plugin/list.go
+++ b/cli/cmd/plugin/list.go
@@ -2,17 +2,46 @@ package plugin
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
+	"text/tabwriter"
 
 	"github.com/spf13/cobra"
+
+	"github.com/apache/ossie/cli/internal/ossiedir"
+	"github.com/apache/ossie/cli/internal/plugin"
 )
 
 var listCmd = &cobra.Command{
 	Use:   "list",
-	Short: "List installed and available plugins",
-	RunE:  runPluginList,
+	Short: "List installed plugins",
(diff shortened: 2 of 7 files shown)


QMalcolm:
This is a stacked PR

- https://github.com/open-semantic-interchange/OSI/pull/151 
  - https://github.com/open-semantic-interchange/OSI/pull/154 (👈🏻 you are here)
    - https://github.com/open-semantic-interchange/OSI/pull/155
      - https://github.com/open-semantic-interchange/OSI/pull/156
        - https://github.com/open-semantic-interchange/OSI/pull/158

khush-bhatia (dismissed):
This is great, thanks.

One question, should we add a "name" field for plugin metadata ?

QMalcolm:
> This is great, thanks.
> 
> One question, should we add a "name" field for plugin metadata ?

@khush-bhatia I agree that names are important! Also, thank you for calling this out 🙂 Right now the `name` of a plugin is buried under `platform`, which itself has two properties `platform.name` and `platform.vendor`. I think your question highlights that my burying of these properties was perhaps not the best idea 😅 I think both of these can probably be pushed up to the top level and `platform` can go away.

khush-bhatia:
> > This is great, thanks.
> > One question, should we add a "name" field for plugin metadata ?
> 
> @khush-bhatia I agree that names are important, and thank you for calling this out 🙂 Right now the `name` of a plugin is buried under `platform`, which itself has two properties `platform.name` and `platform.vendor`. I think your question highlights that my burying of these properties was perhaps not the best idea 😅 I think both of these can probably be pushed up to the top level and `platform` can go away.

Agreed on having plugin name and vendor being top level.

QMalcolm:
> > > This is great, thanks.
> > > One question, should we add a "name" field for plugin metadata ?
> > 
> > 
> > @khush-bhatia I agree that names are important, and thank you for calling this out 🙂 Right now the `name` of a plugin is buried under `platform`, which itself has two properties `platform.name` and `platform.vendor`. I think your question highlights that my burying of these properties was perhaps not the best idea 😅 I think both of these can probably be pushed up to the top level and `platform` can go away.
> 
> Agreed on having plugin name and vendor being top level.

`name` has been moved to top level, and `platform` became an optional string field (what `vendor` was). I think `platform` might be better terminology than `vendor`. I feel like "vendor" might imply "authored by" whereas "platform" feels like a more abstract "to be used with" 🙂

khush-bhatia (approved):

PR #155, opened 2026-06-15
feat(cli): add plugin invocation protocol

QMalcolm:
> **Stacked on** #154

## Summary

- Adds `Request`, `Response`, and `Issue` JSON envelope types for the plugin stdin/stdout protocol
- Adds `Invoke(ctx, pluginDir, invoke, req, pluginStderr)` — runs a plugin subprocess, pipes the request JSON to stdin, captures stdout as the response, and enforces a context timeout
- A non-empty `Issues` slice in the response is not a Go error; the caller (defined in a later segment of work) will decide whether error-severity issues produce a non-zero exit code

### DIFF
--- a/cli/internal/plugin/invoke.go
+++ b/cli/internal/plugin/invoke.go
@@ -1,0 +1,74 @@
+package plugin
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os/exec"
+)
+
+// Request is the JSON envelope sent to a plugin on stdin.
+type Request struct {
+	Files map[string]string `json:"files"`
+}
+
+// Response is the JSON envelope received from a plugin on stdout.
+type Response struct {
+	Files  map[string]string `json:"files"`
+	Issues []Issue           `json:"issues,omitempty"`
+}
+
+// Issue is a single diagnostic emitted by a plugin during conversion.
+// Severity is one of "error", "warning", or "info".
+// Path is optional; it is omitted for issues not tied to a specific location.
+type Issue struct {
+	Severity string `json:"severity"`
+	Message  string `json:"message"`
+	Path     string `json:"path,omitempty"`
+}
+
+// Invoke runs a plugin subprocess, pipes req as JSON to its stdin,
+// and decodes its stdout as a JSON Response.
+//
+// pluginDir is used as the working directory for the subprocess so that
+// relative paths in the invoke array resolve correctly.
+// invoke is the command and its arguments (invoke[0] is the executable).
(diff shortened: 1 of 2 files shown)


QMalcolm:
This is a stacked PR

- https://github.com/open-semantic-interchange/OSI/pull/151 
  - https://github.com/open-semantic-interchange/OSI/pull/154 
    - https://github.com/open-semantic-interchange/OSI/pull/155 (👈🏻 you are here)
      - https://github.com/open-semantic-interchange/OSI/pull/156
        - https://github.com/open-semantic-interchange/OSI/pull/158

cli/internal/plugin/invoke.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	reqJSON, err := json.Marshal(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal plugin request: %w", err)
+	}
+
+	var stdout bytes.Buffer
+	cmd := exec.CommandContext(ctx, invoke[0], invoke[1:]...)
+	cmd.Dir = pluginDir
```
khush-bhatia, line 61: Would this fail nicely if pluginDir is not accessible ? Would be good to add error handling for this case.

khush-bhatia:
LGTM, but some good follow ups for future for anyone interested:

1. It will be great to support plugins directly reading the semantic model files.
2. The execution contract for these plugins is something we should continue to evolve as support for query API lands. 

Should be we track these as Issues ?

khush-bhatia (approved):

cli/internal/plugin/invoke.go
QMalcolm: Currently, when the pluginDir is not accessible the failure looks like
```bash
plugin process failed: chdir /nonexistent/path/xyz: no such file or directory
```

Additionally it's gets wrapped in a normal Go error instead of becoming a panic. Considering that, I think we can say it fails gracefully 🙂

QMalcolm:
> LGTM, but some good follow ups for future for anyone interested:
> 
>     1. It will be great to support plugins directly reading the semantic model files.
> 
>     2. The execution contract for these plugins is something we should continue to evolve as support for query API lands.
> 
> 
> Should be we track these as Issues ?

I definitely think we should track (1) as an issue. I'm not sure exactly how we want to do this long term. Having the CLI handle read/write means that each plugin doesn't need to implement handing that with the OS, and ensures that glob based file lookup is always handled the same regardless of plugin used. However for sufficiently large semantic layers, it will probably be meaningfully impactful if the plugins handle file system read/write themselves.

For item (2), if we have specific areas (like protocol spec for query API), those would probably be good to create issues to track 🙂

PR #170, opened 2026-06-26
New validation for to/from column length and PK/UK/Join field referen…

stephen-thibeault-dbt:
This PR 2 concepts to the validation script:

- Checks that the to_cols and from_cols lists are the same length when defining relations
- Checks that all referenced columns in relationships, unique keys, and primary keys are listed as fields

It also adds 9 tests for the newly added validations.


khush-bhatia:
The spec does not require this : "Checks that all referenced columns in relationships, unique keys, and primary keys are listed as fields"

PR #172, opened 2026-07-01
python: accept free-form custom extension vendors

onkardahale:
Custom extension vendor names are free-form strings in the current spec and JSON schema, but the Python package still types OSICustomExtension.vendor_name as the OSIVendor enum. A document that is valid against osi-schema.json, e.g. a custom extension with vendor_name: ACME, is rejected by the Python model.

#91 relaxed vendor_name to a free-form string in the spec and schema but did not update the Python package, so the package is now stricter than the spec it implements.

Change vendor_name to str. OSIVendor stays as well-known constants, and since it is a str enum, comparisons like vendor_name == "DBT" still work. Also fix the converter guide, which still tells contributors to add new vendors to a core enum.

Does not touch root dialects/vendors (#148 covers that). Verified locally that the old model rejects vendor_name: ACME and the new one accepts and round-trips it.

### DIFF
--- a/converters/README.md
+++ b/converters/README.md
@@ -281,7 +281,7 @@ Given the [TPC-DS example](../examples/tpcds_semantic_model.yaml) included in th
 
 To add support for a new vendor:
 
-1. Add the vendor to the `vendors` enum in the [core specification](../core-spec/spec.md) if not already present.
+1. Use a stable `vendor_name` string in each custom extension emitted by the converter.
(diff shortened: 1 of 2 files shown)


khush-bhatia:
@onkardahale Because of the rename, there are some merge conflicts, could you please resolve that.

onkardahale:
@khush-bhatia  Fixed it. Merge looks clean now

khush-bhatia (approved):

PR #194, opened 2026-07-14
Fix models elements min length

MonkeyCanCode:
Looking at https://github.com/apache/ossie/blob/main/core-spec/osi-schema.json, `from_columns`, `to_columns`, and `datasets` should not allow empty list  (which is the current behavior) as their min is `1` based on the spec:
```
    "Expression": {
      "type": "object",
      "description": "Expression definition with multi-dialect support",
      "properties": {
        "dialects": {
          "type": "array",
          "items": {
            "$ref": "#/$defs/DialectExpression"
          },
          "minItems": 1
        }
      },

...

    "Relationship": {
      "type": "object",
      "description": "Foreign key relationship between datasets",
...
        "from_columns": {
          "type": "array",
          "items": {
            "type": "string"
          },
          "minItems": 1,
          "description": "Foreign key columns in the 'from' dataset"
        },
        "to_columns": {
          "type": "array",
          "items": {
            "type": "string"
          },
          "minItems": 1,
          "description": "Primary/unique key columns in the 'to' dataset"
        },

...

    "SemanticModel": {
      "type": "object",
      "description": "Top-level container representing a complete semantic model",
...
        "datasets": {
          "type": "array",
          "items": {
            "$ref": "#/$defs/Dataset"
          },
          "minItems": 1,
          "description": "Collection of logical datasets"
        },
...
```


python/src/ossie/models.py
```diff
@@ -155,7 +155,7 @@ class OSISemanticModel(BaseModel):
     name: str
     description: Optional[str] = None
     ai_context: Optional[OSIAIContext] = None
-    datasets: list[OSIDataset]
+    datasets: list[OSIDataset] = Field(..., min_length=1)
```
xiedeyantu, line 158: Can we leave datasets empty?

MonkeyCanCode: From the spec, it should be min 1. Any reason to keep it empty?

xiedeyantu: Will we reserve some empty datasets? I'm not entirely sure, but if you're sure about this, I think it's fine.

MonkeyCanCode: Per the JSON schema (https://github.com/apache/ossie/blob/main/core-spec/osi-schema.json#L302), `datasets` has `"minItems": 1`. A semantic model with no datasets is meaningless. Also, as models are frozen, so we can't do incremental build case as well. I think keep the min of 1 here is more semantic correct as the spec itself doesn't allow empty.

xiedeyantu (approved):

jbonofre:
@MonkeyCanCode can you please rebase and re-check whether `min_length=1` is still the right call given converters that legitimately produce empty lists?

PR #205, opened 2026-07-14
chore(python): migrate to PEP 639 license metadata

esadek:
Migrate to the [PEP 639](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files) license metadata format. Replace the legacy `[project.license]` table form with the SPDX expression form (`license = "Apache-2.0"`).


python/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
MonkeyCanCode: Extra new line here.

python/pyproject.toml
```diff
@@ -24,6 +24,8 @@ name = "apache-ossie"
 version = "0.2.0.dev0"
 description = "Python types for the Apache Ossie semantic model specification"
 requires-python = ">=3.11"
+license = "Apache-2.0"
+license-files = ["LICENSE"]
```
MonkeyCanCode: I think we will need NOTICED file as well here.

esadek: Added

python/LICENSE
esadek: The Apache-2.0 license includes that newline. See: https://apache.org/licenses/LICENSE-2.0.txt

MonkeyCanCode (dismissed):

jbonofre (changes requested):
We should use the root `LICENSE` and `NOTICE` files for now.

jbonofre:
@esadek can you please rebase?

PR #206, opened 2026-07-15
orionbelt converter: fix round-trip fidelity + validation robustness

ralfbecher:
Mirrors [ralfbecher/orionbelt-semantic-layer#201](https://github.com/ralfbecher/orionbelt-semantic-layer/issues/201) (the canonical converter source) into `converters/orionbelt/`. Three pre-existing bugs found while reviewing #153:

## Metric round-trip when a data object's code differs from its display name
`OSItoOBML` resolved metric references only by dataset **name**, but the OBML -> OSI emitter writes physical codes (e.g. `SUM(fact_orders.amount)`), so such metrics were dropped as unconverted on the return trip. Resolution now indexes datasets/fields by **both** name and physical code (source-table code + field expression, quoted or bare). Resolved refs are bracket-quoted so display names containing spaces stay intact through the `dataset.column` parsers, and Snowflake/Databricks quoted identifiers resolve too.

## Dimension name collision across datasets
`_extract_dimensions` keyed dimensions by bare field name, so `Orders.date` and `Invoices.date` collapsed to one. The later one is now qualified with its data object and a warning is recorded - no silent loss.

## `validate_osi` robustness
It assumed every `datasets`/`fields` value was a list of dicts and raised `AttributeError` on malformed input. It now guards each level and returns schema errors instead.

## Tests
Adds `tests/test_osi_converter_roundtrip_robustness.py`. 154 converter tests pass; ruff clean; mypy clean with the dev extra.

Canonical change: ralfbecher/orionbelt-semantic-layer#202 (merged).

---
_Follow-up converter fixes (canonical #228 dimension-name restoration, #229 N-dimensions-per-column) are mirrored in a separate PR, not this one._

### DIFF
--- a/converters/orionbelt/src/ossie_orionbelt/osi_to_obml.py
+++ b/converters/orionbelt/src/ossie_orionbelt/osi_to_obml.py
@@ -20,6 +20,12 @@ from ossie_orionbelt._common import (
     OSI_TO_OBML_TYPE,
 )
 
+# A dataset/column identifier in a resolved metric expression: either a bare SQL
+# word or a bracket-quoted token. ``_resolve_column_refs`` bracket-quotes any
+# canonical name that is not a bare word (e.g. a display name with spaces) so the
+# downstream ``dataset.column`` parsers below never split it on whitespace.
+_RESOLVED_IDENT = r"(?:\w+|\[[^\]]+\])"
+
 
 class OSItoOBML:
     """Convert an OSI semantic model YAML to OBML format."""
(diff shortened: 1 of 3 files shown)


converters/orionbelt/src/ossie_orionbelt/osi_to_obml.py
```diff
@@ -578,7 +584,23 @@ def _extract_dimensions(self, datasets: list) -> dict:
...
                             pass
                         break
-                dimensions[field_name] = dim_def
+                # Dimension names must be unique across the model. When the same
+                # field name occurs in more than one data object (e.g. Orders.date
+                # and Invoices.date), qualify the later one with its data object
+                # instead of silently overwriting the earlier dimension.
+                key = field_name
+                if key in dimensions and dimensions[key].get("dataObject") != ds_name:
+                    key = f"{ds_name} {field_name}"
```
khush-bhatia, line 593: For the key field, Is the use of space here instead of `.` intentional  ?  Wouldn't it better to use the SQL identifier style and use `.` ?

ralfbecher: Good question. It comes down to what "a dimension" fundamentally is in each model:

- **OSI is field-centric** - a dimension *is* a field (a physical column), so a dot (`table.column`) is the natural SQL-identifier spelling.
- **OBML is name-centric** - a dimension is a *named view* onto a column. Its identity is the name; the physical reference lives in the dimension's `dataObject` + `column` pair, not the name. OBML names are display-grade by convention and convenience (`Country Name`, `Total Sales`), not SQL identifiers, so a dot here would make a display name read like a physical reference.

That's why the qualifier uses a space: the key becomes a *name*, not an identifier. It also only fires as a fallback - when two datasets share a bare field name and there's no OBML-origin name to prefer - so it stays rare.

khush-bhatia (approved):

PR #209, opened 2026-07-15
Make the spec's own examples validate against osi-schema.json

Pawansingh3889:
Four small fixes so the examples in core-spec agree with the rules osi-schema.json already states:

1. The Semantic Model section example used `datasets: []`, which violates the schema's `minItems: 1`; it now carries one minimal dataset.
2. Both metric examples (`total_revenue`, `avg_orders`) wrote `expression:` as a bare list, omitting the required `dialects` wrapper that the Expression Object section directly above them defines. They now match the schema and the Complete Example.
3. The Complete Example presented a full document without the required top-level `version` key; it now begins with `version: 0.2.0.dev0`.
4. `spec.yaml`'s commented `primary_key` examples showed a nested list (`- [customer_id]`) while the schema defines `primary_key` as a flat array of strings; the comments now show the flat form used by the schema and by spec.md's own inline examples.

Verification: extracted each fixed example and ran it through `validation/validate.py` (JSON Schema, name uniqueness, relationship references, sqlglot SQL checks) — all pass, including the Complete Example verbatim as a full document.

Found while building an OSI exporter for a governed semantic-layer tool; happy to adjust scope if the section examples are intentionally fragments.

### DIFF
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -83,7 +83,9 @@ semantic_model:
     description: Sales and customer analytics model
     ai_context:
       instructions: "Use this model for sales analysis and customer insights"
-    datasets: []
+    datasets:
+      - name: orders
+        source: sales.public.orders
     relationships: []
     metrics: []
     custom_extensions:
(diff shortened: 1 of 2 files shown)


jbonofre (approved):
Good catch. @khush-bhatia ok for you?

khush-bhatia (approved):

PR #221, opened 2026-07-17
docs: align spec yaml ai_context shape

### LINKED ISSUE #141: core-spec/spec.yaml declares `ai_context: string` but the JSON schema + canonical example use a structured object
## Summary

There is an inconsistency between the three authoritative representations of the `ai_context` field in OSI:

| Source | Says `ai_context` is | Permalink |
|---|---|---|
| `core-spec/spec.yaml` | a **string** (literally `ai_context: string` at every level — semantic_model, datasets, fields, metrics) | [spec.yaml#L39, #L102, #L181, #L210](https://github.com/open-semantic-interchange/OSI/blob/main/core-spec/spec.yaml) |
| `core-spec/osi-schema.json` (`$defs/AIContext`) | `oneOf: [string, object{instructions, synonyms, examples, …, additionalProperties: true}]` | [osi-schema.json](https://github.com/open-semantic-interchange/OSI/blob/main/core-spec/osi-schema.json) |
| `examples/tpcds_semantic_model.yaml` (canonical example) | a **structured object** with `instructions` (on `semantic_model`) and `synonyms` (on datasets/fields/metrics) — never a plain string | [tpcds_semantic_model.yaml](https://github.com/open-semantic-interchange/OSI/blob/main/examples/tpcds_semantic_model.yaml) |
| `docs/index.md` | "Optional annotations at every level … including instructions, synonyms, and example queries" | [docs/index.md](https://github.com/open-semantic-interchange/OSI/blob/main/docs/index.md) |

The JSON schema (which the example references via `# yaml-language-server: $schema=...`) is the most permissive view and is internally consistent with the canonical example. The spec.yaml is the more restrictive (and easier-to-misread) view — it's the one most readers will look at first.

## Why this matters

Converters being built against the spec right now (we're working on a Databricks Metric View ↔ OSI converter as part of an OSI standardization effort with several vendors) have to pick one form. If we read `spec.yaml` literally, we treat `ai_context` as a free-text string and have to encode synonyms/instructions inside it — which loses structure and breaks round-tripping with vendors that have first-class equivalents. If we read the JSON schema, we use the object form — but then the spec.yaml file is technically misleading.

The discrepancy is the kind of thing that bites first-time implementers because spec.yaml is the most discoverable source and reads as a complete reference.

## Suggested resolution

Tighten `core-spec/spec.yaml` so it matches the JSON schema. Concretely, at each of the five sites where `ai_context: string` appears, replace with something like:

```yaml
# Optional: Additional context for AI tools.
# Can be either:
#   - A free-form string, or
#   - A structured object with optional keys:
#       instructions: string  # how AI should use this entity
#       synonyms:     []      # alternative names / terms
#       examples:     []      # sample questions or use cases
#     (additionalProperties: true — vendors may add more keys)
ai_context: {}  # see AIContext in osi-schema.json
```

That keeps the JSON schema as the source of truth, and the YAML spec accurately summarises it.

## Happy to PR this

If maintainers agree on the direction, happy to put up a PR aligning `spec.yaml` to the JSON schema and bumping the dev-version commentary. Let me know which form you'd prefer for the inline doc.

CooperSheroy:
## Summary
- update core-spec/spec.yaml so ai_context is shown as a structured object while noting simple strings are also accepted
- add the relationship ai_context field that already exists in the JSON schema and spec docs

Closes #141.

## Validation
- ruby -e 'require "yaml"; YAML.load_stream(File.read("core-spec/spec.yaml")); puts "YAML parse OK"'
- jq empty core-spec/osi-schema.json

### DIFF
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -53,8 +53,15 @@ semantic_model:
     # Optional: Human-readable description of the semantic model
     description: string
 
-    # Optional: Additional context for AI tools (e.g., custom prompts, instructions)
-    ai_context: string
+    # Optional: Additional context for AI tools.
+    # Can be a simple string or a structured object with instructions,
+    # synonyms, examples, or vendor-specific keys.
+    ai_context:
+      instructions: string
+      synonyms:
+        - string
+      examples:
+        - string
 
     # Required: Collection of logical datasets (fact and dimension tables)
     # See Logical Dataset section below for detailed structure
@@ -117,7 +124,13 @@ datasets:
 
     # Optional: Additional context for AI tools (e.g., synonyms, common terms)
     # Helps LLMs understand the business meaning and generate better queries
-    ai_context: string
+    # Can also be provided as a simple string.
+    ai_context:
+      instructions: string
+      synonyms:
+        - string
+      examples:
+        - string
 
     # Optional: Row-level calculations for grouping, filtering, and in metric expressions
     # See Fields section below for detailed structure
@@ -160,6 +173,15 @@ relationships:
     #   - [order_id, line_number]    # Composite key
     to_columns: []  # Array of column names
 
+    # Optional: Additional context for AI tools.
+    # Can be a simple string or a structured object.
+    ai_context:
+      instructions: string
+      synonyms:
+        - string
+      examples:
+        - string
+
     # Optional: Vendor-specific attributes for extensibility
     custom_extensions:
       - vendor_name: string  # Free-form string identifying the vendor
@@ -196,7 +218,13 @@ fields:
 
     # Optional: Additional context for AI tools (e.g., synonyms, business terms)
     # Helps LLMs understand the field meaning and generate better queries
-    ai_context: string
+    # Can also be provided as a simple string.
+    ai_context:
+      instructions: string
+      synonyms:
+        - string
+      examples:
+        - string
 
     # Optional: Vendor-specific attributes for extensibility
     custom_extensions:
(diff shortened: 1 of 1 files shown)


jbonofre (approved):
LGTM.

@khush-bhatia @QMalcolm as it's a spec change, can you please do a pass? Thanks!

core-spec/spec.yaml
```diff
@@ -53,8 +53,15 @@ semantic_model:
     # Optional: Human-readable description of the semantic model
     description: string
 
-    # Optional: Additional context for AI tools (e.g., custom prompts, instructions)
-    ai_context: string
+    # Optional: Additional context for AI tools.
+    # Can be a simple string or a structured object with instructions,
+    # synonyms, examples, or vendor-specific keys.
+    ai_context:
```
khush-bhatia, line 59: This change is making the yaml version of the spec strictly only allow three subfields which are instructions, synonyms and examples.

Instead we shoudl change the type here to be `{}` so it can support arbitrary key values . 

The issue https://github.com/apache/ossie/issues/141 also suggests the same thing. See 
```
# Optional: Additional context for AI tools.
# Can be either:
#   - A free-form string, or
#   - A structured object with optional keys:
#       instructions: string  # how AI should use this entity
#       synonyms:     []      # alternative names / terms
#       examples:     []      # sample questions or use cases
#     (additionalProperties: true — vendors may add more keys)
ai_context: {}  # see AIContext in osi-schema.json
```

khush-bhatia (changes requested):

core-spec/spec.yaml
jbonofre: Good point. Agree!

jbonofre (changes requested):

kayemkim:
For anyone triaging: main has carried this alignment since #350 (September 9), with `ai_context: {}` and a schema pointer at each node rather than the expanded example shape proposed here. Unless there is something in this diff that #350 missed, it can probably be closed.

jbonofre:
@kayemkim @CooperSheroy I'm closing this PR for now. Feel free to re-open it if needed.

PR #222, opened 2026-07-17
Adds a SqlGlot implementation of the OSI SQL dialect

willpugh:
## Summary

Implements the expression language from the Ossie spec.  Also, found a spec oversight, that caused an incompatibility with ANSII around NOT operator precedence.  Having ANSII compliance was a clear requirement of the expression langauge group.

Updates the spec and implements the language as a  SqlGlot dialect.

### Specification
- [x] Breaking changes to the spec are clearly called out in the summary

### Tests

### Compliance


python/ossie/src/ossie/models.py
```diff
@@ -15,14 +15,14 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from enum import Enum
-from typing import Any, Optional, Union
+from enum import StrEnum
+from typing import Any
```
khush-bhatia: Why the changes to this file `models.py` ? They look unrelated to the SQL dialect implementation.

python/ossie/src/ossie/models.py
```diff
@@ -15,14 +15,14 @@
 # specific language governing permissions and limitations
 # under the License.
```
khush-bhatia, line 17: How about we rename this to `python/src/ossie-model` ? Right now it reads like /ossie/src/ossie which is confusing.

python/ossie/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "apache-ossie"
```
khush-bhatia, line 23: ossie-model ?

python/ossie-sql/src/ossie_sql/dialect.py
```diff
@@ -0,0 +1,205 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""SQLGlot dialect for the Ossie expression language.
+
+Implements the grammar defined in ``core-spec/expression_language.md``
+("Ossie_SQL_2026"). Registers with SQLGlot as ``"ossie"``, so
+``sqlglot.parse_one(sql, read="ossie")`` parses the spec's SQL subset and
+``expression.sql(dialect="ossie")`` renders it back.
```
khush-bhatia, line 23: The dialect name in the json spec and in the consumers should then just be "ossie" ?

khush-bhatia (commented):
Looks good , just some comments about some unrelated changes. 

Thank for this, this is a great direction to standardize on the expression language.

python/ossie-sql/src/ossie_sql/dialect.py
willpugh: I was going to update the json spec, etc. as a different PR.  I'm trying to keep this one to just the SqlGlot dialect.

python/ossie/pyproject.toml
willpugh: I don't think we want to export a package called ossie-model.  When a user installs ossie, I think it should be apache-ossie not ossie-model.

So, i think we probably want to keep this.

python/ossie/src/ossie/models.py
willpugh: Yeah.  It is making the type system better, but not strictly needed in this pr.  I can remove it.

python/ossie/src/ossie/models.py
willpugh: No, we should stick with more cannonical Python pathing.  Although, the naming will match what we come up with in [Ossie Project Development Guidelines](https://docs.google.com/document/d/1_yk45ZiaSVp4lU5mf4YTP8qigcdsElbytR7_3NlqBTo/edit?usp=sharing)

Currently, that proposes core/python/ossie and core/python/ossie-sql

However, I was going to wait until that got settled before doing the change.

python/pyproject.toml
```diff
@@ -15,33 +15,8 @@
...
-
-[tool.hatch.build.targets.wheel]
-packages = ["src/ossie"]
-
-[tool.uv]
-required-version = ">=0.9.0"
+# uv workspace root: shares one dependency resolution/lockfile across the
+# member packages below. This file is not itself a Python package -- see
+# each member's own pyproject.toml for its package metadata.
+[tool.uv.workspace]
```
kunaljubce, line 21: I am of the opinion that we should have a separate PR to introduce uv workspace, to have clean separation of concerns and also easier reverts. Curious what others think on this?

willpugh: I think I need some pyproject to build this code.  This one is not at the root of the project, just at the root of the code I'm working on.

PR #223, opened 2026-07-17
orionbelt converter: dimension round-trip fidelity (name restoration, N-per-column)

ralfbecher:
Follow-up to #206. Mirrors two further canonical converter fixes from ralfbecher/orionbelt-semantic-layer into `converters/orionbelt/`. Both build on #206's `_extract_dimensions` work, which is why they land now that #206 has merged.

## Restore the OBML dimension name (canonical #228)

The OSI field name is the physical column code, so a round trip renamed every dimension to its code and could trip #206's collision fallback for names that were unique in the source. The name is now preserved in an `obml_dimension_name` extension and restored on import, so the space-qualified collision fallback is genuinely **foreign-OSI only** - for an OBML-origin model it no longer fires (this is the concern @khush-bhatia raised on #206). Non-string values are ignored so a foreign payload cannot crash the converter on an unhashable dict key.

## Preserve N dimensions over one column (canonical #229)

OBML allows many dimensions on one column (grain variants, role-playing); OSI is field-centric (one dimension per field). The converter used to drop every dimension past the first, silently. Extras are now preserved in an `obml_extra_dimensions` list extension with a fidelity **warning** (warn, not raise) and rebuilt on import. Both primary and extra dimensions carry their own synonyms and vendor extensions, with shape guards on the opaque extension data.

## Also

Adds the ASF license header to `test_osi_converter_roundtrip_robustness.py`, which was missing it (the sibling converter files have it).

## Tests

`tests/test_osi_converter_roundtrip_robustness.py` (dimension name restoration, N-per-column, synonym/extension preservation, malformed-payload guards) plus updated `test_osi_converter_properties.py`. **177 converter tests pass; ruff clean.**

Canonical changes: ralfbecher/orionbelt-semantic-layer#228, #229 (both merged).

### DIFF
--- a/converters/orionbelt/src/ossie_orionbelt/obml_to_osi.py
+++ b/converters/orionbelt/src/ossie_orionbelt/obml_to_osi.py
@@ -405,11 +405,24 @@ class OBMLtoOSI:
             ext_data["obml_comment"] = col_obj["comment"]
         if col_obj.get("owner"):
             ext_data["obml_owner"] = col_obj["owner"]
-        # Preserve OBML-only dimension properties (timeGrain, format, resultType, etc.)
+        # Preserve OBML-only dimension properties (timeGrain, format, resultType, etc.).
+        # OBML allows N dimensions over one column (grain variants, role-playing
+        # via ``via``); OSI is field-centric (one dimension per field). The first
+        # match is the primary dimension carried on the field; any extras are
+        # preserved as descriptors so the reverse trip can rebuild them instead of
+        # dropping them silently.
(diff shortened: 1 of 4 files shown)


ralfbecher:
Closing: this no longer applies cleanly (files were renamed in #288, and #396 reworks the same code), and it was framed as a mirror of a downstream repo, which isn't how this converter is maintained anymore. The two fixes (dimension name restoration, N dimensions per column) are still missing on `main`. I'll re-submit them as a fresh PR on top of #396 once that lands.

PR #224, opened 2026-07-17
[OSSIE][DATABRICKS] Add Databricks Unity Catalog Metric View converter

Haoranli503:

Adds a bidirectional, offline converter between Apache Ossie semantic models and Databricks [Unity Catalog Metric Views](https://docs.databricks.com/aws/en/metric-views/) (YAML v1.1), filling the Databricks spoke in the hub-and-spoke architecture (`DATABRICKS` is already a listed vendor in `converters/index.md`). Validated against the live Databricks Metric View engine.

## What's included

- **Export** (`ossie-databricks export`): Apache Ossie semantic model -> Metric View (one fact `source` + a nested `joins` tree, flattened `dimensions`, `measures`).
- **Import** (`ossie-databricks import`): Metric View -> Apache Ossie. Metric-View-only features (filter, window, format, rely, cardinality, parameters, materialization) are preserved in `custom_extensions[DATABRICKS]`, so **`MV -> Apache Ossie -> MV` is lossless**.
- A unified `ossie-databricks` CLI and a string-in / string-out Python API.

## Design highlights

- **Round-trip fidelity** via the `custom_extensions[DATABRICKS]` stash -- the approach the converter guide recommends.
- **Join tree, not flat joins.** The Apache Ossie relationship graph is reassembled into the Metric View's nested fact + `joins` tree; a dataset reached by multiple paths (a diamond) is fanned out into one aliased join per path; cyclic graphs are rejected.
- **Cardinality from direction.** `many_to_one` / `one_to_many` is derived from the relationship `from`/`to` direction relative to a selectable grain (`--source`). Multiple facts are supported only narrowly -- when they share a conformed dimension (named as the source); a general galaxy schema (facts each with their own dimensions) or facts sharing no dimension can't map to one Metric View and are **rejected with a clear error**, never silently mis-converted.
- **Key bridge.** A `primary_key`/`unique_keys` whose columns cover a join key becomes `rely: {at_most_one_match: true}`, and vice versa.
- **On export**, joined-table columns are alias-qualified in dimension/measure expressions; fact columns in measures are emitted bare (the DBR idiom).
- **Fails loudly, never silently.** A join condition an Apache Ossie relationship can't represent (a non-equi or cross join), a cyclic or ambiguous-grain graph, and an unsupported version are rejected with a clear `ConversionError`; anything dropped (a foreign-vendor extension, an Apache-Ossie-only annotation, a non-DATABRICKS/ANSI_SQL dialect) emits a warning naming the object.

## Conventions followed

- Lives at `converters/databricks/`, packaged like the dbt/gooddata spokes: `pyproject.toml` (hatchling, Apache-2.0), `src/ossie_databricks/`, `tests/`.
- Every source file carries the ASF license header, matching the sibling spokes.
- The only runtime dependency is PyYAML; Python 3.11+.

## Testing

```bash
cd converters/databricks
python -m pytest tests/
```

Example-based unit tests plus **Hypothesis property-based round-trip tests** in both directions, with a normalized TPC-DS fixture as a baseline.

## Notes

- Foreign-vendor `custom_extensions` have no slot in the Metric View YAML; they are dropped on export with a warning.

## Credits

@jackstein21 is credited as a co-contributor (see the commit's `Co-authored-by` trailer), including the property-based round-trip testing approach.

## Related Issues

Will close #178, Ossie <-> Databricks converters will be tracked here.

### Specification

### Ontology

### Converters
- [X] New converters include tests under the converter's test directory

### Validation

### Documentation

### Examples

### Tests
- [X] All existing tests pass (`pytest` / CI green)
- [X] New functionality is covered by tests

### Compliance
- [X] ASF license headers are present on all new source files
- [X] No third-party dependencies are added without PMC/IPMC approval

### DIFF
--- a/converters/databricks/src/ossie_databricks/_common.py
+++ b/converters/databricks/src/ossie_databricks/_common.py
@@ -1,0 +1,288 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Shared helpers for the Apache Ossie <-> Databricks Metric View converters.
+
+Both directions are pure offline YAML transforms. The only cross-cutting concerns
+live here: version constants, the dialect preference order, the `custom_extensions`
+stash protocol, and small SQL-string helpers.
+"""
+
+import json
+import re
+
+import yaml
+
+# Apache Ossie semantic model spec version this converter targets (see core-spec).
+#
+# NOTE: this is an exact-match check (see convert_ossie_to_metric_view). Unlike the
+# dbt converter, this spoke intentionally has no `apache-ossie` package dependency, so
+# nothing updates this automatically -- it MUST be bumped in lockstep with the
+# `version` in `core-spec/` whenever the spec version moves, or the converter will
+# reject otherwise-valid Apache Ossie files.
+OSSIE_VERSION = "0.2.0.dev0"
+
+# Databricks Unity Catalog Metric View YAML version. Only 1.1 supports joins,
+# per-column comments, synonyms, and the format/window/parameters surface.
+MV_VERSION = "1.1"
+
+# Vendor id used for the `custom_extensions` stash and for dialect selection.
+VENDOR = "DATABRICKS"
+
+# Expression dialects this converter understands, in preference order.
+DIALECT_DATABRICKS = "DATABRICKS"
+DIALECT_ANSI = "ANSI_SQL"
+
+# Metric Views cap the number of synonyms per column.
+SYNONYM_LIMIT = 10
+
+# Bump when the shape of a stashed `data` blob changes.
+STASH_VERSION = 1
+
+# Metric View join cardinalities (the only two values v1.1 defines). Apache Ossie has no
+# cardinality field; the value is implied by relationship direction -- `from` is the
+# many side, `to` is the one side -- so the converter derives it from / writes it
+# into the from/to orientation rather than relying on a dedicated field.
+CARD_MANY_TO_ONE = "many_to_one"
+CARD_ONE_TO_MANY = "one_to_many"
+
+# Model-level stash key recording which dataset was the Metric View `source` (its
+# grain). Needed only when a one_to_many join puts the source on a relationship's
+# `to` side, where the natural FK-sink heuristic would otherwise pick the wrong
+# fact on re-export. Absent for plain many-to-one stars, so they stay clean.
+STASH_SOURCE_KEY = "source_dataset"
+
+# A bare SQL identifier (single column reference), e.g. `c_name`. Used to decide
+# whether an expression can be safely alias-prefixed on export / de-prefixed on
+# import.
+_IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+
+class ConversionError(Exception):
+    """Raised when an input cannot be converted."""
+
+
+def require(obj, key, what):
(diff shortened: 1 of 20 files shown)


Haoranli503:
RE @jbonofre 's review on #178:
1, Renamed OSI to Ossie, done. Two "OSI"s in the ossie converter I left as-is on purpose:

`$schema=.../core-spec/osi-schema.json` — points at a real file in upstream apache/ossie (never renamed; all sibling spokes use this exact name). Renaming it breaks the schema link.

`"License :: OSI Approved :: Apache Software License"` — this OSI = Open Source Initiative, not Ossie. Fixed PyPI classifier, only valid spelling for Apache-2.0; editing it fails packaging.

Renaming either would be a bug, not a cleanup.

2, Duplicate dimension/measure name is now a `ConversionError`, done. Previously a name collision (a dimension/measure sharing a name, case-insensitively) only emitted a warning and still wrote the duplicate to the output. It now raises a ConversionError naming the offending field, consistent with the rest of the converter — the user is told exactly what to rename before handing a broken file to Databricks. Added tests for both the duplicate-dimension and measure-collides-with-dimension cases.

converters/databricks/src/ossie_databricks/_common.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def read_stash(obj):
+    """Return the DATABRICKS stash dict on an Apache Ossie object, or {} if absent.
+
+    The `_v` version marker is stripped from the returned dict.
+    """
+    for ext in (obj or {}).get("custom_extensions") or []:
+        if ext.get("vendor_name") == VENDOR:
+            data = json.loads(ext.get("data") or "{}")
```
jbonofre: If someone edits (by hand) an Apache Ossie file and produces invalid JSON in a `custom_extensions[DATABRICKS].data` field, this raises a raw `json.JSONDecodeError` traceback rather than a `ConversionError`.

Maybe worth to wrap like this:


```suggestion
           try:
                data = json.loads(ext.get("data") or "{}")
            except json.JSONDecodeError as e:
                raise ConversionError(f"DATABRICKS custom_extensions data is not valid JSON: {e}") from e    
```

converters/databricks/src/ossie_databricks/_common.py
```diff
@@ -0,0 +1,278 @@
...
+stash protocol, and small SQL-string helpers.
+"""
+
+import json
+import re
+
+import yaml
+
+# Apache Ossie semantic model spec version this converter targets (see core-spec).
+OSSIE_VERSION = "0.2.0.dev0"
```
jbonofre, line 37: The version check in `convert_ossie_to_metric_view` is exact match, so this converter will silently reject every Apache Ossie file the moment the spec version moves. 

The dbt converter avoids this by depending on the `apache-ossie` package. This converter intentionally has no such dependency (that's ok), but there is no comment flagging that this constant must be updated in lockstep with `core-spec`.

I would suggest a comment to "track" this.

converters/databricks/tests/fixtures/tpcds_ossie.yaml
```diff
@@ -0,0 +1,89 @@
...
+        custom_extensions:
+          - vendor_name: DATABRICKS
+            data: '{"format": {"type": "currency", "currency_code": "USD"}}'
+      - name: total_quantity
+        expression:
+          dialects: [{dialect: DATABRICKS, expression: SUM(ss_quantity)}]
+        description: Total units sold
+    custom_extensions:
+      - vendor_name: DATABRICKS
+        data: '{"filter": "ss_net_profit > 0"}'
```
jbonofre: `write_stash` always injects `"_v": 1`. The equivalent `fixtureB_ossie.yaml` has `"_v": 1, ...` in all its blobs. The missing marker means `tpcds_ossie.yaml` is inconsistent with what the converter actually writes. The `canon()` helper in `_util.py` parses JSON to dicts (masking the discrepancy), but a contributor comparing hand-authored fixture output to real converter output will see a difference.

jbonofre (commented):
I think the `json.JSONDecodeError` leak is worth to fix.

converters/databricks/src/ossie_databricks/_common.py
Haoranli503: Ah good catch, fixed!

converters/databricks/src/ossie_databricks/_common.py
Haoranli503: Added a comment

converters/databricks/tests/fixtures/tpcds_ossie.yaml
Haoranli503: right, we need to be aligned, fixed!

jbonofre (approved):
That's a great contribution! Thanks!

PR #226, opened 2026-07-17
chore(cli): add gofmt format target to Makefile

esadek:
- Add a `format` target that runs `gofmt -w .` and register it in `.PHONY`.
- Apply gofmt alignment to the const block in `ossiedir.go`.


QMalcolm (approved):
Looks good to me! Thank you for doing this work ❤️

PR #229, opened 2026-07-18
feat: add DUCKDB dialect and bidirectional DuckDB converter

michellepellon:
# Add DUCKDB dialect and a bidirectional DuckDB converter

## What

This PR adds DuckDB as a supported expression dialect and contributes a reference converter for it, in three parts:

1. **Dialect registration** — adds `DUCKDB` to the `Dialect` enum in `core-spec/osi-schema.json`, `core-spec/spec.yaml`, the dialect table in `core-spec/spec.md`, and the `OSIDialect` / `OSIVendor` enums in `python/src/ossie/models.py`.
2. **Validation support** — maps `DUCKDB` to sqlglot's `duckdb` dialect in `validation/validate.py`, so DuckDB expressions get full syntax validation (sqlglot's DuckDB support is first-class; no skip-list entry needed).
3. **Converter** — `converters/duckdb/`, a bidirectional Ossie ↔ DuckDB converter following the hub-and-spoke pattern in `converters/README.md`:
   - **Export**: Ossie model → DuckDB SQL (one view per dataset with field expressions, `COMMENT ON` from descriptions, one view per metric with joins derived from declared relationships). Dialect selection prefers `DUCKDB`, falls back to `ANSI_SQL` with a warning.
   - **Import**: DuckDB database → Ossie YAML via `information_schema` and `duckdb_constraints()` (primary keys, unique keys, and foreign keys → relationships; table/column comments → descriptions).
   - CLI (`ossie-duckdb export|import`), pytest suite with round-trip tests executing generated SQL against in-memory DuckDB, and schema validation of imported models against `osi-schema.json`.

## Why

DuckDB (and MotherDuck, its managed cloud service) is widely used as an analytics engine in lakehouse and embedded-analytics stacks, and is a common substrate for the AI-agent workloads that `ai_context` targets. Because DuckDB connections are passed through, the converter works against MotherDuck databases (`md:` connection strings) unchanged.

## Notes

- No changes to existing converters; unknown-dialect handling elsewhere is unaffected (existing fallback chains ignore dialects they don't target).
- The example models in `examples/` are untouched and still validate.
- Aware of #52 (single dialect per document); the enum addition is orthogonal — whichever way that lands, DuckDB needs to be a valid dialect value.

### DIFF
--- a/converters/duckdb/README.md
+++ b/converters/duckdb/README.md
@@ -1,0 +1,98 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Apache Ossie ↔ DuckDB Converter
+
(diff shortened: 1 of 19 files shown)


jbonofre:
@michellepellon do you mind to rebase and resolve conflicts?

PR #230, opened 2026-07-18
add semantido<>ossie converter 

hikarilabs:
## Summary

Converts between semantido a code-native semantic layer authored as decorators on SQLAlchemy models and
Apache Ossie semantic model documents.

Unlike file-to-file converters, semantido's source of truth is live Python: the forward direction imports a module of decorated models, syncs the semantic layer, and emits Ossie YAML through the typed `apache-ossie`
objects. The reverse direction is code generation: it produces a Python module of `@semantic_table`-decorated models from an Ossie document.

### Converters

### Compliance


converters/semantido/src/ossie_semantido/semantido_to_ossie.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+        return None
+
+    left_table, left_col, right_table, right_col = match.groups()
+    # Normalize so from/to match the declared direction regardless of
+    # which side of the equality each table appeared on.
+    if left_table == rel.from_table:
+        from_columns, to_columns = [left_col], [right_col]
+    else:
+        from_columns, to_columns = [right_col], [left_col]
```
jbonofre, line 176: The `else` branch silently assumes `right_table == rel.from_table`. If the join condition references tables that don't match either `rel.from_table` or `rel.to_table` (aliases, schema names, ...), the columns are swapped with no issue recorded. The regex already handles the "doesn't parse" case. This internal case should probably be:

```
elif right_table == rel.from_table:
    from_columns, to_columns = [right_col], [left_col]
else:
    issues.append(ConverterIssue(ConverterIssueType.RELATIONSHIP_JOIN_UNPARSED, ...))
    return None
```

converters/semantido/src/ossie_semantido/osi_to_semantido.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+
+import json
+from typing import List
+
+from ossie import OSIDataset, OSIDocument, OSIField
+
+from ossie_semantido.converter_issues import ConverterIssue, ConverterIssueType, ConverterResult
+
+VENDOR_NAME = "SEMANTIDO"
```
jbonofre: nit: we could have vendor name in a `constant.py` or something as `VENDOR_NAME` is defined in both `osi_to_semantido` and `semantido_to_osi`.

converters/semantido/src/ossie_semantido/semantido_to_ossie.py
hikarilabs: fixed, thank you for spotting that one !

converters/semantido/src/ossie_semantido/osi_to_semantido.py
hikarilabs: done, moved into constants.py | could be an enum from the main library itself in the future.

PR #237, opened 2026-07-20
Create the initial compliance suite

willpugh:
## Summary

This creates the initial harness and some initial tests for the compliance suite

### Tests

### Compliance

### DIFF
--- a/compliance/ARCHITECTURE.md
+++ b/compliance/ARCHITECTURE.md
@@ -1,0 +1,344 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Ossie Compliance Suite — Architecture
+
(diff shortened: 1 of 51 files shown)


compliance/foundation/tests/cross_grain/moderate/t-005a-single-step-sum/model.yaml
```diff
@@ -0,0 +1,37 @@
+name: f_prelude_model
```
christianeu-db, line 1: We have a few example models in the repo. Perhaps, those could be augmented with data as default compliance-suite assets. Some more complex cases may involve additional models but, when building relationships at Tableau, it was really helpful to have 2-3 canonical models that we used for unit tests and reasoning through specific semantics (e.g. filter flow, join types) from both a query shape / data perspective.

This could also let us build Layer 2 / 3 compliance tests in a way that shows more clearly how these layers stack (e.g. a Layer 3 test could show both a result-set and SQL in terms of Layer 2 measure support)

compliance/ARCHITECTURE.md
```diff
@@ -0,0 +1,344 @@
...
+├── harness/                 # engine-agnostic runner — shared by every suite
+│   ├── README.md
+│   ├── pyproject.toml        # workspace member: osi_compliance_harness
+│   └── src/harness/
+│       ├── runner.py         # discovery, adapter invocation, orchestration, CLI
+│       ├── db_manager.py     # in-memory DuckDB: load fixtures, execute SQL
+│       ├── result_compare.py # order-insensitive, numeric-tolerant row compare
+│       ├── reporter.py       # failures.csv + summary.md + console summary
+│       ├── models.py         # TestCase / TestResult / SuiteResult dataclasses
+│       └── proposals_check.py# CI gate: metadata references a known proposal ID
```
christianeu-db, line 85: Harness seems useful for both Layer 2 & Layer 3 compliance suites. In other words, we can validate a Layer 2/3 query against the same model using the same expected file format

christianeu-db (commented):
The harness is very topical (we are planning to add Layer 2 compliance tests as well as the Layer 3 tests you have here). I wonder if it makes sense to split out the Layer 3 specific affordances and land the harness first?

PR #192, opened 2026-07-14
Add WisdomAI domain converter (both directions)

jbonofre:
## What

Adds `converters/wisdom`, a bidirectional converter between [WisdomAI](https://wisdom.ai) domain exports (format `1.0`, the JSON produced by wisdom's `exportDomain` API and consumed by `importDomain`) and Ossie semantic model YAML. It follows the layout and conventions of the dbt converter (`apache-ossie` dependency, `ConverterResult`/`ConverterIssue` loss reporting, argparse CLI). Also registers `WISDOM` as a well-known custom-extension vendor in the spec documents and `OSIVendor` enum.

```bash
ossie-wisdom wisdom-to-osi -i domain-export.json -o semantic_model.yaml
ossie-wisdom osi-to-wisdom -i semantic_model.yaml -o domain-export.json
```

## Import mapping (wisdom → Ossie)

| Ossie | Wisdom |
|-------|--------|
| `semantic_model[].name` / `description` | domain `ref.name` / `description` |
| `ai_context` (model level) | `domainSystemInstructions` + domain knowledge contents as a bulleted list |
| `datasets[]` | tables (`location.database.schema.dbTable` → `source`; `primaryKey.columns` or `isPrimaryKey` flags → `primary_key`) |
| `datasets[].fields[]` | columns (bare, dialect-quoted names) and formulas (verbatim expressions); `displayName` → `label`; DATE/DATETIME/TIMESTAMP → `dimension.is_time` |
| `relationships[]` | relationship graph edges — cardinality folded into edge direction (`from` = many side); compound AND-joins flattened to composite key arrays; OR/non-equi joins dropped with a warning |
| `metrics[]` | per-table measures hoisted to the model level, cross-table name collisions resolved by table-name prefixing |

Expressions are emitted verbatim under the Ossie dialect mapped from the table's connection (`snowflake → SNOWFLAKE`, `databricks → DATABRICKS`); other dialects fall back to `ANSI_SQL` verbatim with an `UNSUPPORTED_DIALECT` warning.

## Export mapping (Ossie → wisdom)

The exact inverse, so round-trips are stable in both directions:

- Model `ai_context` splits back into `domainSystemInstructions` (leading text) + one knowledge item per `- ` bullet.
- A field whose expression is just its own (possibly quoted) name becomes a **column**; anything else becomes a **formula**. `dimension.is_time` maps to a `TIMESTAMP` data type (wisdom re-derives exact types from the warehouse).
[...]

PR #239, opened 2026-07-21
Add WisdomAI domain converter (both directions)

mwu-wisdom:
Supersedes #192, which was closed during the repository migration (the old repo is now `apache/ossie-temp`, and its head points at a stale mirror branch). This is the same converter rebased onto the current `main`, with two adaptations to changes that landed since:

- BigQuery connections now map to the new `BIGQUERY` dialect (with backtick identifier quoting) instead of falling back to `ANSI_SQL` with a warning.
- The converter `pyproject.toml` follows the repo's UV conventions (`dependency-groups`, `tool.uv.sources` for the in-repo `apache-ossie`).

## What

Adds `converters/wisdom`, a bidirectional converter between [WisdomAI](https://wisdom.ai) domain exports (format `1.0`, the JSON produced by wisdom's `exportDomain` API and consumed by `importDomain`) and Ossie semantic model YAML. It follows the layout and conventions of the dbt converter (`apache-ossie` dependency, `ConverterResult`/`ConverterIssue` loss reporting, argparse CLI). Also registers `WISDOM` as a well-known custom-extension vendor in the spec documents and `OSIVendor` enum.

```bash
ossie-wisdom wisdom-to-osi -i domain-export.json -o semantic_model.yaml
ossie-wisdom osi-to-wisdom -i semantic_model.yaml -o domain-export.json
```

## Import mapping (wisdom → Ossie)

| Ossie | Wisdom |
|-------|--------|
| `semantic_model[].name` / `description` | domain `ref.name` / `description` |
| `ai_context` (model level) | `domainSystemInstructions` + domain knowledge contents as a bulleted list |
| `datasets[]` | tables (`location.database.schema.dbTable` → `source`; `primaryKey.columns` or `isPrimaryKey` flags → `primary_key`) |
| `datasets[].fields[]` | columns (bare, dialect-quoted names) and formulas (verbatim expressions); `displayName` → `label`; DATE/DATETIME/TIMESTAMP → `dimension.is_time` |
| `relationships[]` | relationship graph edges — cardinality folded into edge direction (`from` = many side); compound AND-joins flattened to composite key arrays; OR/non-equi joins dropped with a warning |
| `metrics[]` | per-table measures hoisted to the model level, cross-table name collisions resolved by table-name prefixing |

Expressions are emitted verbatim under the Ossie dialect mapped from the table's connection (`snowflake → SNOWFLAKE`, `databricks → DATABRICKS`, `bigquery → BIGQUERY`); other dialects fall back to `ANSI_SQL` verbatim with an `UNSUPPORTED_DIALECT` warning.

## Export mapping (Ossie → wisdom)

The exact inverse, so round-trips are stable in both directions:

- Model `ai_context` splits back into `domainSystemInstructions` (leading text) + one knowledge item per `- ` bullet.
- A field whose expression is just its own (possibly quoted) name becomes a **column**; anything else becomes a **formula**. `dimension.is_time` maps to a `TIMESTAMP` data type (wisdom re-derives exact types from the warehouse).
- Relationships become `MANY_TO_ONE` edges; the `ai_context` notes written by the import direction restore `ONE_TO_ONE`/`MANY_TO_MANY`, and composite keys become compound `AND` join conditions.
- Metrics attach to the table their expression references (fallback to the first dataset with a warning).
- IDs are derived deterministically from element names so re-runs are reproducible; connections are per-dialect placeholders expected to be remapped at import.
- Not representable in wisdom — dropped with warnings: semantic models beyond the first, `unique_keys`, `custom_extensions`, and `ai_context` on fields/metrics (plus synonyms/examples anywhere).

## Testing

- 27 tests against a synthetic fixture covering all four cardinalities, compound AND-join flattening, OR-join dropping, dialect fallback, hidden columns, stale measures, duplicate fields, metric name collisions, full round-trip equality (`OSI → wisdom → OSI` yields an identical `OSIDocument`), and determinism of the emitted export.
- Verified end-to-end against two real wisdom domain exports (Snowflake and BigQuery connections): **both outputs pass `validation/validate.py` fully** (schema, uniqueness, reference integrity, sqlglot SQL parsing), and `OSI → wisdom → OSI` round-trips are **byte-identical** YAML for both.


converters/wisdom/src/ossie_wisdom/cli.py
```diff
@@ -0,0 +1,116 @@
...
+from pathlib import Path
+
+import yaml
+
+from ossie import OSIDocument
+from ossie_wisdom.converter_issues import ConverterIssueType
+from ossie_wisdom.osi_to_wisdom import OSIToWisdomConverter
+from ossie_wisdom.wisdom_to_osi import WisdomToOSIConverter
+
+_ISSUE_REASON: dict[ConverterIssueType, str] = {
```
jbonofre, line 37: The dict is manually kept in sync with `ConverterIssueType`. Any new type added to the enum without a corresponding entry will crash the CLI at runtime with a `KeyError`. 

I suggest to use `.get()` with a fallback:

```
reason = _ISSUE_REASON.get(issue.issue_type, issue.issue_type.value)
```

converters/wisdom/src/ossie_wisdom/osi_to_wisdom.py
```diff
@@ -0,0 +1,394 @@
...
+        for metric in model.metrics or []:
+            elements.append((metric.name, metric.custom_extensions))
+        for name, extensions in elements:
+            if extensions:
+                issues.append(
+                    ConverterIssue(issue_type=ConverterIssueType.CUSTOM_EXTENSION_DROPPED, element_name=name)
+                )
+
+    def _infer_dataset_dialect(self, dataset: OSIDataset) -> OSIDialect:
+        for field in dataset.fields or []:
```
jbonofre, line 151: We only inspect fields here, not formulas.

A dataset with no columns (only formulas) silently falls back to `ANSI_SQL` regardless of what dialect the formula was written in. This is an edge case but could produce a subtly wrong export without any warning. 

Checking formula expressions for a vendor dialect would make this more robust.

converters/wisdom/src/ossie_wisdom/osi_to_wisdom.py
```diff
@@ -0,0 +1,394 @@
...
+_WISDOM_DIALECT = {
+    OSIDialect.SNOWFLAKE: "snowflake",
+    OSIDialect.DATABRICKS: "databricks",
+    OSIDialect.BIGQUERY: "bigquery",
+    OSIDialect.ANSI_SQL: "ansi",
+}
+
+
+def _stable_id(prefix: str, value: str) -> str:
+    return f"{prefix}_{hashlib.md5(value.encode('utf-8')).hexdigest()}"
```
jbonofre: On FIPS-compliant systems, MD5 raises `ValueError` by default. Since this is for ID generation (not security), adding `usedForSecurity=False` makes the intent clear and avoids the crash:


```suggestion
    return f"{prefix}_{hashlib.md5(value.encode('utf-8'), usedForSecurity=False).hexdigest()}"
```

converters/wisdom/src/ossie_wisdom/osi_to_wisdom.py
mwu-wisdom: I may be conflating concepts between Wisdom and Ossie but there doesn't seem to be a formula construct in Ossie? At least there doesn't seem to be one on the OSIDataset class https://github.com/apache/ossie/blob/8a73a38fafbf3c78ef9f7570d07ba8ace6fafa91/python/src/ossie/models.py#L110-L122

If we're referring to Wisdom formulas, they are converted into dataset fields in the Wisdom -> OSI conversion already, so they would be discoverable here.

converters/wisdom/src/ossie_wisdom/cli.py
mwu-wisdom: Good catch — fixed in 63f125b using `.get()` with the enum value as the fallback reason, so a missing entry degrades to a terse-but-correct warning instead of a `KeyError`.

converters/wisdom/src/ossie_wisdom/osi_to_wisdom.py
mwu-wisdom: Agreed and fixed in 63f125b. One note: the Python keyword is lowercase `usedforsecurity` (the camelCase `usedForSecurity` from the suggestion raises `TypeError`), so I applied it with the correct spelling plus a comment explaining it's for FIPS-enabled builds.

converters/wisdom/src/ossie_wisdom/osi_to_wisdom.py
jbonofre: You are right, sorry I was confused during the review with Wisdom formulas 😄

jbonofre (approved):

PR #247, opened 2026-07-22
[OSSIE][NVIDIA_GSF] Add bidirectional GSF semantic model converter

liofnv:
## Summary

Adds an offline, bidirectional converter between Apache Ossie YAML and NVIDIA GSF's semantic model YAML.

- `ossie-gsf export`: Ossie → NVIDIA GSF
- `ossie-gsf import`: NVIDIA GSF → Ossie
- Python string-in/string-out API
- No GSF, Neo4j, database, or network connection required to convert
- NVIDIA-specific metadata uses the `NVIDIA_GSF` custom-extension identifier

## Why

GSF keeps its active semantic layer in Neo4j. It now supports exporting and importing that layer as a standalone YAML document, which gives Ossie a portable representation to convert against — versionable and reviewable, with no live GSF deployment involved.

## Mapping

The converter emits and consumes GSF's `GsfModelDocument` contract, whose roots are `data_layer`, `semantic_layer`, and `zones`.

| Ossie | GSF |
| --- | --- |
| Dataset | Term, plus its catalog table |
| Physical field | `columns_attributes` entry |
| Computed field | `sql_attributes.manual` entry |
| Model-level metric | `custom_analyses` entry |
| Relationship | `joins`, `foreign_keys`, and `semantic_fks` |

Ossie-origin entities get deterministic UUIDv5 identifiers so repeated conversions are stable. A GSF-origin document keeps its live identifiers, catalog properties, SQL text, and zones, so a GSF → Ossie → GSF cycle is lossless. Ossie edits stay authoritative: preserved SQL and relationship records are reused only while they still match the Ossie model.

## Testing

- 33 converter tests, including round-trip, validation, and error cases
- Checked-in Ossie/GSF fixture pair
- Generated Ossie YAML validated with this repository's official schema validator
- Converter output validated against GSF's own schema model
- Verified against a live GSF deployment: a real exported model converts in both directions with no identifiers lost or invented, imports into a running GSF instance, and a second import is a no-op

## Notes

The physical catalog must already be ingested in the target GSF, and that database needs a configured connection, since GSF validates imported SQL against the connection's dialect.

Known limitations are documented in `converters/gsf/README.md`, including one represented table per term and the loss of Ossie composite unique keys, which GSF records per column.

The GSF-side import/export implementation has merged: https://github.com/NVIDIA/GSF/pull/134


johntomcat7408-cmyk:
Superseded by #275 (https://github.com/apache/ossie/pull/275). The replacement preserves the original contribution context, consolidates the change into one commit, and includes the review fixes from this PR.

Update: #275 has been closed. #247 remains the active PR; please disregard the previous supersession note.

liofnv:
#275 is not a supersession of this PR. It's a copy of this branch's code re-committed under different authorship, based on an earlier state of this PR.

This PR was updated in 791b440 to target GSF's actual model interchange contract (`GsfModelDocument`). #275 predates that change and still emits the earlier provisional shape, which current GSF cannot import. I've verified the code here end to end against a running GSF instance: a real model exports, converts in both directions, and re-imports with every record and identifier intact.

This remains the active contribution and I'll keep it current. Happy to fold in the CI workflow idea from #275.

converters/gsf/src/ossie_gsf/native_converter.py
```diff
@@ -0,0 +1,1937 @@
...
+
+
+def _parse_ossie(value: str) -> tuple[dict[str, Any], dict[str, Any]]:
+    root = _load_yaml(value, "Ossie")
+    unknown = sorted(set(root) - {"version", "semantic_model"})
+    if unknown:
+        raise GSFConversionError(
+            "Unsupported Ossie root properties: " + ", ".join(unknown)
+        )
+    if str(root.get("version", "")) != OSSIE_VERSION:
```
jbonofre: **Exact version pin is fragile.** This rejects anything but `0.2.0.dev0` via exact equality. Since the spec is actively on a `.dev0` line, the next version bump silently breaks every real model until this constant is hand-edited. Consider accepting a compatible range (e.g. major/minor) rather than exact match.

converters/gsf/src/ossie_gsf/native_converter.py
```diff
@@ -0,0 +1,1937 @@
...
+    if isinstance(parent, exp.Anonymous):
+        return str(parent.this).lower() in _DATE_FUNCTIONS
+    try:
+        names = parent.sql_names()
+    except (AttributeError, IndexError):
+        return False
+    return any(name.lower() in _DATE_FUNCTIONS for name in names)
+
+
+def _parse_sql(sql: str) -> exp.Expression:
```
jbonofre: **SQL parsed with no dialect, and any `ParseError` aborts the whole conversion.** Converter-emitted SQL is ANSI-ish so that's fine, but *preserved* native SQL on a GSF→Ossie import can be dialect-specific (the README notes per-connection dialects). A model that imported and validated cleanly could then fail to re-export. Worth either parsing preserved SQL more leniently or documenting the assumption.

converters/gsf/src/ossie_gsf/native_converter.py
```diff
@@ -0,0 +1,1937 @@
...
+OSSIE_VERSION = "0.2.0.dev0"
+NVIDIA_GSF_VENDOR = "NVIDIA_GSF"
+GSF_VENDOR_ALIASES = {NVIDIA_GSF_VENDOR, "GSF"}
+_ID_NAMESPACE = UUID("03d14261-6432-50fe-b099-77e8061af4f9")
+_SQL_GROUPS = ("manual", "table", "sql", "bridge_table")
+_SIMPLE_COLUMN = re.compile(
+    r"^(?:(?P<qualifier>[A-Za-z_][A-Za-z0-9_]*)\.)?"
+    r"(?P<column>[A-Za-z_][A-Za-z0-9_]*)$"
+)
+_DATE_PART_UNITS = frozenset(
```
jbonofre, line 65: **Date-part filtering is a load-bearing denylist for Ossie-origin catalogs** (columns are synthesized from expression identifiers). Two failure modes: (a) a real column literally named `day`/`month`/`q`/`us` used as a genuine argument to a recognized date function is dropped; (b) a unit argument to a date function *not* in `_DATE_FUNCTIONS` gets added as a fictional physical column. The single-letter/abbrev entries (`q`, `dd`, `mm`, `us`, `ns`) widen the false-drop surface. Partly inherent to synthesizing a catalog from SQL text, mostly flagging for awareness.

converters/gsf/src/ossie_gsf/native_converter.py
```diff
@@ -0,0 +1,1937 @@
...
+        "terms": {},
+        "column_attributes": {},
+        "sql_attributes": {},
+        "custom_analyses": {},
+    }
+    if not root:
+        return result
+    try:
+        catalog = _read_catalog(root)
+    except GSFConversionError:
```
jbonofre: **Inconsistent handling of a malformed native snapshot.** Here a `GSFConversionError` from `_read_catalog` is swallowed and an empty index returned, but `_reconcile_native_relationships` (line 1784) calls `_read_catalog(native)` without catching (so the same bad snapshot silently degrades in one path and hard-fails in the other. Only reachable via a hand-edited extension, but the two paths should agree).

converters/gsf/src/ossie_gsf/native_converter.py
```diff
@@ -0,0 +1,1937 @@
...
+        or "gsf_model"
+    )
+    semantic_model: dict[str, Any] = {
+        "name": inferred_name,
+        "datasets": datasets,
+        "custom_extensions": [
+            _gsf_extension(
+                {
+                    "model_name": inferred_name,
+                    "native_document": root,
```
jbonofre, line 767: **Design note (not a blocker):** embedding the entire native document here (plus per-field/metric extensions carrying `sql`/`sql_column_is`) means the Ossie output contains a full duplicate of the GSF catalog alongside the derived model. Great for round-trip fidelity, but large and hard to hand-review. Worth calling out explicitly as a tradeoff in the README.

converters/gsf/src/ossie_gsf/native_converter.py
```diff
@@ -0,0 +1,1937 @@
...
+def _expression_matches(current: str, emitted: Any) -> bool:
+    if not isinstance(emitted, str) or not emitted.strip():
+        return False
+    try:
+        return parse_one(current).sql() == parse_one(emitted).sql()
+    except (ParseError, ValueError):
+        return current.strip() == emitted.strip()
+
+
+def _ossie_expression(expression: str) -> dict[str, Any]:
```
jbonofre: Minor: on import this always labels expressions `ANSI_SQL` even when the source GSF SQL was dialect-specific — a potential mislabel, though harmless for validation.

jbonofre (commented):
Thanks for this, it's a well-structured and impressively well-tested (nice touch running the official `validate.py` against generated output). A few robustness/maintainability notes inline. The two I'd prioritize before merge are the exact version pin and the preserved-SQL dialect-parse behavior; the rest are follow-ups or doc notes.

jbonofre (dismissed):
It looks good to me. Just minor comments from my side.

converters/gsf/src/ossie_gsf/native_converter.py
liofnv: Agreed, that was fragile. Fixed in 972e486: the check now compares major and minor only, so any 0.2.x is accepted including .dev releases, and the series is derived from OSSIE_VERSION itself rather than a second constant that could drift out of sync. Output is still written as the version the converter targets. Tests cover 0.2.0, 0.2.1 and 0.2.7.dev3 as accepted, and 0.1.9, 0.3.0 and 1.0.0 as rejected.

converters/gsf/src/ossie_gsf/native_converter.py
liofnv: Good catch, i fixed it now, both the dialect list and the opaque-SQL behavior are now documented in the README

converters/gsf/src/ossie_gsf/native_converter.py
liofnv: Reworked rather than trimmed. A keyword now counts as a unit only when it sits in the unit slot itself, so LAST_DAY(day), TRUNC(day) and DATEDIFF(month, day, x) all keep day as a real column, that's the false-drop case gone. I left q/dd/mm in deliberately, with the position check they're harmless, and removing them would turn the common T-SQL DATEDIFF(dd, ...) into a fictional column, trading your first failure mode for your second. The second one is still possible for a date function the converter doesn't recognize, which is inherent to deriving a catalog from SQL text, so I documented it rather than implying it's solved.

converters/gsf/src/ossie_gsf/native_converter.py
liofnv: Agreed, I made both fail rather than both degrade: silently ignoring a bad snapshot means quietly discarding the live GSF identifiers and emitting a document with regenerated ones that won't match the instance it came from, which felt like the worse outcome for a hand-edit to cause. There's a test asserting both entry points raise alike on the same bad snapshot.

converters/gsf/src/ossie_gsf/native_converter.py
liofnv: fixed.

converters/gsf/src/ossie_gsf/native_converter.py
liofnv: Fixed too!

liofnv:
Also rebased onto current main and picked up #113: datatype now maps to and from the physical type on the GSF column behind a field, with the two directions as inverses so a declared type survives a full cycle. A type outside Ossie's vocabulary becomes Opaque, and a type preserved from a live GSF catalog always wins over the logical one. Computed fields and metrics have no column and GSF stores no type for them, so theirs isn't carried, documented in the README. There's a test that reads the DataType enum from osi-schema.json so a future member can't be added without the mapping noticing.

jbonofre:
I added CI for this converter (on the PR directly).

jbonofre (approved):
This is a great contribution! Thanks!

PR #254, opened 2026-07-23
Add tests/CI for python models

MonkeyCanCode:
## Summary

Per [ML](https://lists.apache.org/thread/jlnk5bpy3kytmkdwvfvpjt0docfom4yx), this PR adds the missing tests as well as CI for python models.

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance


MonkeyCanCode:
As https://github.com/apache/ossie/pull/113 got merged after I created this PR, I had refactor the new test cases to follow the same pattern and refactor the existed one to use conftest.

MonkeyCanCode:
The failure CI is due to issue reported in https://github.com/apache/ossie/issues/260.

crossoverJie (approved):

MonkeyCanCode:
@jbonofre mind take a look at this? I had renamed OSI to Ossie as well for this pytest for the spec implementation in python.

python/tests/test_serialization.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_to_ossie_json_excludes_none(document_data: dict) -> None:
+    document = OssieDocument.model_validate(document_data)
+    output = document.to_ossie_json()
+    parsed = json.loads(output)
+    semantic_model = parsed["semantic_model"][0]
+    assert "description" not in semantic_model
+    assert "relationships" not in semantic_model
+    assert "vendors" not in semantic_model
```
jbonofre: This assertion checks `"vendors" not in semantic model`, but `OssieSemanticModel` doesn't have a `vendors` field at all (only `OssieDocument` does, see the `vendors` field check on `parsed` a few lines up). 
So this is tautologically true regardless of whether `exclude_none` works correctly for this model. It would never catch a regression here, unlike the `description`/`relationships` checks right next to it.

I suggest swapping it for a field that actually exists on `OssieSemanticModel` and can be `None`:

```suggestion
    assert "description" not in semantic_model # or ai_context
```

MonkeyCanCode: Fair point. Making this change.

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+  push:
+    branches: [ "main" ]
+    paths:
+      - 'python/**'
+      - '.github/workflows/python-ci.yml'
+  pull_request:
+    branches: [ "main" ]
+    paths:
+      - 'python/**'
+      - '.github/workflows/python-ci.yml'
```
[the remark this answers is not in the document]
MonkeyCanCode: Lets not add this for now.

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+on:
+  push:
+    branches: [ "main" ]
+    paths:
+      - 'python/**'
+      - '.github/workflows/python-ci.yml'
+  pull_request:
+    branches: [ "main" ]
+    paths:
+      - 'python/**'
```
jbonofre, line 31: This filter only includes `python/**` and the workflow file itself, but `test_data_type_enum_matches_core_schema` reads `core-spec/ossie-schema.json` and asserts it matches `OssieDataType`. 

A schema only change won't trigger this workflow, so schema/enum drift can land on main undetected.

`validation-ci.yml` already includes `core-spec/**` in its filters for this reason.

Maybe we should match that here?

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
```
jbonofre, line 52: Installing `uv` via `curl` with no version pin or checksum is inconsistent with the SHA-pinned `actions/*` (required by the ASF).

If that install script is ever compromised, every matrix job (including from forks) runs it unpinned. 

We should consider pinning a `uv` version/checksum, or using `astral-sh/setup-uv` pinned to a commit SHA. Thoughts?

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+    paths:
+      - 'python/**'
+      - '.github/workflows/python-ci.yml'
+  pull_request:
+    branches: [ "main" ]
+    paths:
+      - 'python/**'
+      - '.github/workflows/python-ci.yml'
+
+jobs:
```
jbonofre, line 34: I don't see `permissions:` block, so `GITHUB_TOKEN` gets the default scope.
This runs on `pull_request` (including from forks).

I recommend adding `permissions: contents: read` at the top level.

.github/workflows/python-ci.yml
```diff
@@ -0,0 +1,63 @@
...
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
jbonofre, line 38: nit: the matrix (3.11-3.14) doesn't set `fail-fast: false`, so a failure on one Python version cancels the other in flight jobs before they report.

That hides whether the code is actually broken on all four versions or just one.

python/tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+
+from collections.abc import Callable
+
+import pytest
+
+from ossie import OssieDialect, OssieExpression
+
+
+def _expression_data(value: str = "value") -> dict:
+    return {"dialects": [{"dialect": OssieDialect.ANSI_SQL, "expression": value}]}
```
jbonofre, line 26: `_expression_data()` dialect value was changed from a raw string to the enum member `OssieDialect.ANSI_SQL`. Since every test using `document_data`/`make_expression` now passes an already constructed enum, string -> enum coercion for a valid dialect value is no longer exercised anywhere (only the invalid string path remains, in `test_invalid_dialect_in_expression`).

I think it's worth keeping a plain string here, or adding a dedicated coercion test, so a regression in dialect parsing wouldn't slip through.

jbonofre (changes requested):

MonkeyCanCode:
@jbonofre thanks for review. I will review those later this weekend.

PR #257, opened 2026-07-23
[Ontology] Flatten concept declarations in ontology spec

jochenchrist:
Changes the ontology component syntax so the concept name is the value of `concept:` instead of a nested object with a `name:` key.

**Before:**

```yaml
name: EnterpriseOntology
ontology:
  - concept:
      name: SocialSecurityNr
      type: ValueType
      extends: [Integer]
  - concept:
      name: Employee
      type: EntityType
      extends: [Person]
```

**After:**

```yaml
name: EnterpriseOntology
ontology:
  - concept: SocialSecurityNr
    type: ValueType
    extends: [Integer]
  - concept: Employee
    type: EntityType
    extends: [Person]
```

- `ontology.md`: all YAML snippets use the flattened form; the component/concept schema tables are merged into one table with `concept` as the name field and `relationships` alongside it.
- `ontology.json`: `OntologyComponent` now takes `concept` as a string plus the former `Concept` fields inline; the separate `Concept` def is removed; `concept` and `type` are required.
- `examples/flights.yaml`: converted to the flattened form; content is otherwise unchanged and validates against the updated schema.
- Fixes in existing snippets: a `verbalizes` value that was a string instead of an array, missing required `verbalizes` in the `OrderLineItem` example, a missing list dash in a `derived_by` example, and a few typos.

All complete snippets in the doc and the flights example validate against the updated schema.

### DIFF
--- a/examples/flights.yaml
+++ b/examples/flights.yaml
@@ -22,110 +22,88 @@ requires:
   - COUNT[Airport] > 0         # there must be at least one Airport
   - COUNT[Carrier] > 0         # there must be at least one Carrier
 ontology:
-- concept:
-    name: NrFeet
-    description: "Unit of measure for distance in feet"
-    type: ValueType
-    extends: [ Decimal ]
-- concept:
-    name: NrPounds
-    description: "Unit of measure for weight in pounds"
-    type: ValueType
-    extends: [ Integer ]
-- concept:
-    name: NrMiles
-    description: "Unit of measure for distance in miles"
-    type: ValueType
-    extends: [ Decimal ]
-- concept:
-    name: NrMinutes
-    description: "Unit of measure for time in minutes"
-    type: ValueType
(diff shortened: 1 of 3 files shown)


kurtStirewalt:
@jochenchrist , I like this change! Please note that there is a larger ontology in `examples/flights.yaml` that will fail to validate if we merge this PR without updating it. Let me know if you'd like us to update it and share it with you for inclusion.

kurtStirewalt (approved):
LGTM

jochenchrist:
> @jochenchrist , I like this change! Please note that there is a larger ontology in `examples/flights.yaml` that will fail to validate if we merge this PR without updating it. Let me know if you'd like us to update it and share it with you for inclusion.

flights.yaml is now updated.

khush-bhatia (approved):

PR #259, opened 2026-07-23
Adding AI disclosures updates

sfc-gh-swalia:
## Summary

This PR adds section to CONTRIBUTING.md and adds optional disclosure checkbox as part of PR template.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,6 +93,22 @@ on file with the ASF. If you are contributing on behalf of your employer, a
 may also be required. Please keep individual commits signed off and attributed to
 the correct author so that provenance is clear.
 
+## AI-Assisted Contributions
+
+We welcome contributions that use AI-assisted tooling (GitHub Copilot, Claude,
+ChatGPT, etc.). Per the
+[ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html):
+
+- You may use any AI tool provided the output does not include third-party
+  copyrighted material under an incompatible license
+- You remain personally responsible for all code you submit, regardless of
+  how it was produced
+- Please disclose AI tool usage in your PR description or commit message
+  (e.g., `Co-authored-by: Copilot <copilot@github.com>` or a note in the PR)
+
+This is not a gate — it is a transparency practice that helps the community
+understand how contributions are produced.
+
 ## Contribution Workflow
 
 The project's canonical repository is hosted at
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,19 +95,8 @@ the correct author so that provenance is clear.
 
 ## AI-Assisted Contributions
 
-We welcome contributions that use AI-assisted tooling (GitHub Copilot, Claude,
-ChatGPT, etc.). Per the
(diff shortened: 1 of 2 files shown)


jbonofre (commented):
First glance: I love it 😄 

I will do the review/merge tomorrow morning my time.

CONTRIBUTING.md
```diff
@@ -93,6 +93,22 @@ on file with the ASF. If you are contributing on behalf of your employer, a
...
+
+- You may use any AI tool provided the output does not include third-party
+  copyrighted material under an incompatible license
+- You remain personally responsible for all code you submit, regardless of
+  how it was produced
+- Please disclose AI tool usage in your PR description or commit message
+  (e.g., `Co-authored-by: Copilot <copilot@github.com>` or a note in the PR)
+
+This is not a gate — it is a transparency practice that helps the community
+understand how contributions are produced.
```
khush-bhatia: Wondering if this is required by Apache ? Else I think we should skip it since directionally it looks like most PRs would be assisted by AI tools. 

cc @jbonofre 

We also want to keep our PR template bounded.

jbonofre: This is not strictly required from ASF standpoint. Especially the co-authored-by statement is not mandatory.

At the ASF, we might ask the AI tool used mostly to verify there's no legal issue (trademark, IP, etc).

I'm expecting from this PR just a link to https://www.apache.org/legal/generative-tooling.html for instance.

jbonofre: The Apache-2.0 license, and the Apache Individual Contribution License Agreement, both remind contributors that they are responsible for disclosing any copyrighted materials in submitted contributions that are not their original creation. This is as true when using generative AI tooling, as it is when using materials from public websites or code from other open-source projects.

LiJie20190102 (approved):
LGTM,+1

CONTRIBUTING.md
willpugh: I think the critical part is that each person is responsible for their contribution.

I think the "co-authored" is a little funny.  I thought AI had no copyright claims on the code it generated.  Also, it is common for me to use multiple harnesses (codex, claudecode, opencode, cursor) and multiple models with them (opus, sonnet, gpt, composer (and I expect some GLM in the future)).

Should we list all of them?

willpugh: O.k.  I looked at the generative-tooling link.  If you think the harness and/or model helps later analysis.  We can add it.  However, let's cover the "multi" case and whether we want models as well as harness.

sfc-gh-swalia: @khush-bhatia @jbonofre @willpugh - thanks for the feedback! Here's what I am thinking

- We remove the PR template checkbox entirely
- In ```contributing.md``` section - we limit to ```You remain personally responsible for all code you submit, regardless of how it was produced``` + ASF Gen tooling guidance. 

On Will's question about the multi-tool case - I think we skip prescribing format altogether. People can note tooling if they want, but we don't need to standardize how. The ASF page covers the legal angle and that should be enough.

I can push those changes quickly. Sound good?

khush-bhatia: Yes sounds good. Thanks

khush-bhatia (approved):

jbonofre (approved):

Issue #260, opened 2026-07-24
Broken CI due to failure tests in omni


MonkeyCanCode:
Getting following on main with changes introduced in https://github.com/apache/ossie/pull/113:
```
➜  omni git:(main) uv run pytest
      Built osi-omni @ file:///Users/yong/Desktop/GitHome/ossie/converters/omni
Uninstalled 1 package in 1ms
Installed 1 package in 2ms
========================================================================== test session starts ==========================================================================
platform darwin -- Python 3.11.13, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/yong/Desktop/GitHome/ossie/converters/omni
configfile: pyproject.toml
testpaths: tests
plugins: hypothesis-6.156.6
collected 86 items

tests/test_omni_to_osi.py ....................................                                                                                                    [ 41%]
tests/test_osi_to_omni.py .F..................................                                                                                                    [ 83%]
tests/test_real_world_layout.py ....                                                                                                                              [ 88%]
tests/test_roundtrip.py ....Fsss                                                                                                                                  [ 97%]
tests/test_roundtrip_properties.py ..                                                                                                                             [100%]

=============================================================================== FAILURES ================================================================================
__________________________________________________________________ test_tpcds_export_matches_expected ___________________________________________________________________

    def test_tpcds_export_matches_expected():
        with open(REPO_ROOT / "examples" / "tpcds_semantic_model.yaml") as fh:
            files = export(fh.read())
>       assert parse_files(files) == parse_files(load_fixture_dir("tpcds_omni"))
E       AssertionError: assert {'views/store...}, ...}}, ...} == {'relationshi...}, ...}}, ...}
E
E         Omitting 6 identical items, use -vv to show
E         Differing items:
E         {'views/date_dim.view.yaml': {'catalog': 'tpcds', 'schema': 'public', 'description': 'Date dimension with calendar att..., 'synonyms': ['
E
E         ...Full output truncated (2 lines hidden), use '-vv' to show

tests/test_osi_to_omni.py:63: AssertionError
_______________________ test_osi_roundtrip_up_to_documented_normalizations[/Users/yong/Desktop/GitHome/ossie/examples/tpcds_semantic_model.yaml] ________________________

path = '/Users/yong/Desktop/GitHome/ossie/examples/tpcds_semantic_model.yaml'

    @pytest.mark.parametrize("path", [
        "fixtureA_osi.yaml",
[...]

PR #263, opened 2026-07-24
Fix Omni pytest

MonkeyCanCode:
## Summary

This is sample fix for https://github.com/apache/ossie/issues/260. There are two issues introduced by https://github.com/apache/ossie/pull/113:
1. the new `datatype` field caused `test_tpcds_export_matches_expected` to fail due to Omni exporter lacks the logic to default `is_time: true` for temporal-typed fields whose explicit dimension role was unset
2. the `test_osi_roundtrip_up_to_documented_normalizations` fails due to non-native `datatype` property is dropped upon export but the test's normalization helper didn't remove it from the compared semantic model.

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance

### DIFF
--- a/converters/omni/src/osi_omni/osi_to_omni.py
+++ b/converters/omni/src/osi_omni/osi_to_omni.py
@@ -373,7 +373,15 @@ def _convert_field(field, dname, expr, fscope):
     if instructions:
         dim["ai_context"] = instructions
 
-    if (field.get("dimension") or {}).get("is_time"):
+    dimension = field.get("dimension")
+    is_time = False
+    if dimension is not None:
+        is_time_explicit = dimension.get("is_time")
+        if is_time_explicit is not None:
+            is_time = bool(is_time_explicit)
+        else:
+            is_time = field.get("datatype") in ("Date", "Time", "DateTime", "DateTimez")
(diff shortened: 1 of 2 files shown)


wForget:
Should we also trigger `converter-omni-ci` when there are changes to `core-spec/` or `examples/`?

MonkeyCanCode:
> Should we also trigger `converter-omni-ci` when there are changes to `core-spec/` or `examples/`?

For this specific one, I think it should be trigger on `python/src/ossie` as `core-spec/` doesn't directly get used by the converters. Now for examples, not everything uses `examples` as direct references, thus, I am not sure if that would be a good idea for apply for all (but for sure we can apply for this specific one). However, maybe we should raise those up in a different PR? The original problem with [113](https://github.com/apache/ossie/pull/113) got merged was no CI was ran as it didn't rebase.

MonkeyCanCode:
@jonmmease @jbonofre @khush-bhatia please take a look?

jbonofre (approved):

jonmmease (commented):
Looks good. Thanks, and apologies for the churn

PR #271, opened 2026-07-26
fix(validation): report schema errors for invalid YAML

### LINKED ISSUE #270: Validator crashes instead of reporting errors for invalid YAML structures
## Summary

`validation/validate.py` can raise `AttributeError` after JSON Schema has already identified an invalid YAML shape. The exception prevents the existing validation reporter from showing the useful schema errors it collected.

This occurs when the document root is empty or not a mapping. It can also occur when `semantic_model` has an invalid nested shape because the semantic checks assume schema-valid mappings and lists.

## Reproduction

From the repository root at `8c410f904c9bb20f683ef31292442e7bc432a8f6`:

```bash
uv run validation/validate.py /dev/null --schema core-spec/osi-schema.json
```

The command exits with an internal exception:

```text
AttributeError: 'NoneType' object has no attribute 'get'
```

Other affected inputs include a sequence root (`[]`), a scalar root, and a malformed nested value such as `semantic_model: invalid`.

## Expected behavior

The command should exit 1 through the existing validation reporter:

```text
Validation FAILED with 1 error(s):

  [Schema] (root): None is not of type 'object'
```

It should not emit an internal traceback.

## Proposed scope

- Report the schema errors already collected for structurally invalid documents.
- Run semantic checks only after schema validation has established their expected nested shapes.
- Add focused regression coverage for empty, sequence, scalar, and malformed nested inputs.
- Keep the specification, public CLI, and dependency set unchanged.

mattfaltyn:
## Summary

JSON Schema validation already detects invalid YAML shapes, but the CLI continued into semantic checks that assume schema-valid mappings and lists. Those checks could raise `AttributeError` before the existing reporter displayed the schema errors.

This change:

- reports schema errors for empty, sequence, scalar, and malformed nested YAML;
- runs semantic checks only after schema validation succeeds;
- tells users when semantic checks were deferred;
- adds focused regression coverage to the pytest suite run by Validation CI; and
- documents the schema-first validation order.

## Behavior note

When a document has both schema and semantic errors, the first run reports the schema errors and states that semantic checks were skipped. After the schema errors are fixed, a subsequent run reports any remaining semantic errors. This avoids tracebacks from malformed nested values while making the validation order explicit.

The specification, CLI arguments, and dependency set are unchanged.

## Related Issues

Closes #270

## Testing

```bash
uv run validation/test_validate.py
uv run --with pytest --with pyyaml --with jsonschema -m pytest validation/tests/
uv run validation/validate.py examples/tpcds_semantic_model.yaml
python3 -m py_compile validation/validate.py validation/test_validate.py validation/tests/test_validate.py
git diff --check
```

Results:

- The upstream duplicate-key unittest suite passes all 35 tests.
- The validation pytest suite passes all 20 tests, including the four invalid-shape cases and the schema-plus-semantic-error ordering case.
- The valid TPC-DS example passes.
- Compilation and diff checks pass.

### Validation

### Documentation

- [x] `docs/index.md` explains the schema-first validation order

### Tests

- [x] All tests relevant to this change pass locally

### Compliance

- [x] No new source files or third-party dependencies are added

### DIFF
--- a/validation/test_validate.py
+++ b/validation/test_validate.py
@@ -1,0 +1,61 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import io
+import sys
+import tempfile
+import unittest
+from contextlib import redirect_stdout
+from pathlib import Path
+from unittest.mock import patch
+
+from validation import validate
+
+
+class ValidateCliTest(unittest.TestCase):
+    def test_schema_invalid_documents_report_errors(self) -> None:
+        schema_path = Path(__file__).parents[1] / "core-spec" / "osi-schema.json"
+
+        cases = (
+            ("", "(root)"),
+            ("[]\n", "(root)"),
+            ("scalar\n", "(root)"),
+            ("semantic_model: invalid\n", "semantic_model"),
+        )
+        for content, error_path in cases:
+            with self.subTest(content=content), tempfile.TemporaryDirectory() as tmpdir:
+                yaml_path = Path(tmpdir) / "model.yaml"
+                yaml_path.write_text(content, encoding="utf-8")
+                stdout = io.StringIO()
+
+                with (
+                    patch.object(
+                        sys,
+                        "argv",
+                        ["validate.py", str(yaml_path), "--schema", str(schema_path)],
+                    ),
+                    redirect_stdout(stdout),
+                    self.assertRaises(SystemExit) as raised,
+                ):
+                    validate.main()
+
+                self.assertEqual(raised.exception.code, 1)
+                self.assertIn(f"[Schema] {error_path}:", stdout.getvalue())
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/validation/validate.py
+++ b/validation/validate.py
@@ -75,7 +75,7 @@ DIALECT_MAP = {
 SKIP_SQL_VALIDATION = {"MDX", "TABLEAU", "MAQL"}
 
 
-def validate_schema(data: dict, schema: dict) -> list[str]:
+def validate_schema(data: object, schema: dict) -> list[str]:
     """Validate against JSON Schema."""
     validator = Draft202012Validator(schema)
     errors = []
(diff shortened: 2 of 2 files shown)


kayemkim (commented):
Picked this up since #270 still reproduces on main (831f48e): `/dev/null`, `[]`, a bare scalar, and `semantic_model: invalid` all end in an `AttributeError` traceback. With the `validate.py` hunk from this PR applied on top of main, all four exit 1 with the `[Schema]` line and no traceback, the existing suites pass (`uv run validation/test_validate.py` and pytest on `validation/tests/`), and the 14 Ossie documents under `examples/` and `converters/**/tests` produce byte-identical output before and after.

Two things before this can go in:

1. It no longer merges. Main gained its own `validation/test_validate.py` in #337 (a uv script that CI runs directly), so this PR's file is an add/add conflict, and it also points at `core-spec/osi-schema.json`, which is now `ossie-schema.json`. The four cases would fit naturally as pytest functions in `validation/tests/test_validate.py`, which CI already runs.

2. `if not errors and ...` changes what a schema-invalid document reports. On main a document with one unknown property plus a duplicate dataset name and a dangling relationship target gets all three errors; with this change it gets only the schema one, and the other two show up on the next run. That is a defensible trade for never crashing, but it is a behaviour change worth stating in the PR, and maybe worth one line in the output ("semantic checks skipped until schema errors are fixed") so nobody is surprised. Gating on shape instead (root is a mapping, `semantic_model` is a list of mappings) would keep today's behaviour for the common case, but the semantic checks would still need guards for deeper malformed values, so I can see why the simpler gate was chosen.

For ordering, #373 and #383 both edit the same part of `main()` (#383 changes the `semantic_model` gate itself), so whichever lands first, the others will need a small rebase.

flyrain (approved):
+1 Thanks @mattfaltyn. cc @khush-bhatia

mattfaltyn:
Thanks for the thorough review. Both points are addressed.

> Main gained its own `validation/test_validate.py` in #337

Rebased onto current `main`, preserved that suite, and moved the four regression cases into `validation/tests/test_validate.py`, which Validation CI runs. The tests now use the renamed `ossie-schema.json` through the validator default.

> it is a behaviour change worth stating in the PR

Agreed. The CLI now states when semantic checks are skipped, the behavior is documented in the PR description and `docs/index.md`, and a regression test covers documents containing both schema and semantic errors.

Thanks again for the clear guidance.

kayemkim (commented):
Re-ran the rebased head (972e377) merged into current main (fc6c9df): the validation-ci matrix (3.11 to 3.14) passes, the four #270 inputs (empty file, `[]`, a bare scalar, `semantic_model: invalid`) exit 1 with a `[Schema]` line and the new notice instead of a traceback, and the 14 Ossie documents under `examples/` and `converters/**` keep their exit codes, the only output change being the notice on the three ontology files that already fail the core schema. The tests in `validation/tests/`, the notice, and the `docs/index.md` line cover both points I raised. LGTM.

One ordering note: #383 (approved yesterday) adds the same `not errors and isinstance(data, dict)` gate on its own, and with that branch alone the four inputs also stop crashing. If it lands first, this conflicts in both files and reduces to the notice, the docs line, and the tests, which #383 does not have for malformed shapes. The `test_schema_errors_defer_semantic_errors` fixture would then need the flat shape (`datasets:` at the root); otherwise the wrapper itself trips the schema and the duplicate-name and dangling-target parts of the case are no longer exercised.

mattfaltyn:
> The `test_schema_errors_defer_semantic_errors` fixture would then need the flat shape (`datasets:` at the root); otherwise the wrapper itself trips the schema and the duplicate-name and dangling-target parts of the case are no longer exercised.

Done, thank you. I rebased onto current `main`, converted the fixture to the flat root shape, and kept both the duplicate dataset and dangling target so the test still proves schema errors defer semantic checks. The validation suites pass locally.

PR #275, opened 2026-07-27
[OSSIE][NVIDIA_GSF] Add bidirectional GSF semantic model converter

johntomcat7408-cmyk:
## Summary

- Add bidirectional conversion between Apache Ossie and NVIDIA GSF semantic-model YAML.
- Preserve SQL dialect semantics, field labels, native metadata, and relationship integrity.
- Support quoted dataset references and dialect-correct identifier quoting.
- Add regression coverage and a dedicated GSF converter CI workflow.

## Context

This PR supersedes #247. The original implementation and discussion came from @liofnv; this replacement consolidates the change into one commit and incorporates the review fixes identified on #247.

## Validation

- Added focused regression tests for quoted references, BigQuery quoting, field-label round trips, and relationship column validation.
- Added CI coverage for Python 3.11 through 3.14.
- Local tests were not run in this session.

### DIFF
--- a/.github/workflows/converter-gsf-ci.yml
+++ b/.github/workflows/converter-gsf-ci.yml
@@ -1,0 +1,60 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Converters GSF CI
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "converters/gsf/**"
+      - ".github/**"
+  pull_request:
+    branches: [main]
+    paths:
+      - "converters/gsf/**"
+      - ".github/**"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


liofnv:
This PR contains the converter from #247, which I authored and opened on July 22. The code is substantially identical — ~95% of the non-empty lines in `converters/gsf/src/ossie_gsf/converter.py` here match #247 verbatim — but the commit is authored under a different name, so the history doesn't reflect the original authorship.

There's also a correctness problem. This copy is based on #247 as of commit 21aa2d3. That revision targeted a provisional GSF document shape predating GSF's actual interchange format. #247 has since been updated (791b440) to emit GSF's real `GsfModelDocument` contract, verified by importing the converter's output into a running GSF instance. As written, this PR would ship a GSF converter whose output GSF cannot import.

The CI workflow and regression tests added here are useful, and I'm happy to incorporate that idea into #247 with attribution.

I'd ask that this be closed in favor of #247.

johntomcat7408-cmyk:
> This PR contains the converter from #247, which I authored and opened on July 22. The code is substantially identical — ~95% of the non-empty lines in `converters/gsf/src/ossie_gsf/converter.py` here match #247 verbatim — but the commit is authored under a different name, so the history doesn't reflect the original authorship.
> 
> There's also a correctness problem. This copy is based on #247 as of commit [21aa2d3](https://github.com/apache/ossie/commit/21aa2d3226f521c93a4eb6ca993260b8ef553c69). That revision targeted a provisional GSF document shape predating GSF's actual interchange format. #247 has since been updated ([791b440](https://github.com/apache/ossie/commit/791b440f75902a12ea30c57ab0f4cea933618797)) to emit GSF's real `GsfModelDocument` contract, verified by importing the converter's output into a running GSF instance. As written, this PR would ship a GSF converter whose output GSF cannot import.
> 
> The CI workflow and regression tests added here are useful, and I'm happy to incorporate that idea into #247 with attribution.
> 
> I'd ask that this be closed in favor of #247.

Thanks for clarifying the history and for updating #247 to the current GSF interchange contract (`GsfModelDocument`). I agree that #247 is the appropriate place to continue this work, so I'll close #275.

If helpful, I'm happy to coordinate the CI and test improvements through #247 with clear attribution.

Issue #100, opened 2026-03-26
Extended Metatada Proposal for OSI


jklahr:
This proposal introduces a set of optional, non-executional metadata extensions to OSI semantic models to improve how data is interpreted, presented, and consumed by downstream tools and agents. While OSI effectively standardizes structural and logical semantics, there is currently limited support for conveying interpretability context such as units of measurement, display conventions, default aggregation behavior, KPI polarity, sorting preferences, and alignment to external semantic concepts. As a result, these details are often redefined or inferred inconsistently across developers, BI tools, and AI systems.

The proposed metadata fields (e.g., measurement, display_format, semantic_type, default_aggregation, desired_direction, default_sort, and semantic_mappings) are designed to be fully optional, backward compatible, and non-impacting to execution semantics. They provide a lightweight, extensible layer that enables more consistent rendering, more accurate automated reasoning, and reduced duplication across consuming systems, while aligning with ongoing OSI discussions around semantic typing (#42), metrics and aggregation (#43), and metadata extensibility (#55).

Detailed proposal here: https://docs.google.com/document/d/1DLuteF0WwlVbpD5kmHhcqa6vItqNFL4MoAQrImy6dKo/edit?tab=t.0

khush-bhatia:
Hey @jklahr Could you add a new public visible and comment-able doc for the above ? Also instead of semantic_mappings, how about we call it glossary references ? It seems that glossary is the more common term in the industry for this.

willpugh:
+1 on making the doc comment able.  Also, sorting behaviour DOES affect execution.  So that should be treated differently than many of the others in the doc.

PR #287, opened 2026-07-29
feat: add extended metadata fields for fields and metrics

jklahr:
## Summary

This PR proposes a new **extended metadata** layer for the Apache Ossie core spec — a standardized, optional set of non-executional, interpretability metadata fields for `fields` and `metrics`.

A companion announcement and discussion will be sent to `dev@ossie.apache.org`.

### Motivation

OSI/Ossie defines *what data is*. There is currently limited standardization for *how data should be interpreted and presented*. As a result, consumers (BI tools, AI agents, developers) must repeatedly infer or redefine units, display conventions, aggregation behavior, KPI polarity, and field visibility.

This proposal addresses those gaps through optional metadata that does not affect query execution.

**Extended proposal document** (design rationale, field-by-field detail, open questions): https://docs.google.com/document/d/1DLuteF0WwlVbpD5kmHhcqa6vItqNFL4MoAQrImy6dKo

### Changes

Three files in `core-spec/` are updated:

- **`spec.yaml`** — adds extended metadata fields to `fields` and `metrics`
- **`osi-schema.json`** — adds new `$defs`: `SemanticType`, `Measurement`, `DefaultSort`, `SemanticMapping`, `DesiredDirection`, `DefaultAggregation`, `DefaultTimeGranularity`, `UnitSystem`
- **`spec.md`** — documents all new fields with tables, type descriptions, and examples

### New fields

| Field | Applies to | Purpose |
|-------|-----------|---------|
| `display_label` | fields, metrics | Human-readable display name |
| `semantic_type` | fields, metrics | Token or URI classification (e.g., `monetary`, or `https://www.iso20022.org/glossary/LEI`) |
| `measurement` | fields, metrics | Unit/quantity metadata (`quantity_kind`, `unit` ISO 4217, `unit_system`) |
| `display_format` | fields, metrics | Excel-compatible format string (e.g., `$#,##0.00`) |
| `default_sort` | fields, metrics | Default sort direction, null handling, and optional sort-by-field |
| `glossary_references` | fields, metrics | References to concepts in external glossaries or ontologies via a SKOS-based open predicate vocabulary (`target`, `predicate`, `provenance`) |
| `hidden` | fields, metrics | Hide from consumer UIs while remaining available in expressions |
| `group_label` | fields, metrics | Organizational grouping for UI presentation |
| `desired_direction` | metrics only | KPI polarity: `higher_is_better`, `lower_is_better`, `neutral` |
| `default_aggregation` | fields only | Default aggregation when field is used as a measure |
| `default_time_granularity` | fields only | Default time bucket for temporal fields: `day`, `week`, `month`, `quarter`, `year` |

### Design principles

- **Non-executional**: no field affects query execution
- **Optional and backward compatible**: all existing models remain valid
- **Consumer-focused**: designed for BI tools, AI agents, and developers

### Community input incorporated

Dragos Crintea (maintainer of semantido, a SQLAlchemy-based semantic layer used in capital-markets/regulatory-reporting contexts) provided substantive feedback that has been incorporated into this PR:

1. `glossary_references` uses an **open predicate vocabulary** (SKOS baseline: `exactMatch`, `closeMatch`, `broadMatch`, `narrowMatch`, `relatedMatch`) rather than a closed enum, allowing non-positive assertions like `DISTINCT_FROM` via extensions. This addresses the regulatory homonym problem — e.g., "Counterparty" under EMIR and "Counterparty" under MiFIR are distinct concepts that should not be conflated.
2. `semantic_type` accepts either a **well-known token or a URI**, keeping simple cases simple while supporting governed external type systems (ISO 20022, FIBO, SKOS vocabularies, LEI, ISIN, etc.).

Dragos has also offered to contribute a reference implementation in the semantido-OSI converter and benchmark data showing accuracy improvements from this class of metadata. Those contributions will follow in subsequent PRs.

### Naming

`glossary_references` was originally proposed as `semantic_mappings`; renamed per review input from Khushboo. The name describes what the field carries more directly — pointers to concepts in an external glossary, ontology, or standard — and avoids confusion with `semantic_type`. The structure is unchanged.

### Impact on existing implementations

None — all additions are optional. The JSON schema uses `additionalProperties: false` per existing convention; all new fields are added explicitly to the `Field` and `Metric` definitions.

### Related discussions

- Discussion #42 (Semantic typing): https://github.com/apache/ossie/discussions/42
- Discussion #43 (Aggregation semantics): https://github.com/apache/ossie/discussions/43
- Discussion #55 (Metadata extensibility): https://github.com/apache/ossie/discussions/55
- Issue #100: https://github.com/apache/ossie/issues/100

### AI disclosure

This contribution was developed with AI assistance (Snowflake Cortex Code). The author (Josh Klahr) has reviewed all changes and is personally responsible for the content per the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).

.... Generated with [Cortex Code](https://docs.snowflake.com/en/user-guide/cortex-code/cortex-code)

Co-Authored-By: Cortex Code <noreply@snowflake.com>

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -351,7 +351,7 @@ Ossie standardizes structural and logical semantics well, but there is limited s
 
 **Roadmap Deliverables:**
 
-- [Extended Metadata Proposal for Ossie](https://github.com/apache/ossie/issues/100) — optional, backward-compatible metadata fields (e.g., `measurement`, `display_format`, `semantic_type`, `default_aggregation`, `desired_direction`, `default_sort`, `semantic_mappings`)
+- [Extended Metadata Proposal for Ossie](https://github.com/apache/ossie/issues/100) — optional, backward-compatible metadata fields (e.g., `measurement`, `display_format`, `semantic_type`, `default_aggregation`, `desired_direction`, `default_sort`, `glossary_references`)
 - Richer application-specific extension points beyond `custom_extensions`
 - Sample value annotations for documentation and AI grounding
 
--- a/core-spec/osi-schema.json
+++ b/core-spec/osi-schema.json
@@ -124,6 +124,86 @@
       ],
       "description": "Logical data type for fields and metrics, independent of role (e.g. dimension vs fact) and physical representation. `Decimal` is exact base-10 with unspecified precision and scale; `Float` is approximate. `DateTime` has no timezone or offset, while `DateTimeTz` identifies an instant using offset or timezone context but does not guarantee preservation of a named timezone. Omit `datatype` when unknown; use `Opaque` plus `custom_extensions` for a known type outside the portable vocabulary."
     },
+    "SemanticType": {
+      "type": "string",
+      "enum": ["categorical", "quantitative", "monetary", "temporal", "geographic", "ordinal", "identifier"],
+      "description": "High-level semantic classification of a field or metric"
+    },
+    "DesiredDirection": {
+      "type": "string",
+      "enum": ["higher_is_better", "lower_is_better", "neutral"],
+      "description": "KPI polarity indicating whether higher or lower values are preferred"
+    },
+    "DefaultAggregation": {
+      "type": "string",
+      "enum": ["sum", "avg", "min", "max", "count", "count_distinct"],
+      "description": "Default aggregation function to apply when this field is used as a measure"
+    },
+    "DefaultTimeGranularity": {
+      "type": "string",
+      "enum": ["day", "week", "month", "quarter", "year"],
+      "description": "Default time bucket for temporal fields"
+    },
+    "UnitSystem": {
+      "type": "string",
+      "enum": ["si", "imperial", "custom"],
+      "description": "Unit system classification"
+    },
+    "Measurement": {
+      "type": "object",
+      "description": "Describes what a numeric value represents (units and measurement)",
+      "properties": {
+        "quantity_kind": {
+          "type": "string",
+          "description": "The kind of quantity (e.g., currency, length, weight, temperature, percentage, duration)"
+        },
+        "unit": {
+          "type": "string",
+          "description": "Specific unit. For currency use ISO 4217 (e.g., usd, eur). For others use UCUM or descriptive strings."
+        },
+        "unit_system": {
+          "$ref": "#/$defs/UnitSystem"
+        }
+      },
+      "additionalProperties": false
+    },
+    "DefaultSort": {
+      "type": "object",
+      "description": "Default sorting behavior",
+      "properties": {
+        "direction": {
+          "type": "string",
+          "enum": ["asc", "desc"],
+          "description": "Sort direction"
+        },
+        "nulls": {
+          "type": "string",
+          "enum": ["first", "last"],
+          "description": "Null value positioning"
+        },
+        "by_field": {
+          "type": "string",
+          "description": "Sort by a different field (e.g., sort month names by month number)"
+        }
+      },
+      "additionalProperties": false
+    },
+    "SemanticMapping": {
+      "type": "object",
+      "description": "Link to an external ontology or standard",
+      "properties": {
+        "source": {
+          "type": "string",
+          "description": "Ontology or standard name (e.g., schema.org, FIBO)"
+        },
+        "identifier": {
+          "type": "string",
+          "description": "URI or identifier within that ontology"
+        }
+      },
+      "required": ["source", "identifier"],
+      "additionalProperties": false
+    },
     "Dimension": {
       "type": "object",
       "description": "Dimension metadata",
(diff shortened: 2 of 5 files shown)


djwaldo:
1. Have we considered locale settings in regards to display_format? I.e. for EN_AU i might want my currency to be displayed as $#,##0.00 however for SE_SE i might want the display to be $# ##0,00

2. Have we considered multiple group_labels? A column may be present in multiple group sections. I.e. i could define groups for HR Team, Sales Team then also KPIs, Attributes. So 'Sales Amount' could be in three groups.

jklahr:
[djwaldo](https://github.com/djwaldo) good question
- for the first one I wonder if we want to introduce the concept of locale files (basically extra files per local that include the ability to create locale specific metadata). I could imagine wanting to do this for descriptions, synonyms, etc. 

- for groups, yes I think we might want to do this, I can update

djwaldo:
Thanks for picking up the grouping point,`group_labels` as an array with "A field may belong to more than one group" is exactly it.

On the locale question, let me offer what ThoughtSpot actually does, since we shipped this and the shape may be useful input. Short version: **we came to the same "locale files" conclusion you floated, but the key turned out not to be locale.**

**How it works.** Localized names live in a *separate document*, a `column_alias` object, not inline on the model, and each record is keyed by a four-tuple:

```
(column, locale, org, group)  ->  alias, description
```

So the same column can read "Revenue" for one audience, "Chiffre d'affaires" for a French locale, and "Net Bookings (AUD)" for a particular tenant, all at once, resolved at read time against the viewer's locale *and* their org and group membership.

Three things fell out of that which might be worth folding into the design:

1. **Locale alone was not enough a key.** We started there and had to widen it. In practice the same question, "what should this column be called for *this* reader", is answered by tenant and audience at least as often as by language, and once you have two mechanisms for renaming a column you have a precedence problem. One keyed record handles all three.

2. **It carries `description`, not just the label.** A localized display name over an English description is a half-translated model, and users notice. If `display_label` becomes localizable, `description` almost certainly has to travel with it.

3. **It is a separate document because it does not fit inline.** This is the practical part. Our TML import has size limits where a wide model across several locales and a few dozen tenants gets there faster than you would guess.  Anything that puts the full locale matrix inline in the core document will hit it. We do have delta strategies to mitigate this. 

**On `display_format` specifically** — and this is a genuine gap on our side, not advice. Our alias record carries `alias` and `description` and *not* format, so ThoughtSpot has no per-column-per-locale format string; number and currency presentation is resolved from instance and user settings instead. That works until a model is genuinely multi-region, and then `$#,##0.00` versus `$# ##0,00` is not a preference, it is wrong output.

So I would gently push back on `display_format` being a single Excel-compatible string on the field. Either it is understood to be locale-neutral, in which case the spec should say what a consumer does with it in a non-`en_US` locale, or it belongs in whatever locale mechanism eventually lands, alongside `display_label` and `description`. What I would most want to avoid is it being the one presentation attribute that *cannot* be localized, since it is arguably the one where getting it wrong is most visible.

Happy to write up the alias schema in more detail if a locale proposal gets going — including the parts we would do differently.

djwaldo:
Two concrete mismatches from mapping ThoughtSpot onto this PR's fields. Both are cheap to change now and awkward once the enum ships, so raising them while it is open.

**1. `default_time_granularity` has no equivalent of "let the engine decide."**

ThoughtSpot's per-column default grain is `DAILY | WEEKLY | MONTHLY | QUARTERLY | YEARLY | AUTO`. The five named values map onto `day | week | month | quarter | year` one-for-one. `AUTO` does not, and it is not an edge case, it means "pick the grain from the time span the query actually covers", so a five-year query buckets by month and a one-week query by day, from the same column definition.

Under the enum as drafted, a model using it either loses the setting or has to pick a fixed grain that is wrong at most zoom levels. Adding `auto` costs one enum value; a consumer that does not implement it can fall back to whatever it does today. Worth noting the value is genuinely different in kind from the other five, it is a deferral rather than a grain, so if that distinction is unwelcome in the enum, a separate boolean would also solve it.

**2. Is "Excel-compatible" the canonical vocabulary for `display_format`, or the example?**

The field is specified as an "Excel-compatible display format string (e.g. `$#,##0.00`)". ThoughtSpot has a native format-pattern vocabulary of its own, and I would guess most tools with a formatting layer do. That makes this a *translation* on the way in and out rather than a passthrough, which is fine, but it needs saying, because translation raises questions the spec is currently silent on:

- What should a converter do with a native pattern that has no Excel equivalent? Drop it, approximate it, or refuse?
- Is a consumer expected to *render* Excel format strings, or to translate them back into its own vocabulary?
- Round-tripping `native -> Excel -> native` will not be lossless for every pattern. Is that understood and accepted, or should the original be preserved alongside?

One option worth considering: allow the field to carry the format plus an optional vocabulary tag, defaulting to Excel. That keeps the simple case simple and gives a lossless path for tools that have their own. But mainly I want to flag that "Excel-compatible" is doing a lot of load-bearing work in one adjective, and the round-trip semantics are the part that will bite implementers.

Neither of these blocks the PR from my side, the extended-metadata layer is a clear improvement and I would rather see it land than see it perfect.

Issue #248, opened 2026-07-22
Add a Cube converter


0sparsh2:
Hi all 👋

I'd like to gauge interest in adding a converter between Ossie and **Cube** (cube.dev) under `converters/`.

[Cube](https://github.com/cube-js/cube) is a widely-used open-source semantic layer. Its data model is defined in YAML as `cubes` with `sql_table`/`sql` (source), `dimensions`, `measures`, and `joins` — conceptually very close to Ossie's `dataset` / `field` / `metric` / `relationship`:

| Cube concept | Ossie concept |
|---|---|
| `cube.name` | `dataset.name` |
| `cube.sql_table` / `cube.sql` | `dataset.source` |
| `cube.dimensions[]` | `dataset.fields[]` |
| `dimension.sql` | `field` expression |
| `dimension.type: time` | `field.dimension.is_time` |
| `cube.measures[]` | `metric` (dataset-scoped) |
| `measure.sql` + `measure.type` (sum/count/avg/...) | `metric.expression` + aggregation |
| `cube.joins[]` (`relationship: many_to_one` etc., `sql` join condition) | `relationship` (`from`/`to`/`from_columns`/`to_columns`) |
| `dimension.meta` / `measure.meta` | candidate home for round-trip metadata via `custom_extensions` |

I'd propose following the existing converter pattern (e.g. `converters/omni`, `converters/honeydew`):

1. A stable `vendor_name: CUBE` for `custom_extensions`.
2. Export (Ossie → Cube): one Cube YAML file (or one-file-per-cube layout) per Ossie dataset, with metrics attached as measures on their owning dataset and relationships translated to Cube `joins`.
3. Import (Cube → Ossie): parse a Cube schema directory into a single Ossie YAML, preserving anything Cube-specific (e.g. `pre_aggregations`, `data_source`, `public`/`shown` flags, `format`) in `CUBE` custom extensions so nothing is silently dropped.
4. Tests built off the TPC-DS example (`examples/tpcds_semantic_model.yaml`), including a round-trip test, following the `converters/omni` test layout.
5. A README documenting the mapping table and any unsupported constructs (Cube's `segments`, `pre_aggregations`, and Jinja-templated SQL are likely out of scope for v1 and would be called out as limitations, similar to how `converters/honeydew` documents its limitations).

[...]

PR #289, opened 2026-07-29
Add bidirectional Apache Ossie <-> Cube converter

MikeNitsenko:
## Summary

Adds a bidirectional converter between Apache Ossie semantic models and [Cube](https://cube.dev/docs/product/data-modeling/overview) data models, under `converters/cube/`. Pure offline YAML transform — no Cube deployment, API token, or network access required, matching the other Python converters in this repo.

- **Import** (`ossie-cube import`): Cube files → Ossie. Cube-only constructs (segments, pre-aggregations, hierarchies, folders, view curation, formats, access policies, …) are preserved in `custom_extensions[CUBE]`, so **Cube → Ossie → Cube is lossless**.
- **Export** (`ossie-cube export`): Ossie → Cube files. Cube has a `meta` field at every level, so Ossie constructs Cube has no slot for (`unique_keys`, foreign-vendor `custom_extensions`, the structured form of `ai_context`) are parked under `meta.ossie` rather than dropped — making **Ossie → Cube → Ossie lossless too**.

The Ossie `semantic_model` maps to a Cube **view**, not a cube. Cube users are view-first, and Cube's own AI agent reads `meta.ai_context` only from views and individual members — cube-level AI context is explicitly not consumed — so the view is the natural model boundary.

## Fan-out semantics

Cube corrects for join row-multiplication **at query time**: when a cube sits on the multiplied side of a join it builds `SELECT DISTINCT <primary key> FROM <join>`, joins that key set back to the measure's own cube, and aggregates there, so each source row is counted once. A static Ossie expression has no way to inherit that. Cube also refuses outright when the measures themselves span cubes that fan out.

So the converter emits the fan-out-safe form wherever one exists, and reports the cases where none exists:

| Cube measure | Ossie expression | Safe under fan-out? |
|---|---|---|
| bare `count` | `COUNT(DISTINCT <pk>)` | **Yes, exactly** — Cube renders `count(pk)` normally and `count(distinct pk)` when multiplied; `COUNT(DISTINCT pk)` equals both |
| `count_distinct` / `count_distinct_approx` | `COUNT(DISTINCT x)` / `APPROX_COUNT_DISTINCT(x)` | Yes, inherently |
| `min` / `max` | `MIN(x)` / `MAX(x)` | Yes — idempotent under duplication |
| `sum`, `avg`, `count` + `sql` | `SUM(x)`, `AVG(x)`, `COUNT(x)` | **No** |

Only the last row is at risk, and only when its cube is the `to` (one) side of a relationship in the model. That is computable from the Ossie graph, and the converter records a structured `FANOUT_UNSAFE_METRIC` issue naming the metric, the dataset, and the responsible relationship. Refusing the model outright would leave the spoke on the other side with nothing, and these are the metrics most worth converting; `--strict-fanout` restores the refusal, mirroring Cube's own.

Going the other way, an Ossie metric combining several aggregates is **decomposed** into one `public: false` measure per aggregate, each declared on the cube its own operand reads, plus a `type: number` measure referencing them. Cube's correction then applies per aggregate instead of once for the whole expression. The parts carry `meta.ossie.part_of`, so import skips them and inlines their SQL back through the references, recovering the original expression exactly.

**This points at a spec gap.** Ossie has no additivity or grain declaration to record non-additivity properly. dbt's `non_additive_dimension` is the nearest precedent, and this repo's dbt converter already loses the same information (`osi_to_msi.py` hard-codes `non_additive_dimension=None`, with a named `CUMULATIVE_SEMANTICS_LOSS` issue type). Raised separately as #290.

## Other design notes

- **Member references follow Cube's semantics rather than being uniform**: `{CUBE.member}` when the dataset declares a field of that name (reuses the member's SQL, compile-time checked), `{CUBE}.column` for a raw physical column, and `{other_cube.member}` across cubes — which is also what gives a cross-dataset Ossie metric its implicit join. The cube's own name is never emitted, so models survive `extends`.
- **Calculated measures inline their `{other_measure}` references**, because that is what Cube itself does; Ossie has no metric-to-metric reference. Cycles are rejected. Locating the aggregates inside a composite expression uses `sqlglot`, already a runtime dependency of the dbt and NVIDIA GSF converters for the same purpose.
- **Measure `filters` fold into `CASE WHEN … END`** inside the aggregate, matching Cube's own `applyMeasureFilters` rendering and the filtered-aggregation idiom the Ossie expression language endorses.
- **Dimension `type: number` omits Ossie `datatype`** rather than asserting a precision the model does not carry — Cube collapses Integer/Decimal/Float into one type, and the spec says to omit when unknown. The original type is stashed.
- **`type: geo` dimensions split** into `<name>_latitude` / `<name>_longitude`, since an Ossie field holds one expression and a geo dimension has two.

## Unsupported constructs

- **`extends`** — resolving it means reproducing Cube's definition-merge semantics exactly, so it is refused rather than half-applied.
- **Jinja-templated YAML and `.js`/`.ts` models** — preserved verbatim and never half-converted. Jinja is detected per file, the same rule Cube's own `CubeSchemaConverter` uses for the Rollup Designer.

Losses the converter *can* absorb are returned as structured `ConverterIssue`s (following the `osi-dbt` converter) rather than printed to stderr and forgotten, so a pipeline can gate on them.

## Testing

226 tests, 96% line coverage:

- example-based unit tests per direction, plus CLI behavior tests;
- fixture round-trip tests in both directions, including a TPC-DS model derived from `examples/tpcds_semantic_model.yaml` as the converter guide asks;
- every emitted Ossie document validated against `core-spec/osi-schema.json`;
- Hypothesis property-based round-trip tests over generated Cube models, with a seeded fallback so the properties still run where `hypothesis` is unavailable.

Also verified by hand against a real production Cube model: round trip content-identical with original filenames preserved, output passing `validation/validate.py`, and the fan-out guard correctly flagging the two measures on the joined cube's `one` side.

## Related Issues

Related to #290 (spec has no way to declare a non-additive metric).

### Specification
- [ ] N/A — no `core-spec/` changes

### Ontology
- [ ] N/A — no `ontology/` changes

### Converters
- [x] `CUBE` registered in the supported-vendors table in `converters/README.md`

### Validation
- [ ] N/A — no `validation/` changes; emitted models are validated *by* the existing `validation/validate.py`

### Documentation
- [x] `converters/cube/README.md` documents the full mapping, the fan-out behavior, requirements, and limitations

### Examples
- [ ] N/A — no new spec constructs; the existing TPC-DS example is used as the test baseline

### Tests
- [x] All existing tests pass (226 tests, 96% coverage; CI workflow added)

### Compliance
- [x] No new runtime dependencies (PyYAML only, as with the other Python converters). The `jsonschema` dev dependency already ships in `converters/orionbelt/`

## AI assistance

This contribution was developed with AI assistance (Claude). I have reviewed the code and tests and take responsibility for them, per the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).


MikeNitsenko:
Both addressed in d362c9e.

**`members_by_cube` collisions.** Correct, and worse than described. The two sanitization passes could disagree: `{"Order Status", "order status"}` collapsed to a single `order_status` with no error in `_convert_model`, and was only rejected later in `_build_dimensions` — after measures had already been placed against the short member set.

Names are now resolved once in `_resolve_dimension_names` and reused by every stage, so sanitization and collision detection happen in exactly one place. Added a test that a collision is rejected with a metric present, which is the ordering that was fragile.

This also fixed something not mentioned: the old set included both halves of a split `geo` dimension (`location_latitude`, `location_longitude`). Those never exist as Cube dimensions — they merge back into `location` — so a metric referencing one emitted an unresolvable `{CUBE.location_latitude}`. The halves now resolve to the dimension they merge into.

**`_HypothesisRnd.chance()` ignoring `p`.** Also correct. Now weighted, and drawn so the minimal value means `False`, which shrinks toward the smallest model rather than the largest. Verified the realised probability matches `p` exactly for every value the builders use.

MikeNitsenko:
Both worth addressing, and the first one was a real bug. Fixed in 31d49d2.

**Dropped foreign-vendor extensions.** Reproduced it: import a Cube model with two views (so no view is mapped), add a `SNOWFLAKE` extension to the Ossie model, export — and the extension is gone, reported under an issue type that says "parked".

Now refused, with the fix in the message (`--view <name>` on the import so a view is mapped). I considered parking on a deterministic view as suggested, but rejected it: import only reads parked extensions back from the *mapped* view, so that would have looked lossless while still losing them. Refusing is the honest option and matches the converter's stated contract of never dropping a field silently.

**The type/message inconsistency was in two places, not one.** `PARKED_IN_META` is documented as preserved-but-invisible-to-Cube, so reporting a genuine drop under it would let a caller gating on issue types conclude the data survived. Added `DROPPED_NO_CUBE_EQUIVALENT` and used it for relationship `ai_context` — a Cube join entry takes only `name`/`sql`/`relationship`, with no `meta`, which makes it the one construct here with nowhere to go.

**Hard-coded test count.** Removed from the README. It had already drifted out of sync with the PR description, which is exactly the point — the number carries no information a reader needs, while the description of what the suite covers does.

MikeNitsenko:
Confirmed and fixed in d92711e. Reproduced it first — with `longitude` first and a `city` dimension between the halves, `city` disappeared from the output entirely:

```
emitted dimensions:
   {'name': 'home', 'type': 'geo', 'latitude': {...}, 'longitude': {...}}
'city' present? False
```

I took the second of your two suggestions rather than the first. Reserving the placeholder earlier would have worked, but the index arithmetic was the fragile part, so it's gone: dimensions are built into a dict keyed by target name, with order taken from each name's first appearance. That's well defined however the halves are arranged — either order, adjacent or not.

Probing around it turned up two more order-dependent paths in the same code:

- **A geo base colliding with an ordinary field of the same name.** With the ordinary field first, export emitted two dimensions called `home` (invalid Cube); with it second, `sanitize_name` correctly rejected it. Order was deciding whether an invalid model was produced. Now checked during name resolution.
- **Two fields both claiming the same half** silently discarded one. Now rejected.

Also validated the geo `part` and `of` values, and moved the missing-half check into name resolution so every geo problem is caught in one place before anything is built.

Six cases pinned by tests: either order, collision in both orders, duplicate half, missing half, unknown part.

MikeNitsenko:
Correct, and it was the third instance of this — so rather than patch the flagged line I audited all seven export-side uses. Exactly one was a genuine park. Fixed in c22e6d2.

| site | was | now |
|---|---|---|
| `unique_keys` | parked | parked (correct) |
| extra `semantic_model` entries | parked | **dropped** |
| `dimension.is_time` role | parked | **dropped** |
| `dimension.is_time` opt-out | parked | **dropped** |
| synthesized primary-key dimension | parked | **approximated** |
| no `datatype` → Cube `type: string` | parked | **approximated** |
| cross-dataset metric placement | parked | **approximated** |

The import-direction uses were all genuine parks and are unchanged.

Added `APPROXIMATED` for the middle group, which neither existing type described: nothing is lost and nothing is hidden, but Cube requires a value Ossie does not carry, so the converter chose one and the output asserts slightly more than the input did. Labelling that "parked" was wrong in the same way as labelling a drop "parked" — nothing was parked in either case.

Three types kept apart on purpose, since that is what a caller gating on them needs: preserved-but-unreadable-by-Cube, actually lost, and emitted-with-a-guess. Two of the three were indistinguishable before.

MikeNitsenko:
PR #289 — reviewer note

Tests: 534, or 561 with the optional Cube gate. Coverage: 97%.

Layers:

* Three whole models: a small one, TPC-DS (as requested by the converter guide), and one produced by the Databricks converter, where nothing was written “for Cube.” They are compared as whole documents, so any unintended change appears as a readable diff.
* Feature matrix: one fixture per Cube feature, ten in total: case/switch dimensions, granularities, presentation metadata, measure variants, hierarchies, segments, pre-aggregations, access policies, view curation, sub_query, and computed primary key. Each fixture is a valid Cube model and is asked the same four questions: does it convert; is the Ossie output valid according to the spec; does Cube → Ossie → Cube reproduce it; and does Cube still compile the result.
* Property tests from both ends: 60 seeds plus 150 generated cases per direction. Generating only from Cube says nothing about hand-authored Ossie, and that is the harder direction.
* Cube itself as a gate: tools/cube_compile.js asks Cube whether what we emit actually loads. It requires a locally built Cube checkout (OSSIE_CUBE_REPO); without one, those tests are skipped, so CI is unaffected.
* Idempotence: a second export cycle must produce the same Cube model as the first. This is the only way to catch a record that one side writes and the other reads differently.

Cross-matrix (tools/interop_matrix.py): our Ossie output is run through all nine Python converters, with a report showing what each one made of it. This is what the stash reduction was measured against: 7 CUBE entries on TPC-DS instead of 41, and 2 Databricks warnings instead of 32.

Databricks — both directions:

* Cube → Ossie → metric view.
* metric view → Ossie → Cube, using all three fixtures from that converter; the result compiles in Cube.
* Full loop: metric view → Ossie → Cube → Ossie → metric view. Source, joins, dimensions, and measures come back with the same names.

Snowflake: that converter is export-only, so there is no reverse path in the repository. One direction was verified: Cube → Ossie → Cortex Analyst.

These interop paths are what the tests specifically pin down: which dialect an expression carries, where the primary key comes from, and whether a field has a dimension role. Each is a point where the two formats disagree and the conversion must make an explicit choice.

Limitations:

* Ossie cannot declare a measure as non-additive, so the fan-out risk can only be reported, not recorded in the model — issue #290.
* A metric view does not carry a primary key, while Cube requires one on any cube with a join. It must therefore be added manually using primary_key or unique_keys. We report this using Cube’s own error message.
* The compile gate requires a local Cube checkout, so it is skipped in CI.
* Polaris and Salesforce are written in Java and require Maven. They are listed in the matrix as unsupported rather than being silently skipped.
* dbt fails in the matrix because of its own CLI bug — issue #296, not ours.

What I need: CI has never run on this PR because GitHub holds workflows from external contributors until a committer approves them. The PR needs a review and a CI run.

mfournioux:
Hi @MikeNitsenko @jbonofre , this PR can be merged or additional checks are expected ?

MikeNitsenko:
@mfournioux, all checks and updates from my side were implemented - looking for review and approval.

@jbonofre @khush-bhatia  this adds a bidirectional Cube <-> Ossie converter, following the same shape as #247 and #239. It’s the implementation for issue #248, which asked whether a Cube converter would be welcome and hasn’t had an answer yet.

Would one of you be able to review it and approve the CI workflow run?

paveltiunov (approved):

MikeNitsenko:
@jbonofre @QMalcolm @khush-bhatia could you please review and approve?

MikeNitsenko:
@jbonofre could you please take a look

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1850 @@
...
+    issues.add(IssueType.GEO_DIMENSION_SPLIT, f"{cname}.{dname}",
+               f"split into '{dname}_latitude' and '{dname}_longitude'; an Ossie "
+               f"field holds a single expression")
+    host_extras = {
+        snake(k): v for k, v in dim.items()
+        if snake(k) not in ("name", "type", "latitude", "longitude")
+    }
+    out = []
+    for part in ("latitude", "longitude"):
+        sub = (dim.get(part) or {}).get("sql")
```
jbonofre: Bare-string `latitude`/`longitude` in a geo dimension crashes with `AttributeError` instead of a clean conversion error.

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1850 @@
...
+def _decompose_join_sql(sql, own_cube, target, what, cubes, issues):
+    """Split a Cube join `sql` into (own_column, target_column) pairs.
+
+    Only an AND-chain of equalities between one own-cube reference and one
+    target-cube reference has an Ossie relationship form. Anything else -- a
+    range/non-equi condition, a comparison against a literal, a third cube --
+    returns None, and the caller preserves the join in the stash instead.
+    """
+    pairs = []
+    for clause in _AND_SPLIT_RE.split(sql):
```
jbonofre: `_AND_SPLIT_RE` splits join SQL inside quoted string literals (e.g. `OPEN AND PENDING`), marking valid joins as unconvertible.

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1850 @@
...
+    if unqualified:
+        # An unsafe aggregate over an unqualified column reads the declaring cube.
+        found.add(own_cube)
+    return found
+
+
+def _windowing_key(measure):
+    """The first windowing key present on a measure, or None."""
+    for key in _WINDOWING_KEYS:
+        if measure.get(key):
```
jbonofre: `rolling_window: {}` (Cube's default trailing-30-day window) is falsy, so windowed measures are silently converted as plain aggregates.

converters/cube/src/ossie_cube/osi_to_cube.py
```diff
@@ -0,0 +1,1707 @@
...
+        # as `id_pk`. Both name something the table need not have, so the column list is
+        # recorded -- and only then, so a model whose names already agree keeps a clean
+        # round trip.
+        parked["primary_key"] = key_columns
+    if plan.key_from_unique_keys and pk_names:
+        issues.add(IssueType.APPROXIMATED, scope,
+                   f"no primary_key, so the first unique_keys entry "
+                   f"({', '.join(plan.primary_key)}) is marked as the Cube primary key; "
+                   f"Cube requires one on any cube that declares a join")
+    if joins and not pk_names:
```
jbonofre: "joins but no primary key" warning fires before stashed joins are merged in. Models with only stashed joins emit no warning but fail Cube compilation.

converters/cube/src/ossie_cube/osi_to_cube.py
```diff
@@ -0,0 +1,1707 @@
...
+            operand = str(expr).strip()
+        elif mtype == "count" and operand is None:
+            # Declared `count_distinct` over the primary key: the operand is the key
+            # itself, which the bare-count collapse set aside.
+            operand = pk_operand
+        mtype = declared
+    if mtype == "count" and operand is None:
+        measure["type"] = "count"
+    else:
+        measure["sql"] = stash.get("sql") or refs.to_cube_sql(
```
jbonofre: `stash.get('sql') or ...` discards `sql: 0` (integer zero). I think the fix is `is not None` check.

converters/cube/src/ossie_cube/_common.py
```diff
@@ -0,0 +1,1360 @@
...
+
+# An aggregate call that maps back onto a structured Cube measure.
+AGG_CALL_RE = re.compile(
+    r"^\s*(SUM|AVG|MIN|MAX|COUNT|APPROX_COUNT_DISTINCT)\s*\((.*)\)\s*$",
+    re.IGNORECASE | re.DOTALL,
+)
+DISTINCT_RE = re.compile(r"^DISTINCT\s+(.+)$", re.IGNORECASE | re.DOTALL)
+
+
+def balanced_parens(s):
```
jbonofre, line 1314: `balenced_parens()` doesn't skip quoted literals. `)` inside a string misclassifies a valid aggregate.

converters/cube/src/ossie_cube/expressions.py
```diff
@@ -0,0 +1,407 @@
...
+
+def _match_paren(text, open_at):
+    """Index of the `)` closing the `(` at `open_at`, honouring quotes."""
+    depth, quote = 0, None
+    for i in range(open_at, len(text)):
+        ch = text[i]
+        if quote:
+            if ch == quote:
+                quote = None
+        elif ch in "'\"":
```
jbonofre: `_match_paren()` ignores backtick quoting (BigQuery column names with parens break paren depth for instance).

converters/cube/src/ossie_cube/_common.py
```diff
@@ -0,0 +1,1360 @@
...
+def _split_top_level_and(where):
+    """Split `(f1) AND (f2)` into [f1, f2], or None when not that exact shape."""
+    pieces, start, depth, quote = [], 0, 0, None
+    i = 0
+    while i < len(where):
+        ch = where[i]
+        if quote:
+            if ch == quote:
+                quote = None
+        elif ch in "'\"":
```
jbonofre: `_split_top_level_and` ignores backtick quoting. `AND` inside a backtick identifier is treated as a top-level separator.

converters/cube/src/ossie_cube/_common.py
```diff
@@ -0,0 +1,1360 @@
...
+    return None
+
+
+def source_part_count(source):
+    """How many identifier parts a dotted dataset `source` has, or None for a query.
+
+    Dots inside double quotes or backticks belong to a quoted identifier, not to the
+    path -- `"My.Catalog".public.t` is three parts, not four.
+    """
+    s = str(source).strip()
```
jbonofre, line 749: `source_part_count` ignores single-quote quoting. Dots inside single-quoted identifiers inflate part count, suppressing `SOURCE_NOT_FULLY_QUALIFIED`.

jbonofre (changes requested):

converters/cube/src/ossie_cube/_common.py
MikeNitsenko: `balanced_parens()` now ignores quoted regions using `quoted_char_mask()` in [_common.py (line 1314)](/Users/mnitsenko/.codex/worktrees/12b0/ossie/converters/cube/src/ossie_cube/_common.py:1314), with a `SUM(')')` regression test

converters/cube/src/ossie_cube/_common.py
MikeNitsenko: `source_part_count()` now uses the same quote mask, including single quotes, in [_common.py (line 743)](/Users/mnitsenko/.codex/worktrees/12b0/ossie/converters/cube/src/ossie_cube/_common.py:743), with a single-quoted dotted-name test

MikeNitsenko:
@jbonofre All review comments have been addressed, and the full Cube compilation gate passes. Could you please re-review?

jbonofre:
@MikeNitsenko awesome! Thanks! I will do a new pass!

MikeNitsenko:
@jbonofre, looking forward to the next review pass

MikeNitsenko:
@jbonofre, it would be really helpful to get the second review pass, as this is currently blocking us from proceeding with further updates and improvements to Ossie. Thanks!

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1863 @@
...
+        # A throwaway issue log: the real report was made when the view was
+        # generated; predicting it again is not a second event.
+        "cubes": generated_view_cubes(identity, relationships, base, members,
+                                      name, IssueLog()),
+    }
+    body = {k: v for k, v in view.items() if k not in ("description", "meta")}
+    return body == predicted
+
+
+def _fanned_out_datasets(relationships):
```
jbonofre: Fan-out detection only looks at joins whose predicate reduced to a plain physical column equality (made it into `relationship`).
A join whose predicate touches a computed/aliased member is parked in `extra_joins`and is invisible to `_fanned_out_datasets`.

Example: a `many_to_one` join with `sql: "{CUBE}.customer_key = {customer.full_key}"` where `full_key` is computed: `_decompose_join_sql` returns `None`, so the join never reaches `relastionships`, and fan-out-unsafe `SUM` measure on `customers` converts with zero `FANOUT_UNSAFE_METRIC` warning. It looks like a corruption the fan-out check exists to catch.

converters/cube/src/ossie_cube/expressions.py
```diff
@@ -0,0 +1,405 @@
...
+    """
+    tree = parse(expr)
+    return tree is not None and isinstance(tree, _AGGREGATE_NODES)
+
+
+# The aggregate call names this converter maps to a Cube measure type. Scanned for
+# in the source text: sqlglot renames some when it renders (`APPROX_COUNT_DISTINCT`
+# comes back as `APPROX_DISTINCT`), and two calls of the same name render
+# identically, so node text cannot be used to find them in the original string.
+_AGGREGATE_NAMES = (
```
jbonofre: `aggregate_spans` (drives export-side decomposition) scan for a hardcoded name list (`COUNT`/`SUM`/ `AVG`/`MIN`/`MAX`/`APPROX_*`), but the fan-out classifier `unsafe_aggregate_datasets`/`_is_aggregate_scope` recognizes any `AggFunc`, `WithinGroup`, or `Anonymous` call.

`STDDEV(customer.ltv)` / `SUM(orders.amount)` across fanned-out relationship leaves `STDDEV` un-decomposed and inlined into the wrong cube's measure, with only a generated APPROXIMATED issue (no `FANNOUT_UNSAFE_METRIC`) even though the per-measure correction can't apply correctly here.

I believe both paths should share the same aggregate-detection logic. Wdyt?

converters/cube/src/ossie_cube/osi_to_cube.py
```diff
@@ -0,0 +1,1711 @@
...
+            part_name = f"{mname}_part_{index}"
+        taken.add(part_name.lower())
+
+        part = _measure_from_expression(
+            piece, part_target, part_name, {}, plan, refs)
+        part["public"] = False
+        part["meta"] = {"ossie": {"part_of": mname}}
+        _place(measures_by_cube, part_target, part, model_name)
+
+        out.append(refs.to_cube_sql(expr[cursor:start], fallback, context=expr))
```
jbonofre, line 1453: The literal glue text between/after aggregate spans in `_decompose_measure` is fed through `refs.to_cube_sql` -> `qualify_bare_columns` -> sqlglot parsing as if it were a standalone expression. Trailing syntax like a bare `END` from an outer `CASE...END` getes misread as a column reference.

`CASE WHEN COUNT(DISTINCT users.id)=0 THEN 0 ELSE SUM(orders.amount)/COUNT(DISTINCT users.id) END` exports with the trailing `END` rewritten to `{CUBE}.END`: it looks like invalid Cube YAML, and re-import can't recover the original expression. It seems to contradict the "lossless round-trip" claim in the README. I believe this is a very common pattern for safe-division ratio metrics, so not really an edge case.

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1863 @@
...
+    def _operand(self, cname, sql, stack, inline_refs):
+        """Translate an aggregate's operand into an Ossie reference.
+
+        A same-cube member or bare column becomes `cube.name` -- the qualified form
+        Ossie model-level metrics use. A computed operand keeps its own qualifiers
+        and is emitted as-is; the owning cube rides in the stash either way, so
+        export still puts the measure back on the right cube.
+        """
+        translated = self._translate(sql, cname, stack, inline_refs).strip()
+        if is_simple_identifier(translated):
```
jbonofre: `_MeasureResolver._operand()` blindly re-qualifies any bare-identifier transaction result as `{cname}.{translated}`, without distinguishing "this is a physical column" or "this is already a resolved same-cube measure/metric reference".

I think a Cube measure `type: sum, sql: "{unit_count}"` (referencing another measure) imports as `SUM(orders.unit_count)`, treating the metric as a column. Re-export produces `sql: '{CUBE}.unit_count'`, referencing a physical column that doesn't exist.

Measure referencing measure is a valid Cube pattern, so this will surface in real models I guess. Wdyt?

jbonofre (commented):
We are getting close! I let a few comments for you.

converters/cube/src/ossie_cube/cube_to_osi.py
MikeNitsenko: `_fanned_out_datasets()` now takes `extra_joins`, normalizing orientation the same way `_convert_joins` does. Values became phrases rather than bare names, since a parked join has no relationship to name - the message for a converted relationship reads identically to before. Parked join now warns, and `--strict-fanout` refuses (exit 1).

converters/cube/src/ossie_cube/osi_to_cube.py
MikeNitsenko: `to_cube_sql()` restructured so the fragment and the whole go through one `convert()` chain; column names come from the converted whole and are only applied to the fragment. `CASE…END` now round-trips exactly; `tax_rate` in glue still qualifies.

converters/cube/src/ossie_cube/expressions.py
MikeNitsenko: Shared, as two predicates: `AggFunc` + `WithinGroup` drives `aggregate_spans`, `Anonymous` stays classifier-only since it can't be told from a scalar UDF. The scanner now confirms candidates against sqlglot, which also fixed `WITHIN GROUP` spans being cut short and `SUM(x) OVER (...)` being split out of its frame.

converters/cube/src/ossie_cube/cube_to_osi.py
MikeNitsenko: You're right, and wider than the operand - `{unit_count} * 2`, `count_distinct` and filters all hit the same nested aggregate. Those are parked with `PARKED_IN_META` and come back verbatim; the bare-identifier qualification is gone, it also turned `sql: NULL` into `SUM(orders.NULL)`.

MikeNitsenko:
@jbonofre all four addressed, plus a few more bugs in the same paths that the pass found and resolved.

jbonofre:
@MikeNitsenko awesome! Thank you again! Let me do a final pass 😄

converters/cube/src/ossie_cube/osi_to_cube.py
```diff
@@ -0,0 +1,1762 @@
...
+        mname = stash.get("name") or sanitize_name(mname_raw, scope, set())
+
+        if "measure" in stash:
+            # The whole-measure record older documents carry (current imports record
+            # only the spellings regeneration would not reproduce); restore it
+            # verbatim and re-inject the natively mapped metadata.
+            measure = dict(stash["measure"])
+            measure["name"] = mname
+            _apply_measure_metadata(metric, measure, stash)
+            target = stash.get("cube") or resolve_base()
```
jbonofre: The metric's target cube is read from the stash (`stash.get("cube") or resolve_base()` / `refs.cube_of_metric(...)`) without validating it against the model's current cube names before being used as the `measures_by_cube` key. If the underlying dataset is renamed after import, `_convert_model`'s per-dataset loop only does `measures_by_cube.get(cname)` against current cube names (the stale-keyed bucket is never retrieved, and the metric silently disappears from output with no issue raised).

Please validate the stashed cube name against `cube_names` and fall back / raise an issue if it no longer matches.

converters/cube/src/ossie_cube/osi_to_cube.py
```diff
@@ -0,0 +1,1762 @@
...
+                f"Relationship '{rname}': from_columns ({len(from_cols)}) and "
+                f"to_columns ({len(to_cols)}) must have the same length")
+
+        stash = read_stash(rel)
+        from_cube = cube_names[rel["from"]]
+        to_cube = cube_names[rel["to"]]
+        declared_on = stash.get("declared_on")
+        relationship = stash.get("relationship", "many_to_one")
+
+        if declared_on == to_cube:
```
jbonofre, line 988: `_build_joins` compares the raw stashed `declared_on` string directly against `to_cube`, instead of resolving both through the same `cube_names` table used for the relationship's from/to. If the declared-on dataset is renamed, `declared_on == to_cube` fails, orientation silently falls back to the default (`own = from_cube`), but the stale `relationship` cardinality is kept as-is, producing a join on the wrong cube with backwards cardinality and no error.

Please resolve `declared_on` through `cube_names` the same way `from_cube`/`to_cube` are resolved.

converters/cube/src/ossie_cube/_common.py
```diff
@@ -0,0 +1,1389 @@
...
+    return f"{CUBE_DIR}/{cube_name}.yml"
+
+
+def view_file(view_name):
+    return f"{VIEW_DIR}/{view_name}.yml"
+
+
+# --- stash protocol -------------------------------------------------------------
+
+def read_stash(obj):
```
jbonofre, line 277: `read_stash` calls `json.loads(ext.get("data") or "{}")` with no try/except and unconditionally trusts any `custom_extensions` entry with `vendor_name == "CUBE"`. Malformed JSON in that field (hand-authored input, or an unrelated tool reusing the same vendor id) crashes with an unhandled `JSONDecodeError`, not caught by `cli.py`'s `except (ConversionError, OSError)`. Non-dict-but-valid JSON (e.g. a list) then raises `AttributeError` on the following `.pop("_v", ...)`.

Please wrap the parse and validate the result is a dict, raising a `ConversionError` on failure.

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1952 @@
...
+    for key in _WINDOWING_KEYS:
+        # An empty rolling_window is meaningful in Cube: it selects the default
+        # trailing-30-day window. The other directives use falsy values to mean off.
+        if measure.get(key) or (key == "rolling_window" and key in measure
+                                and measure[key] is not None):
+            return key
+    return None
+
+
+def _is_generated_part(measure):
```
jbonofre: `_is_generated_part` treats any measure with a truthy `meta.ossie.part_of` as an export-generated decomposition artifact and unconditionally skips it, with no corroboration that a real parent metric exists. A hand-authored Cube measure that happens to carry this key is silently dropped from Ossie output with no issue logged, and can't be recovered on a later re-export, breaking the claimed lossless Cube->Ossie->Cube round-trip.

Please corrobore against an actual parent metric before skipping.

converters/cube/src/ossie_cube/osi_to_cube.py
```diff
@@ -0,0 +1,1762 @@
...
+    # column -- because `primary_key: true` in Cube declares that dimension's own
+    # sql to be the key. A computed dimension would declare the wrong expression,
+    # and a merged geo dimension has no single sql at all, so neither counts even
+    # when its name matches. Anything left uncovered gets a private dimension.
+    pk_names = []
+    computed_keys = set(stash.get("computed_primary_key") or [])
+    taken = {d["name"].lower() for d in dimensions}
+    # `plan.primary_key`, not the dataset's own field: the plan is where the fallback to
+    # `unique_keys` is resolved, and every stage has to agree on the answer.
+    for entry in plan.primary_key:
```
jbonofre, line 352: The primary-key-to-dimension matching loop only checks `by_name_scalar`/`by_column`/`by_name_corrupted`, none of which contain a merged geo dimension's name (geo field `continue` before that classification runs). If a dataset's `primary_key` names a field that was originally a Cube geo dimension, the match fails and falls through to `_unique_pk_dimension_name`, emitting a bogus extra dimension referencing a column that doesn't actually exist, producing an invalid Cube model.

Please include geo dimension names in the PK match before falling back to synthesizing a new dimension.

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1952 @@
...
+    plain = set()
+    for dim in _as_named_list(cube.get("dimensions"), f"cube '{cname}' dimensions"):
+        name = dim.get("name")
+        sql = dim.get("sql")
+        if name and (sql is None or str(sql).strip() == name):
+            plain.add(name)
+    return plain
+
+
+def _primary_key_of(cube, cname):
```
jbonofre, line 624: I see two related points here in `_primary_key_of`:

1. the dimension-flag fallback scan never excludes type: `geo`/`switch`/`sub_query` dimensions, even though none of these produce a same-named Ossie field.
2. if recorded: treats an explicitly stashed empty `primary_key`: `[]` override the same as absent, silently reverting to the flag scan.
       a. produces a dangling PK referemce to a column that doesn't exist (e.g. a geo dimension split into `_latitude`/`_longitude`)
       b. can reintroduce a key the model intentionally left empty.

Please exclude non-plain_column dimension types from the fallback scan, and check for stash-key presence rather than truthiness.

converters/cube/src/ossie_cube/_common.py
```diff
@@ -0,0 +1,1389 @@
...
+    """Coerce an Ossie name into a valid Cube identifier.
+
+    A name that already matches Cube's `identifierRegex` passes through
+    untouched; anything else is lowercased with every invalid character run
+    replaced by `_`. A result colliding case-insensitively with one already in
+    `taken` (a set of casefolded names) is an error rather than a silent merge;
+    the caller adds `result.lower()` to `taken`.
+    """
+    raw = str(name)
+    if _CUBE_NAME_RE.match(raw):
```
jbonofre, line 232: `sanitize_name` and `is_referenceable_name` validate identifiers with `_CUBE_NAME_RE`/`_IDENTIFIER_RE`, both anchored with `$`, which in Python matches immediately before a single trailing `\n`, so a name ending in a newline is wrongly accepted as valid.

`sanitize_name("orders\n", ...)` returns `"orders\n"` unchanged, which then flows into an output file path via `cube_file()`. Similarly `is_referenceable_name("total\n")` returns `True`, splicing a raw newline into generated expression text.

Please anchor with `\Z` (or strip/reject trailing whitespace) instead of `$`.

converters/cube/src/ossie_cube/cube_to_osi.py
```diff
@@ -0,0 +1,1952 @@
...
+    cannot survive natively -- export parks it whole, and it comes back as it went in.
+    A single non-ANSI dialect needs only its label restored.
+    """
+    if parked.get("expression"):
+        target["expression"] = parked["expression"]
+    elif parked.get("dialect"):
+        target["expression"]["dialects"][0]["dialect"] = parked["dialect"]
+
+
+def _convert_geo_dimension(cname, dname, dim, issues):
```
jbonofre, line 962: `_convert_geo_dimension` builds its two Ossie fields directly instead of routing through `_finish_dimension_field` like every other dimension kind, so a geo dimension's `title`/`description`/`meta.ai_context` never becomes native `label`/`description`/`ai_context`. It's swept into an opaque stash  blob instead. Anyone reading the intermediate Ossie document directly (not round-tripping back to Cube) silently loses that metadata, unlike every other field type.

Please route geo field construction through `_finish_dimension_field`.

jbonofre (changes requested):

converters/cube/src/ossie_cube/osi_to_cube.py
MikeNitsenko: Root cause was import-side: `_primary_key_of`'s flag scan included geo/switch/sub_query, so a dimension name ended up in a column list. Fixed there - a geo dimension with `primary_key: true` now round-trips byte-identically. Left the export match alone, since the spec says `primary_key` is column names and matching the geo dim would key on lat/lon instead.

MikeNitsenko:
@jbonofre  all review comments are addressed, and the full suite passes against a built Cube (667, nothing skipped)

converters/cube/src/ossie_cube/_common.py
```diff
@@ -0,0 +1,1408 @@
...
+    Returns ("sql", sql_text) for a SELECT/WITH subquery source, or
+    ("sql_table", table_ref) for a table reference. Cube's `sql_table` takes the
+    reference verbatim (it is interpolated straight into FROM), so no splitting
+    into catalog/schema/table is needed -- unlike Omni, Cube has no separate
+    `schema` key, which also means a bare one-part table name is fine.
+    """
+    if not source or not str(source).strip():
+        raise ConversionError(f"Dataset '{dataset_name}': missing/empty 'source'")
+    s = str(source).strip()
+    if re.match(r"(?i)(select|with)\b", s):
```
jbonofre: I believe a valid SQL source with a leading comment, something like `/* comment */ SELECT ...`, misses this check and is emitted as `sql_table`.

Cube then treats the query as a table reference, so the exported model can fail to compile.

Could we skip leading SQL comments before classifying the source as sql?

WDYT @MikeNitsenko ?

converters/cube/tests/_roundtrip_helpers.py
```diff
@@ -0,0 +1,572 @@
...
+        return '"' + name.upper() + '"'
+    return name
+
+
+def build_ossie_model(rnd):
+    """Generate a hand-authored Ossie model (no stash) as a YAML string."""
+    dim_names = [f"dim_{i}" for i in range(rnd.count(1, 2))]
+    fact = "fact"
+
+    lines = [f"version: {OSSIE_VERSION}", "semantic_model:", "- name: shop"]
```
jbonofre: By the way, this generator still emits the legacy `semantic_model` wrapper, but the current schema expects model fields directly at the document root.

`assert_ossie_is_valid` therefore rejects these generated inputs before the round-trip checks run.

I believe that's why CI is failing.

Could we generate the flat format here and update `check_ossie_model`'s wrapper lookups to match?

kiwamizamurai:
Thank you for working on this

converters/cube/src/ossie_cube/_common.py
MikeNitsenko: Agreed. Leading comments and opening parentheses are now skipped before the source is classified, and the source is emitted unchanged. Fixed in 2555690.

converters/cube/tests/_roundtrip_helpers.py
MikeNitsenko: Right, and it went beyond the generator: the converter itself still read and wrote the wrapper. Both directions, the fixtures and the generator now use the flat format from #383. Fixed in 0678fed.

PR #291, opened 2026-07-30
Automate enum sync for spec files

MonkeyCanCode:
## Summary

The current `core-spec/spec.yaml` has 2 enum sections which would require manual sync from `core-spec/osi-schema.json`. This PR addresses the requested TODO for automating this sync. Also, I removed the TODO annotation and replaced with `# Auto-generated from osi-schema.json (xxxx).`

Things not covered in this PR:
1. CI pipeline that will perform check and raise error if there is a diff detected
- this is due i am not sure how we want to handle dependency for this script (in this case PyYAML). Do we create a different pyproject for it? more detail in [ml](https://lists.apache.org/thread/n4wdmsccosowkvvnq9nko5nq35o6q31k). As the discussion is still on-going, thus I follow the same pattern `validation/validate.py` is using by checking if a package is importable or not.
2. Documentation on where this should be run
- I am not able to find where will be a good spot to document this. I am more toward to make this as part of pre-commit and have it control via make/just instead.

Sample outputs:
```
# helper
(venv) ➜  ossie git:(automate_enum_sync) python3 scripts/generate-spec-types.py
usage: generate-spec-types.py [-h] [--check] [--apply]

options:
  -h, --help  show this help message and exit
  --check     Check spec.yaml is in sync
  --apply     Update spec.yaml in-place

# check (happy path - no drift)
(venv) ➜  ossie git:(automate_enum_sync) python3 scripts/generate-spec-types.py --check
(venv) ➜  ossie git:(automate_enum_sync) echo $?
0

# drift (caused by manual edit)
➜  ossie git:(automate_enum_sync) git diff | tee
diff --git a/core-spec/spec.yaml b/core-spec/spec.yaml
index 32fbb3e..599f2ae 100644
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -36,7 +36,6 @@ dialects:
   - "TABLEAU"               # Tableau
   - "DATABRICKS"            # Databricks SQL
   - "MAQL"                  # GoodData MAQL (Multi-Dimensional Analytical Query Language)
-  - "BIGQUERY"              # Google BigQuery GoogleSQL

 # Supported logical data types for fields and metrics
 # TODO: Generate this list from the authoritative DataType enum in
@@ -50,7 +49,6 @@ datatypes:
   - "Date"                  # Calendar date without time of day
   - "Time"                  # Time of day without a date or timezone
   - "DateTime"              # Date and time without a timezone or offset
-  - "DateTimeTz"            # Instant identified using offset or timezone context
   - "Opaque"                # Known type outside the portable vocabulary

# check (un-happy path - drift)
(venv) ➜  ossie git:(automate_enum_sync)  python3 scripts/generate-spec-types.py --check
--- /Users/yong/Desktop/GitHome/ossie/core-spec/spec.yaml (current)
+++ /Users/yong/Desktop/GitHome/ossie/core-spec/spec.yaml (expected)
@@ -37,6 +37,7 @@
   - "TABLEAU"               # Tableau
   - "DATABRICKS"            # Databricks SQL
   - "MAQL"                  # GoodData MAQL (Multi-Dimensional Analytical Query Language)
+  - "BIGQUERY"              # Google BigQuery GoogleSQL

 # Supported logical data types for fields and metrics
 # Auto-generated from osi-schema.json ($defs.DataType.enum).
@@ -49,6 +50,7 @@
   - "Date"                  # Calendar date without time of day
   - "Time"                  # Time of day without a date or timezone
   - "DateTime"              # Date and time without a timezone or offset
+  - "DateTimeTz"            # Instant identified using offset or timezone context
   - "Opaque"                # Known type outside the portable vocabulary

 # Vendor name for custom extensions (free-form string)
/Users/yong/Desktop/GitHome/ossie/core-spec/spec.yaml is out of sync with /Users/yong/Desktop/GitHome/ossie/core-spec/osi-schema.json. Run --apply to fix.

# status code from un-happy path
(venv) ➜  ossie git:(automate_enum_sync) echo $?
1

# apply fix
(venv) ➜  ossie git:(automate_enum_sync) python3 scripts/generate-spec-types.py --apply
Wrote /Users/yong/Desktop/GitHome/ossie/core-spec/spec.yaml

# validation
(venv) ➜  ossie git:(automate_enum_sync) python3 scripts/generate-spec-types.py --check
(venv) ➜  ossie git:(automate_enum_sync) echo $?
0
```

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance


scripts/generate-spec-types.py
```diff
@@ -0,0 +1,195 @@
...
+SCHEMA = REPO / "core-spec" / "osi-schema.json"
+SPEC = REPO / "core-spec" / "spec.yaml"
+
+SECTIONS = [
+    {
+        "key": "dialects",
+        "schema_path": ["$defs", "Dialect", "enum"],
+        "comment_col": 28,
+        "header": "# Supported expression language dialects",
+        "descriptions": {
```
jbonofre, line 48: The per-value descriptions are hardcoded here. `osi-schema.json` only carries a single `description` per enum (e.g. `"Supported SQL and expression language dialects"`), not per-value text — so today they genuinely can't be read from the schema.

Consequence: when a new value is added to the schema, `--apply` emits it as a bare entry with **no comment**, and someone has to remember to add its description to this dict. I confirmed this by appending `NEWSQL` to the schema:

```yaml
  - "BIGQUERY"              # Google BigQuery GoogleSQL
  - "NEWSQL"
```

So the manual-sync burden isn't eliminated, just relocated. A cleaner long-term design would put per-value descriptions in the schema (single source of truth) and generate both. Not a blocker for this PR — just worth calling out.

scripts/generate-spec-types.py
```diff
@@ -0,0 +1,195 @@
...
+    descs = section["descriptions"]
+    ref = ".".join(section["schema_path"])
+
+    lines = [
+        section["header"],
+        "# Auto-generated from osi-schema.json ({}).".format(ref),
+        "{}:".format(key),
+    ]
+    for v in values:
+        desc = descs.get(v)
```
jbonofre, line 100: At minimum, consider warning to stderr when a schema value has no description in the dict, so the gap is visible instead of silently producing a comment-less entry.

scripts/generate-spec-types.py
```diff
@@ -0,0 +1,195 @@
...
+        entry = '  - "{}"'.format(v)
+        if desc:
+            lines.append('{}{}# {}'.format(entry, " " * (comment_col - len(entry)), desc))
```
jbonofre: Alignment underflow: when an entry is longer than `comment_col` (28), `" " * (comment_col - len(entry))` becomes an empty string and the comment jams against the value (`- "SOMELONGVALUE"# desc`). Not triggered by current values (max length ~10), but fragile — `max(1, comment_col - len(entry))` would harden it.

scripts/generate-spec-types.py
```diff
@@ -0,0 +1,195 @@
...
+        print(f"Error: Generated YAML is invalid: {e}", file=sys.stderr)
+        sys.exit(1)
+    SPEC.write_text(text)
+    print(f"Wrote {SPEC}", file=sys.stderr)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--check", action="store_true", help="Check spec.yaml is in sync")
+    parser.add_argument("--apply", action="store_true", help="Update spec.yaml in-place")
```
jbonofre: `--check` and `--apply` aren't mutually exclusive; passing both silently runs `--apply` (it returns early). A `parser.add_mutually_exclusive_group()` would make the intent explicit.

scripts/generate-spec-types.py
```diff
@@ -0,0 +1,195 @@
...
+    if args.apply:
+        _apply()
+        return
+
+    if args.check:
+        diff = False
+        for section in SECTIONS:
+            expected = _schema_values(section)
+            current = _spec_values(section["key"])
+            if current != expected:
```
jbonofre: `--check` compares enum *values* only, not the comments/formatting. So `--check` can pass while `--apply` would still rewrite the file (e.g. an edited description comment). If this becomes a CI gate meaning "the file is canonical," consider comparing the *rendered* block instead (an `--apply` dry-run + `diff --exit-code`), so description/formatting drift is caught too.

jbonofre (commented):
Thanks for tackling the enum-sync TODO! I ran the script locally: `--check` passes on the committed tree, `--apply` is idempotent, and drift is detected/reconciled correctly. Behavior is solid.

One overall design note plus a few inline comments below. Nothing here is a blocker.

**Main point:** the per-value descriptions are hardcoded in the script rather than sourced from the schema, so this keeps the *values* in sync automatically (which is what the TODO asked for) but moves the *descriptions* sync burden from `spec.yaml` into the script. See the inline comment for details.

scripts/generate-spec-types.py
MonkeyCanCode: For sure, those appears to be added manually assuming as I was not able to find where they were coming from. Now to your point with add those into the schema, we would need to change the schema here. I think we should start a ML on this before making change here as it is a breaking change due to lack of native per enum value description support. In this case, no change will be apply here and I will start a ML for adding this support. WDYT?

scripts/generate-spec-types.py
MonkeyCanCode: This is now fixed.

scripts/generate-spec-types.py
MonkeyCanCode: This is now fixed.

scripts/generate-spec-types.py
MonkeyCanCode: This is fixed.

scripts/generate-spec-types.py
MonkeyCanCode: This is fixed via `difflib.unified_diff`

MonkeyCanCode:
Thanks for the review @jbonofre . 4/5 feedbacks had been addressed. Please let me know if we want to proceed with a ML for the enhancement for the last one.

scripts/generate-spec-types.py
```diff
@@ -0,0 +1,197 @@
...
+            "Opaque": "Known type outside the portable vocabulary",
+        },
+    },
+]
+
+
+def _schema_values(section: dict) -> list[str]:
+    path = section["schema_path"]
+    with open(SCHEMA) as f:
+        data = json.load(f)
```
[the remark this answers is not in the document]
MonkeyCanCode: I think those are fine.

scripts/generate-spec-types.py
```diff
@@ -0,0 +1,207 @@
+ #!/usr/bin/env python3
```
jbonofre: nit: it seems there's a leading space here 😄 


```suggestion
#!/usr/bin/env python3
```

scripts/generate-spec-types.py
jbonofre: That's fair. Let's do like this for now, we can figure it out later.

jbonofre (dismissed):

scripts/generate-spec-types.py
jbonofre: Same point about SIGMA and DAX, but let's move forward on that, we can add this once this PR has been merged.

jbonofre (approved):
At some point, we should create a follow-up PR to run generator as part of the CI 😄

PR #299, opened 2026-08-03
[OSSIE][NVIDIA_GSF] Rename GSF converter directory and package to NVIDIA

liofnv:
Names the converter after the vendor that owns it, matching how the other converter directories are named.

- `converters/gsf` → `converters/nvidia`, `src/ossie_gsf` → `src/ossie_nvidia_gsf`
- Distribution `apache-ossie-gsf` → `apache-ossie-nvidia-gsf`, script `ossie-gsf` → `ossie-nvidia-gsf`
- CI workflow renamed to `converter-nvidia-ci.yml` to keep the `converter-<dir>-ci.yml` convention

No behavior change. All moves are recorded as git renames.


jbonofre (approved):

converters/nvidia/tests/fixtures/sales.gsf.yaml
jbonofre, line 1: FYI (not related to this PR), we should add the ASF header in this file.

converters/nvidia/tests/fixtures/sales.ossie.yaml
jbonofre, line 1: FYI (not related to this PR), we should add the ASF header in this file.

PR #304, opened 2026-08-04
add hex converter

WilliamKelley:
(The *lines changed* metric is a bit inflated by all the apache license file comments)

Disclaimer: This PR description was written completely by me, not AI, so its worth reading (also completely). The code was _not_ all authored by me, but i've revised almost every detail myself.

---

**Bootstrapping:** Sets up the `/converters/hex` directory as a `uv` project with latest `ruff, pytest, syrupy, inline-snapshot`, targeting Python 3.11+. The `CONTRIBUTING.md` outlines the development workflow commands for getting setup, the Python API, the CLI, the project structure, and some useful knowledge. The distributable package in our converter is named `ossie_hex` in the pattern prescribed by Ossie. The `README.md` outlines how public consumers should understand our API surface and conceptual mappings. 

(I’d recommend reading the contributing and readme files along with this PR description).

**Project structure:** Most converters consist of a few files. Typically, one for each Python convert direction, a CLI file, and a common file. The Hex converter follows the same abstract concerns, but splits into a lot of files within each concern. This maps more to how we author in HexSL.

**Hex models:** The `hex_types/` directory encodes Hex’s public semantic spec as Pydantic models. This is a close copy of the models we have authored in our internal semantic layer package.

- Hex Project: a new conception that holds all the resources together in memory along with a name. (NOTE: We should probably establish some project file with a version string, project name/description that’s expected in a hex file directory?)

**Ossie models:** For the most part, Ossie makes its own Pydantic models available for use. There are a couple things we add under the `ossie_types/` directory.

- Expected current version string (trivial)
- Dialect parsing (trivial)
- Patterns and parsing for Ossie field expressions (non-trivial) *— IMO this is something that Ossie should make available to all converters!*

**Conversion surfaces:** There’s a Python surface and a CLI surface that are slightly different. There’s no specific standard prescribed by Ossie about what each respective API should input/output, but most converters take in some file contents text and output file contents text.

**CLI:** Commands `import` and `export`. Accepts file paths, reads/writes from disk. The `cli/` directory.

**Python:** Functions `convert_hex_to_ossie` (import) and `convert_ossie_to_hex` (export). Accepts and returns file(s) contents.

**Conversion logic:** It’s useful to understand the different structures of each spec and the two primary aims of a converter (in my opinion).

```markdown
OSIDocument
  `n` OSISemanticModel
    `m` OSIDataset
      `j` OSIField
    `k` OSIRelationship
    `l` OSIMetric
    
HexProject
  `n` HexModel
    `j` HexDimension
    `k` HexMeasure
    `l` HexRelation
  `m` HexView
```

**Primary Aim 1:** A lossless roundtrip from **Vendor → Ossie → Vendor**. This is explicitly described by the Ossie converter contributing guide. Vendor-specific concepts that Ossie does not prescribe get stored in vendor-specific metadata so that they can be restored later.

This governed my mindset initially, so I authored in that order: import hex → ossie, then export ossie → hex.

Import felt a bit weird. Basically we have to compile-*down* a lot of our Dimension/Measure “features” into SQL and compile-*up* our Relations into an abstract syntax. I spent a while on what metadata needed to be preserved and to do so simply. One decision I made is that **we should aim for roundtrip input/output to be equivalent in eventual consumption, but not to produce exact same file content representation.** Otherwise we’d basically store everything as custom Vendor data.

After I achieved this aim though, I was thinking… what’s the point? Who really cares about converting *back* to your own spec? The whole point of this is inter-op. That lead me to…

**Primary Aim 2:** Interpreting Ossie models that have *no* Hex-specific metadata into Hex models.

This is the *actually useful* part of the converter, and it’s basically HexSL. It’s a lot of parsing expressions to figure out what is being referenced (dataset fields) and then attributing metrics to model measures and relationships to necessary model relations. There’s a lot of ambiguity that’s not easily worked-around.

**Import (Hex → Ossie):** The `hex_to_ossie/` directory.

- Hex doesn’t encode a project name or SQL dialect in its persisted file format. So, import takes these as inputs. The SQL dialect is required since the Ossie spec encodes it directly into the document.
- The Hex project files are read and loaded in the pydantic models. Views are stashed in our custom extension (see next section). Models are the primary conversion logic.
- For each model,
    - Hex Relations → Ossie Relationship: converted first because they’re necessary in order to qualify Hex Dimension expressions into Ossie Field expressions. (A Hex Dimension can reference a named Hex Relation, but Ossie Fields reference underlying Ossie Dataset names, not Ossie Relationship names).
    - Hex Dimensions → Ossie Fields, primary/unique keys: converted into expression syntax
    - Hex Measures → Ossie Metrics: converted into expression syntax.
    - “References”, i.e. referencing another semantic concept: Hex defines this explicitly as “SQL Interpolation” with a syntax boundary `${}` . Ossie does not define this concept or give a syntax boundary. On import, the Hex construction provides more option for unique values with implicit model-scope and named relations. On compiling to Ossie’s construction in the global-scope, collisions can occur and that’s a bit awkward.
    - Hex Model → Ossie Dataset: models & dimensions stay grouped together as datasets & fields. However, relations and measures get promoted to the global namespace as relationships and metrics.
- Unsupported concepts are stashed in the custom extension data.

**Custom extension:** Information that Hex encodes and needs for a lossless roundtrip are stored in the `HEX` vendor data according to Ossie’s standard.

- Hex Relations can express arbitrary SQL while Ossie Relationships are a piecemeal construction of an equality statement. So relations that we cannot parse / decompose into this form are stashed.
- Hex Relations express the join type explicitly while Ossie Relationships express this implicitly. It’s mostly possible to restore this on roundtrip, but the join type is preserved.
- Hex calculation formulas: Hex Dimensions and Measures that use Hex’s calc formula syntax are stashed. We could add support for this if we want to expose our calc parsing logic.
- Hex Measures that are semi-additive are still valid to express as an Ossie Metric but their semi-additivity is stashed.
- Visibility and (Display) Names are stashed.
    - Ossie Fields have this weird `label` field that I don’t think Ossie intended to be a display name field, but some converters are using it as such so I did as well.

Unlike other converters, I added a version string and pydantic models for our extension data so that we’re careful when making future changes. I’m not sure how consumers will use converters in the wild, but it’s possible that people check-in converted documents to source, so an output of one version gets consumed as input for another version. 

**Export (Ossie → Hex):** The `ossie_to_hex/` directory.

- Hex only allows a single SQL expression while an Ossie Document can specify multiple across dialects. So, export allows to choose which dialect expression to pick (or defaults).
- An Ossie Document can specify >1 Ossie Semantic Model, so the name of the one to choose has to be given if there are multiple. Conceptually one Ossie Semantic Models equates to a Hex Project.
- Ossie Relationships → Hex Relations: convert first (like import) because their construction governs how fields/metrics get converted.
- Ossie Dataset, Fields → Hex Model, Dimensions: map pretty naturally, converting expression syntax.
- Ossie Metrics → Hex Measures: have to be attributed to a reasonable dataset given the relations.

**Known flaws:**

Import

- When we drop dimensions and measures with Hex calculation formulas (`dim.expr_calc, measure.func_calc`), we don’t also go around and drop dimensions, measures, relations that depend on the dropped things. Doing so would require even more SQL parsing and I’m already uncomfortable with the amount of that I authored in this.

**Publishing**: As noted at the end of the contributing doc, I do not think we, Hex, are in charge of, or able to, publish our own converter. We give away license and control to Apache Ossie. Pretty sure we’re subject to their timeline(s). And I’m pretty sure nothing is published right now (the main ossie package or any vendor’s converter) since i can’t reach anything on PyPi.


WilliamKelley:
Apologies, meant to open this on our organization's fork for internal review first.

PR #307, opened 2026-08-05
Validate relationship column pair counts

LiangRuoShui:
## Why

The OSI spec requires `from_columns` and `to_columns` in a relationship to have the same number of columns so each source column maps to one target column. The validator checked that relationship datasets existed, but did not enforce this cardinality rule, so invalid relationship definitions could pass validation and be silently truncated by downstream converters that zip the two arrays.

## What changed

- Added a relationship-level validation check in `validation/validate.py` for mismatched `from_columns` / `to_columns` lengths.
- Report mismatches with a clear `[Relationship]` error that includes the relationship name, model name, and both column counts.
- Added validation unit tests for both mismatched and matching relationship column arrays.

## Tests added/updated

- `validation/tests/test_validate.py::test_validate_references_rejects_mismatched_relationship_column_counts`
- `validation/tests/test_validate.py::test_validate_references_accepts_matching_relationship_column_counts`

## Verification

- `uv run --with pytest --with pyyaml --with jsonschema --with sqlglot pytest validation/tests/test_validate.py -q`
- `uv run validation/validate.py examples/tpcds_semantic_model.yaml`
- Verified an invalid temporary OSI document now fails validation with `2 from_columns but 1 to_columns`.

### DIFF
--- a/validation/tests/test_validate.py
+++ b/validation/tests/test_validate.py
@@ -21,7 +21,7 @@ from pathlib import Path
 import pytest
 
 # validate.py exits at import time when its dependencies are missing, which
-# would abort the whole pytest session during collection — skip instead.
+# would abort the whole pytest session during collection -- skip instead.
 pytest.importorskip("yaml")
 pytest.importorskip("jsonschema")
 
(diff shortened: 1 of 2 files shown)


LiangRuoShui:
Closing this PR for now. Thank you for the time and review.

PR #308, opened 2026-08-05
Reject mismatched relationship columns in converters

LiangRuoShui:
## Summary
- Add fail-fast validation in the GoodData and Wisdom Ossie exporters when relationship `from_columns` and `to_columns` have different lengths.
- Prevent Python `zip()` from silently truncating composite join columns and emitting incomplete target joins/references.
- Add regression tests for mismatched relationship column arrays in both converters.

## Why
Ossie relationships define column pairs positionally: each `from_columns[n]` must correspond to `to_columns[n]`. When the arrays differ in length, the relationship is malformed and cannot be faithfully converted. GoodData and Wisdom previously used `zip()` directly, which dropped extra columns without an error.

## Test Plan
- `cd converters/gooddata && uv run pytest -q` -> 75 passed
- `cd converters/wisdom && uv run pytest -q` -> 28 passed
- `git diff --check`


LiangRuoShui:
Closing this PR for now. Thank you for the time and review.

PR #310, opened 2026-08-09
Add SAP Business Partner example for the ontology layer

marioapiphani:
`examples/` has one ontology-layer example today (`flights.yaml`). This adds a second, drawn from an enterprise ERP domain, that exercises ontology constructs the existing example does not reach:

- **entity subtyping** — `Customer` declares `extends: [ BusinessPartner ]`; `flights.yaml` has no EntityType subtyping
- **a derived population** — `Customer` uses concept-level `derived_by` to define its population as exactly those business partners carrying a customer number, rather than asserting it. `flights.yaml` uses `derived_by` only on relationships, for derived attributes.
- **four-part compound identity** — `CustomerSalesArea` is identified by customer + sales organization + distribution channel + division. The existing example tops out at two parts.

It also covers ground shared with `flights.yaml` — value types with `requires`, associations with and without functional multiplicity, verbalizations in both reading directions, and `ontology_mappings` down to a logical model — so it stands alone as a reading example.

Source-system provenance (SAP Data Dictionary domains and check tables) is recorded in the descriptions, showing one way to keep technical origin alongside a business-level model.

The README's `examples/` entry is updated to name both ontology-layer models.

The file validates against `ontology/ontology.json` as shipped, with `core-spec/osi-schema.json` registered to resolve the remote `$ref` from the checkout:

```python
import json, re, yaml
from jsonschema import Draft202012Validator
from referencing import Registry, Resource
onto = open("ontology/ontology.json").read()
core = Resource.from_contents(json.load(open("core-spec/osi-schema.json")))
reg = Registry().with_resources([(u, core) for u in set(re.findall(r'"\$ref": "(https://[^"#]+)', onto))])
v = Draft202012Validator(json.loads(onto), registry=reg)
print(sorted(v.iter_errors(yaml.safe_load(open("examples/sap_business_partner_ontology.yaml"))), key=lambda e: list(e.path)) or "VALID")
```

Prints `VALID`. No code changes.


jochenchrist:
Could you provide the original data model?

jochenchrist:
Also curious, if there are some properties missing in Ossie?

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+    type: EntityType
+    description: >
+      A business partner in its customer role. Source entity: Customer
+      (I_Customer). Modeled as a derived subtype: rather than asserting its
+      population, `derived_by` defines it as exactly those business partners
+      that carry a customer number. This is how SAP's role concept reads under
+      a closed-world assumption.
+    extends: [ BusinessPartner ]
+    derived_by:
+      - "EXISTS ( BusinessPartner.customer_nr )"
```
jbonofre, line 187: `derived_by` requires `EXISTS ( BusinessPartner.customer_nr) `, but `customer_nr` has no `link_mapping` anywhere in the `BusinessPartner` nor `CUSTOMER_SALES_AREA` has a field that could feed it.

Running this file's own mapping yields zero `Customer` objects, so the derived-population feature this example is meant to demonstrate can't actually be observed.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+    relationships:
+      - name: has_sales_area
```
jbonofre: There's no `concept: Customer` entry anywhere in `concept_mappings`, only `BusinessPartner` and `CustomerSalesArea` are mapped.

`has_sales_area` gets zero link_mappings and zero derived_by, so this relationship (meant to illustrate "association without functional multiplicity") can never by populated.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+    identify_by: [ customer, sales_org, channel, division ]
+    requires:
+      - "CustomerSalesArea.customer"
+      - "CustomerSalesArea.sales_org"
+      - "CustomerSalesArea.channel"
+      - "CustomerSalesArea.division"
+    relationships:
+      - name: customer
+        roles:
+          - concept: Customer
```
jbonofre, line 211: These look like the same real-world relationship modeled as two independent declarations instead of one dual-verbalized relationship, the pattern `ontology.md`'s `Person.parent_of` uses.

That means keeping them in sync by hand, and as noted above, the `Customer.has_sales_area` side isn't even mapped.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+        multiplicity: ManyToOne
+        verbalizes:
+          - "{CustomerSalesArea} sells via distribution channel {DistributionChannel}"
+      - name: division
+        roles:
+          - concept: Division
+        multiplicity: ManyToOne
+        verbalizes:
+          - "{CustomerSalesArea} covers division {Division}"
+      - name: delivery_blocked
```
jbonofre, line 234: `DELIVERY_BLOCK` exists in the `CUSTOMER_SALES_AREA` dataset specifically for this, but `CustomerSalesArea`'s concept_mapping stops after `division`, there's no `delivery_blocked` link_mapping. Compare `flights.yaml`'s `canceled` flag, mapped via `FLIGHT.id WHERE ( FLIGHT.cancelled == TRUE )`, for the pattern this is missing.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+          - referent_mappings:
+              - relationship: nr
+                expression: "BUSINESS_PARTNER.BP_NUMBER"
+        link_mappings:
+          - object_mapping:
+              referent_mappings:
+                - relationship: nr
+                  expression: "BUSINESS_PARTNER.BP_NUMBER"
+            children:
+              - relationship: BusinessPartner.category
```
jbonofre: These `LinkMapping.relationship` values are concept-qualified, but `ontology.md`'s worked example and every occurence in `flights.yaml` use bare relationship names for `LinkMapping.relationship` (concept-qualification is for `ReferentMapping.relationship` instead).

If the mapping engine resolves by exact local name, `category`/`full_name`/`country` would never resolve and stay unpopulated.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+
+  - concept: CountryCode
+    type: ValueType
+    description: >
+      Country key. Technical origin: DDIC domain LAND1, check table T005.
+      The check-table relationship is preserved by mapping this concept's
+      population from a country dataset rather than enumerating values here.
+    extends: [ String ]
+    requires:
+      - "LENGTH(CountryCode) = 2"
```
jbonofre: This is the only bare `=` in the file/corpus, every sibling constraint uses `<=` and `ontology.md`/`flights.yaml` use `==` exclusivity for equality. Worth double-checking the ontology-layer expression dialect accepts bare `=`. Separately, SAP's LAND1 field is commonly CHAR(3), so an exact length-2 constraint would reject legitimate 3-character country codes.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+              - relationship: BusinessPartner.country
+                object_mapping:
+                  expression: "BUSINESS_PARTNER.COUNTRY"
+
+      - concept: CustomerSalesArea
+        object_mappings:
+          - referent_mappings:
+              - relationship: customer
+                referent_mappings:
+                  - relationship: BusinessPartner.nr
```
jbonofre, line 390: `Customer` declares no `identify_by` of its own, and neither the spec nor the schema says whether `identify_by` is inherited via `extends` ("inherit" doesn't appear anywhere in `ontology.md`/`spec.md`/the JSON schemas).

Worth flagging that against the actual converter, `identify_by` is built only from a concept's own YAML list, so `Customer.identify_by` resolved `{}` here. 

On top of that, `lookup_concept_relationship` always prefixes `f"{concept.name}.{name}"`, so an already-qualified value like `BusinessPartner.nr` would get double-prefixed and fail to resolve. This isn't specific to this file (it'd hit `ontology.md`'s own `CustOrder.nr` example too), but this PR looks like the first example to actually exercise this path.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+      Identifier of a business partner in its customer role. Technical origin:
+      DDIC domain KUNNR (CHAR 10, ALPHA).
+    extends: [ String ]
+    requires:
+      - "LENGTH(CustomerNumber) <= 10"
+
+  - concept: CountryCode
+    type: ValueType
+    description: >
+      Country key. Technical origin: DDIC domain LAND1, check table T005.
```
jbonofre: This says population comes "from a country dataset" preserving the T005 check-table relationship, but no such dataset exists in `ontology_mappings` (the actual mapping just copies `BUSINESS_PARTNER.COUNTRY` straight through.

It might be worth correcting the description or adding the dataset it describes.

examples/sap_business_partner_ontology.yaml
```diff
@@ -0,0 +1,349 @@
...
+description: >
+  An ontology-layer model of the SAP Business Partner domain, covering business
+  partners, the customer role derived from them, and the customer master data
+  maintained per sales area.
+
+ontology:
+
+  # ---- Value types carrying DDIC domain semantics -------------------------
+
+  - concept: BusinessPartnerNumber
```
jbonofre, line 55: The six sibling `ValueType.requires` blocks mix block-list and inline-flow YAML style for structurally identical constraints. Given this file is meant as a template for future SAP fields, a consistent style would help (and might have caught the `= 2` issue 😄 ).

marioapiphani:
@jbonofre **Status.** `366f1df` is schema-valid against `ontology/ontology.json`, with parser incompatibilities and unresolved semantics. This is a request for review of the revision, not a claim that its runtime behaviour has been verified.

`converters/ontology` at `ae3ee40` rejects the file before reaching the mappings, on `datatype` for dataset fields (`SpecDatasetField` is `extra="forbid"`). The original `51a93cb` fails first on the same field; because that failure stops parsing, it says nothing about what either version would do further in. With `datatype` removed, two further constructs fail: reverse-direction verbalizations (`_parse_verbalization` matches tokens by position), and the qualified nested referent you identified (`'Customer.BusinessPartner.nr'`, `converter.py:323`).

Both of those can be reproduced from the repo without this file: parse the `Person.parent_of` snippet from `ontology/ontology.md` (~line 289) and you get `Role 0: 'Person:Person' does not match verbalization role 'Person:child'`; parse the `OrderLineItem` / `CustOrder.nr` snippet (~lines 466-497) and you get `unknown relationship 'CustOrder.CustOrder.nr'`, which resolves once `CustOrder.nr` is written as bare `nr`. For `datatype`, any dataset field carrying `datatype: String` is enough.

No CI workflow reads this file (`converter-ontology-ci.yml` triggers on `converters/ontology/**` only), so a green check is not evidence about it.

**Completed in the file:**
1. `CUSTOMER` dataset (I_Customer-shaped: `CUSTOMER`, `BUSINESS_PARTNER`) feeding `BusinessPartner.customer_nr`.
2. `Customer` concept mapping: object mapping via `CUSTOMER.BUSINESS_PARTNER`, link mapping feeding `customer_nr` from `CUSTOMER.CUSTOMER`.
3. One `CustomerSalesArea.customer` relationship, verbalized in both directions, replacing the two declarations.
4. `delivery_blocked` mapped with `… WHERE ( CUSTOMER_SALES_AREA.DELIVERY_BLOCK <> '' )`, after `canceled` in `flights.yaml`. LIFSD is CHAR 2, blank when not blocked, so the field stays a string.
5. Bare `LinkMapping.relationship` names; `ReferentMapping.relationship` stays qualified.
6. `LENGTH(CountryCode) <= 3`; `==` throughout; the `IN (…)` constraint rewritten as an `OR` chain to match the corpus.
7. `identify_by: [ nr ]` declared on `Customer`.
8. `CountryCode` description says the value comes from `BUSINESS_PARTNER.COUNTRY`; T005 recorded as provenance, not modelled.
9. `requires` blocks in block-list style.

**Verification, blocked.** On a copy of the committed file with exactly this change:

```diff
-          - "{BusinessPartnerNumber} identifies {BusinessPartner}"
-          - "{Customer} sells through {CustomerSalesArea}"
-              datatype: String            (11 occurrences)
-                  - relationship: BusinessPartner.nr
+                  - relationship: nr        (2 occurrences)
```

the engine resolves: `Customer.identify_by` `[]` → `['BusinessPartner.nr']`; `Customer` object mappings 0 → 1; `customer_nr`, `delivery_blocked` and customer↔sales-area unfed → fed; 10/10 relationships fed. These describe the modified copy, not the committed file. They are resolved references, not object counts, and the engine carries `WHERE` expressions unparsed, so they do not validate the filter expressions.

**Two decisions needed from you to finish:**

(a) `Customer.identify_by: [ nr ]` names a relationship declared on `BusinessPartner`. `ontology.md` defines `identify_by` without addressing `extends`, and its own subtype examples declare none; the engine resolves it by walking `extends`. Is naming an inherited relationship the intended form?

(b) With a four-part identifier, `delivery_blocked` repeats the `WHERE` predicate on each identifying referent. `ontology.md` defines no row-level filter; the only corpus instance is `flights.yaml`, on a one-part identifier. I cannot confirm from the spec that this form is correct.

I kept the reverse verbalization you asked for in #3 although the parser rejects it; removing it would change what the example demonstrates. If you would rather the file parse as-is today, that is your call and I will change it.

---

@jochenchrist, two answers I owe you from August.

**The original data model.** The source is a CAP-style CSN document: four definitions from the S/4HANA Business Partner family, trimmed to this slice. `sap.slice.BusinessPartnerCategory` is a `cds.String(1)` type with the fixed values 1/2/3 (domain BU_TYPE); `sap.slice.BusinessPartner` has key `BusinessPartner` (CHAR 10, BU_PARTNER), `BusinessPartnerCategory`, `BusinessPartnerFullName`, `Country` (LAND1, check table T005) and a to-one association `_Customer` joined on the partner's own key; `sap.slice.Customer` is the I_Customer role view with key `Customer` (KUNNR), element `BusinessPartner`, and a to-many association `_SalesArea`; `sap.slice.CustomerSalesArea` is KNVV-class with the four-part key `Customer` + `SalesOrganization` (VKORG) + `DistributionChannel` (VTWEG) + `Division` (SPART), `DeliveryIsBlocked` (LIFSD), and a to-one `_Customer` joined on the customer key. The full CSN is below so the pair is reviewable side by side; the ontology's descriptions name the DDIC domain behind each value type for the same reason.

<details>
<summary>sap-bp-slice.csn.json</summary>

```json
{
  "$comment": "SAP Business Partner slice — CSN source fixture. Pairs with sap-bp-slice.ossie.yaml (which validates against ontology/ontology.json). Together they form the round-trip fixture pair for a converters/python/sap-csn leg, in the style of the Databricks converter's fixture pairs. CAP-style CSN, trimmed to the slice's scope.",
  "csnInteropEffective": "1.0",
  "definitions": {

    "sap.slice.BusinessPartnerCategory": {
      "kind": "type",
      "type": "cds.String",
      "length": 1,
      "@EndUserText.label": "Business Partner Category",
      "enum": {
        "Person":       { "val": "1" },
        "Organization": { "val": "2" },
        "Group":        { "val": "3" }
      }
    },

    "sap.slice.BusinessPartner": {
      "kind": "entity",
      "@EndUserText.label": "Business Partner",
      "@ODM.entityName": "BusinessPartner",
      "elements": {
        "BusinessPartner": {
          "key": true,
          "type": "cds.String",
          "length": 10,
          "@EndUserText.label": "Business Partner Number",
          "@Semantics.businessPartner.id": true
        },
        "BusinessPartnerCategory": {
          "type": "sap.slice.BusinessPartnerCategory",
          "@EndUserText.label": "Business Partner Category"
        },
        "BusinessPartnerFullName": {
          "type": "cds.String",
          "length": 81,
          "@EndUserText.label": "Full Name"
        },
        "Country": {
          "type": "cds.String",
          "length": 3,
          "@EndUserText.label": "Country Key",
          "$comment": "Domain LAND1, check table T005 — value set maintained as data, not schema"
        },
        "_Customer": {
          "type": "cds.Association",
          "target": "sap.slice.Customer",
          "cardinality": { "max": 1 },
          "on": [ { "ref": ["_Customer", "BusinessPartner"] }, "=", { "ref": ["BusinessPartner"] } ],
          "@EndUserText.label": "Customer role"
        }
      }
    },

    "sap.slice.Customer": {
      "kind": "entity",
      "@EndUserText.label": "Customer",
      "@ODM.entityName": "Customer",
      "$comment": "Role view (I_Customer / FLCU00-class): a business partner acting as customer",
      "elements": {
        "Customer": {
          "key": true,
          "type": "cds.String",
          "length": 10,
          "@EndUserText.label": "Customer Number",
          "$comment": "Domain KUNNR (CHAR 10, ALPHA)"
        },
        "BusinessPartner": {
          "type": "cds.String",
          "length": 10,
          "@EndUserText.label": "Business Partner Number"
        },
        "_SalesArea": {
          "type": "cds.Association",
          "target": "sap.slice.CustomerSalesArea",
          "cardinality": { "max": "*" },
          "on": [ { "ref": ["_SalesArea", "Customer"] }, "=", { "ref": ["Customer"] } ],
          "@EndUserText.label": "Sales Areas"
        }
      }
    },

    "sap.slice.CustomerSalesArea": {
      "kind": "entity",
      "@EndUserText.label": "Customer Sales Area Data",
      "$comment": "KNVV-class: customer master data per sales area; four-part compound key",
      "elements": {
        "Customer": {
          "key": true,
          "type": "cds.String",
          "length": 10,
          "@EndUserText.label": "Customer Number"
        },
        "SalesOrganization": {
          "key": true,
          "type": "cds.String",
          "length": 4,
          "@EndUserText.label": "Sales Organization",
          "$comment": "Domain VKORG, check table TVKO"
        },
        "DistributionChannel": {
          "key": true,
          "type": "cds.String",
          "length": 2,
          "@EndUserText.label": "Distribution Channel",
          "$comment": "Domain VTWEG, check table TVTW"
        },
        "Division": {
          "key": true,
          "type": "cds.String",
          "length": 2,
          "@EndUserText.label": "Division",
          "$comment": "Domain SPART, check table TSPA"
        },
        "DeliveryIsBlocked": {
          "type": "cds.Boolean",
          "@EndUserText.label": "Delivery Block",
          "$comment": "Reduced from LIFSD fixed-value domain to a flag for the slice; the release action is behavior, not structure"
        },
        "_Customer": {
          "type": "cds.Association",
          "target": "sap.slice.Customer",
          "cardinality": { "max": 1 },
          "on": [ { "ref": ["_Customer", "Customer"] }, "=", { "ref": ["Customer"] } ],
          "@EndUserText.label": "Customer"
        }
      }
    }
  }
}
```

</details>

**What is missing in Ossie, from this slice.** Structure converts cleanly: entities, simple and compound identity, domains with length and fixed values, associations in both directions, and the role view as a derived subtype. What has no home yet:

- Annotations other than the label. `@ODM.entityName`, `@Semantics.*`, `@ObjectModel.*` and value-help annotations are dropped: `OntologyComponent`, `Relationship` and `Role` are `additionalProperties: false` with no `custom_extensions`, and `ai_context` exists only at the document root.
- The association `ON` predicate. A relationship carries no join condition, so the predicate is read for its shape (which side's key it references) and then lost.
- `cds.Composition` versus association: containment is not distinguishable.
- Cardinality `min`. `Multiplicity` offers `ManyToOne` and `OneToOne`; optional-versus-mandatory to-one has no field, which is why the compound key carries a `requires` per part.
- `precision` and `scale` on decimals, and `length` on non-string types: no portable constraint form.
- Language-dependent texts. Every SAP label exists in dozens of languages keyed by language; the spec has no multi-language keys.
- Actions and events (the delivery-block release, for instance): structural layer only today, which is why `DeliveryIsBlocked` is reduced to a unary flag here.
- Hierarchies, time-dependent validity and authorization semantics: deliberately out of this slice, and also without a spec construct.

Conversion routines (ALPHA leading zeros) are recorded in descriptions on purpose; they are representation, not meaning.

jbonofre:
@marioapiphani thanks for the update! Let me do a new pass.

PR #312, opened 2026-08-09
fix(dbt): reject composite keys

### LINKED ISSUE #311: dbt osi-to-msi silently corrupts composite primary and unique keys
## Problem

The dbt `OSIToMSIConverter` flattens every Ossie key group into a set of columns. For a composite primary key such as `[tenant_id, order_id]`, it emits two independent MetricFlow `PRIMARY` entities. The reverse conversion then keeps only the final primary entity, silently changing the key to `[order_id]`. Composite unique keys are likewise converted into independent singleton unique keys.

## Reproduction

On `main` at `88e0011148283302c9a04cd0287e00e0b9d87354`, convert a valid dataset containing:

```yaml
primary_key: [tenant_id, order_id]
```

Observed after Ossie -> MSI:

```text
entities=[(tenant_id, primary), (order_id, primary)]
```

Observed after Ossie -> MSI -> Ossie:

```yaml
primary_key: [order_id]
```

A singleton primary key round-trips unchanged.

## Root cause

`converters/dbt/src/ossie_dbt/osi_to_msi.py::_build_key_sets` converts `primary_key` to a set and flattens all `unique_keys` groups. `_classify_field` then treats every member as independently primary or unique. `MSIToOSIConverter._extract_keys` has no grouping information to reconstruct.

## Impact

Composite keys are common for tenant-scoped and line-item datasets. Emitting each component as independently unique can produce incorrect join cardinality or aggregations, while round trips silently lose key components.

## Proposed fix

Until MSI can represent and round-trip grouped keys losslessly, fail Ossie -> MSI conversion clearly when a dataset contains a composite primary or unique key. Keep singleton key behavior unchanged and add regression tests for both composite key types. This prevents generating a plausible but semantically incorrect manifest.

mattfaltyn:
## Summary

- Reject composite primary and unique keys before Ossie-to-MSI conversion can flatten their grouped semantics.
- Preserve existing singleton primary and unique key behavior.
- Document the unsupported input and cover both composite key forms with regression tests.

This replaces plausible but incorrect MetricFlow output with an explicit `ValueError` until grouped keys have a lossless MSI representation.

## Related Issues

Closes #311

### Specification
- [x] No specification changes

### Converters
- [x] Regression tests cover composite primary and unique keys

### Documentation
- [x] The dbt conversion notes document the rejected input

### Tests
- [x] `uv run --python 3.11 --frozen pytest -q` — 101 passed, 5 snapshots
- [x] `uv run --frozen pytest -q` on Python 3.12 — 101 passed, 5 snapshots
- [x] `uv run --python 3.14 --frozen pytest -q` — 101 passed, 5 snapshots
- [x] CLI reproduction exits nonzero and writes no manifest

### Compliance
- [x] No dependencies or new files added

### DIFF
--- a/converters/dbt/README.md
+++ b/converters/dbt/README.md
@@ -119,8 +119,9 @@ manifest_json = result.output.model_dump_json(by_alias=True, exclude_none=True, 
 | `NATURAL_ENTITY_DROPPED` | Ossie has no natural-key entity type |
 | `CUMULATIVE_SEMANTICS_LOSS` | Window/grain semantics cannot be expressed in an Ossie expression string; the base aggregation is preserved |
 
-**Ossie → MSI** reconstructs a best-effort MSI manifest from Ossie's simpler schema. Nothing is dropped, but Ossie carries less structural information than MSI, so the converter makes the following choices:
+**Ossie → MSI** reconstructs a best-effort MSI manifest from Ossie's simpler schema. Nothing is dropped for supported inputs, but Ossie carries less structural information than MSI, so the converter makes the following choices:
 
+- Composite primary and unique keys are rejected because MSI entities cannot preserve grouped key semantics
 - Single aggregations (`SUM(col)`, `COUNT(DISTINCT col)`, etc.) → SIMPLE metric with `metric_aggregation_params`
 - `(expr_a) / (expr_b)` → RATIO metric with auto-generated sub-metrics
 - Anything else → SIMPLE metric with the raw expression stored verbatim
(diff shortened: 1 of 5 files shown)


QMalcolm (commented):
@mattfaltyn It looks like this PR currently conflicts with main. Could you rebase and update the PR? Thank you for doing this work ❤️

mattfaltyn:
> @mattfaltyn It looks like this PR currently conflicts with main. Could you rebase and update the PR? Thank you for doing this work ❤️

Thanks for flagging this, @QMalcolm! I’ve rebased the branch onto the latest main and force-pushed the update. The dbt converter suite is green: 102 passed, including 5 snapshots. I appreciate the review and the kind note.

QMalcolm (approved):
Thanks again, LGTM!

PR #315, opened 2026-08-11
Fix DBT converter round-trip losses entity issue

MonkeyCanCode:
## Summary

Currently  a primary, unique, and foreign entity whose expression differs from its name is silently lost on MSI → OSI → MSI conversion. This is due to the following:

From `converters/dbt/src/ossie_dbt/msi_to_osi.py`, we have following:
```
    @staticmethod
    def _extract_keys(entities: Sequence[Entity]) -> Tuple[Optional[List[str]], List[List[str]]]:
        primary_key: Optional[List[str]] = None
        unique_keys: List[List[str]] = []

        for entity in entities:
            col = entity.expr if entity.expr is not None else entity.name
            if entity.type is EntityType.PRIMARY:
                primary_key = [col]
            elif entity.type is EntityType.UNIQUE:
                unique_keys.append([col])

        return primary_key, unique_keys
```

However, from `converters/dbt/src/ossie_dbt/osi_to_msi.py` following:
```
        if field.name in primary_key_cols:
            entities.append(
                PydanticEntity(
                    name=field.name,
                    type=EntityType.PRIMARY,
                    expr=expr_or_none,
                    description=field.description,
                    label=field.label,
                    role=None,
                    config=None,
                )
            )
            return
        if field.name in unique_key_cols:
            entities.append(
                PydanticEntity(
                    name=field.name,
                    type=EntityType.UNIQUE,
                    expr=expr_or_none,
                    description=field.description,
                    label=field.label,
                    role=None,
                    config=None,
                )
            )
            return
        if field.name in foreign_key_cols:
            entities.append(
                PydanticEntity(
                    name=field.name,
                    type=EntityType.FOREIGN,
                    expr=expr_or_none,
                    description=field.description,
                    label=field.label,
                    role=None,
                    config=None,
                )
            )
            return
```

Here is the local test result with the fixed code and newly added test:
```
➜  dbt git:(fix_dbt_roundtrip) ✗ uv run pytest
========================================================================== test session starts ==========================================================================
platform darwin -- Python 3.11.13, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/yong/Desktop/GitHome/ossie/converters/dbt
configfile: pyproject.toml
testpaths: tests
plugins: syrupy-5.5.3
collected 100 items

tests/test_msi_to_osi.py ..................................................................                                                                       [ 66%]
tests/test_osi_to_msi.py ..................................                                                                                                       [100%]

------------------------------------------------------------------------ snapshot report summary ------------------------------------------------------------------------
5 snapshots passed.
========================================================================== 100 passed in 0.34s ==========================================================================
```

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance


MonkeyCanCode:
@QMalcolm mind take a look?

converters/dbt/src/ossie_dbt/ossie_to_msi.py
```diff
@@ -188,7 +188,7 @@ def _classify_field(
         Aggregation info lives on metrics (`metric_aggregation_params`), not on
         semantic model measures, so there is no measure classification step.
         """
-        if field.name in primary_key_cols:
+        if field.name in primary_key_cols or expr in primary_key_cols:
```
jbonofre, line 191: I think that matching by `expr` membership (in addition to `name`) opens a false-positive collision risk: if an unrelated field's `expr` happens to equal another field's key expression, it gets misclassified.

For instance, if we have a dataset `orders` with `primary_key=["id"]` and two fields sharing `expr="id"` (something like `customer_id` and `foo_id`) both get classified as PRIMARY entities, and the dimension is lost.

On the other hand, a genuine FK field whose `expr` collides with another field's PK `expr` gets classified PRIMARY instead of FOREIGN, losing the relationship's cardinality on round-trip.

Since `expr` isn't guaranteed unique across fields the way `name` is, this trades one false-negative bug for a false-positive one 😄

jbonofre (changes requested):

PR #318, opened 2026-08-11
fix(dbt): warn on non-portable metric semantics

### LINKED ISSUE #317: dbt converter loses derived metric time-context semantics in MetricFlow to Ossie conversion
## Problem

The MetricFlow → Ossie converter can silently lose time-context semantics from derived metrics.

For example, a month-over-month metric can use an offset input:

```yaml
- name: month_over_month_growth
  type: derived
  type_params:
    expr: (cur - pre) / pre
    metrics:
      - name: revenue
        alias: cur
      - name: revenue
        alias: pre
        offset_window:
          count: 1
          granularity: month
```

The current converter flattens this into an ordinary Ossie SQL expression:

```text
(SUM(orders.amount) - SUM(orders.amount)) / SUM(orders.amount)
```

The `pre` offset semantics are lost. A top-level cumulative metric emits `CUMULATIVE_SEMANTICS_LOSS`, but a derived metric using offset-period inputs does not emit an equivalent warning.

## Expected behavior

The converter should report a `ConverterIssue` for derived metrics using non-portable time-context semantics, including `offset_window`, `offset_to_grain`, rolling windows, and grain-to-date semantics. The warning should identify the metric whose semantics were lost.

Strict mode could fail conversion instead of emitting a potentially misleading ordinary SQL expression.

fanruan-tuco:
## Summary

MetricFlow time-context semantics such as cumulative metrics, windows, time spine behavior, and offset-period inputs cannot be represented faithfully in Ossie. This change emits `NON_PORTABLE_METRIC_SEMANTICS` instead of silently producing a lossy conversion.

It also checks derived and ratio dependencies recursively, including cases such as:

```yaml
- name: month_over_month_growth
  type: derived
  type_params:
    expr: (cur - pre) / pre
    metrics:
      - name: revenue
        alias: cur
      - name: revenue
        alias: pre
        offset_window:
          count: 1
          granularity: month
```

Fixes #317

## Validation

`uv run pytest -q` in `converters/dbt`: 103 passed.


jbonofre:
@fanruan-tuco can you please rebase and resolve conflicts? Thanks!

converters/dbt/src/ossie_dbt/msi_to_ossie.py
```diff
@@ -110,6 +145,13 @@ def convert(
                 issues.append(
                     ConverterIssue(issue_type=ConverterIssueType.CUMULATIVE_SEMANTICS_LOSS, element_name=metric.name)
                 )
+            elif _has_nonportable_semantics(metric, metric_index):
```
jbonofre, line 148: This new branch records a `NON_PORTABLE_METRIC_SEMANTICS` issue but doesn't change control flow: the next line (`expr = self._resolve_metric_expression(...)`, unchanged) still runs and emits the lossy expression regardless.

I verified against the example in the PR (`month_over_month_growth = (cur - pre) / pre`): the emitted expression is `(SUM(orders.amount) - SUM(orders.amount)) / SUM(orders.amount)`, always 0. The output document carries no marker of the problem, only a side-channel entry in `result.issues`: a caller that doesn't check `issues` ships an always-zero metric silently.

Should this branch `continue`/skip emitting the metric (like `CONVERSION_METRIC_DROPPED` above) instead of failing through?

converters/dbt/src/ossie_dbt/msi_to_ossie.py
```diff
@@ -65,6 +65,41 @@ class _EntityEntry:
     entity_type: EntityType
 
 
+def _has_nonportable_semantics(
+    metric: Metric,
+    metric_index: Dict[str, Metric],
+    visiting: Optional[set[str]] = None,
+) -> bool:
+    """Return whether a metric uses semantics Ossie cannot represent."""
+    if metric.type is MetricType.CUMULATIVE:
```
jbonofre, line 74: This only special-cases `CUMULATIVE`. A `DERIVED`/`RATIO` metric that depends on a `CONVERSION`-type metric return `False` here (not portable-safe, just unrecognized), so `convert()` proceeds into `_unresolve_metric_expression`, which raises an unhandled `RuntimeError` ("Unexpected CONVERSION metric in expression resolver") and aborts conversion of the entire manifest.

Maybe worth adding a `MetricType.CONVERSION` case here too.

jbonofre (changes requested):

PR #250, opened 2026-07-23
feat: Add DAX, and examples to the Dialects

MikeCarlo:
### Specification
- [x] Spec changes are included in `core-spec/` and follow the existing structure
- [x] Spec changes have been discussed on the mailing list or in a linked issue
- [x] Breaking changes to the spec are clearly called out in the summary

### Ontology
- [x] Ontology changes in `ontology/` are consistent with spec changes
- [x] New or modified terms are defined and documented

### Converters
- [x] Converter logic in `converters/` is updated to reflect spec or ontology changes
- [x] New converters include tests under the converter's test directory

### Validation
- [x] Validation rules in `validation/` are updated if the spec changed
- [x] New validation cases are covered by tests

### Documentation
- [x] `docs/` is updated to reflect any user-facing changes
- [x] New features or behaviors are documented with examples where appropriate
- [x] `CONTRIBUTING.md` is updated if the contribution process changed

### Examples
- [x] `examples/` are added or updated for any new spec constructs or converter support

### Tests
- [x] All existing tests pass (`pytest` / CI green)
- [x] New functionality is covered by tests

### Compliance
- [x] ASF license headers are present on all new source files
- [x] No third-party dependencies are added without PMC/IPMC approval

PR #329, opened 2026-08-16
Add PowerBI <> Ossie Converters

m-kovalsky:
## Summary

This PR contains converters to and from Microsoft Power BI Semantic Models and Ossie models.

### Specification
- This, in combination with #328 enables developers to convert to/from Microsoft Power BI Semantic Model and Ossie.

### Validation
- [ ] Tests and sample models are are contained in this PR

### Examples

### Tests
- [ ] Tests and sample models are are contained in this PR


core-spec/spec.md
```diff
@@ -446,6 +447,7 @@ The following are well-known examples:
 | `GOODDATA` | GoodData-specific attributes |
 | `HONEYDEW` | Honeydew-specific attributes |
 | `WISDOM` | WisdomAI-specific attributes |
+| `POWER_BI` | Power BI / Fabric semantic model-specific attributes |
```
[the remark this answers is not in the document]
m-kovalsky: This will be addressed in #250

converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
```diff
@@ -0,0 +1,588 @@
...
+# `Schema="dbo", Item="Sales"` navigation in a Power Query (M) partition.
+_M_ITEM_RE = re.compile(r'Item\s*=\s*"([^"]+)"')
+_M_SCHEMA_RE = re.compile(r'Schema\s*=\s*"([^"]+)"')
+_M_DATABASE_RE = re.compile(r'Sql\.Databases?\s*\(\s*"[^"]*"\s*,\s*"([^"]+)"')
+
+# TMSL properties consumed by the Apache Ossie mapping at each level. Everything else is
+# preserved verbatim in the stash. This is deliberately a deny-list rather than an
+# allow-list: TMSL grows new properties over time, and an allow-list would silently drop
+# any it had not been taught about, which is exactly what the losslessness rule forbids.
+_MODEL_CONSUMED = frozenset({"description", "tables", "relationships"})
```
jbonofre, line 76: TMSL `annotations` are not in `_TABLE_CONSUMED`/`_MODEL_CONSUMED`, so they're never mapped back to `ai_context` on import.
If a dataset's `ai_context` was exported to an `OssieAIContext` annotation, re-importing that TMSL sweeps the annotation into `custom_extensions` instead of restoring `ai_context`.

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
```diff
@@ -0,0 +1,793 @@
...
+        semantic_model.get("relationships") or [], table_columns
+    )
+    relationships.extend(stash.get("excludedRelationships") or [])
+
+    # Tables the import excluded (private, calculation group, auto date table) are
+    # restored verbatim so a round trip reproduces the original model.
+    tables.extend(stash.get("excludedTables") or [])
+
+    model = {"tables": tables}
+    if generated_partitions:
```
jbonofre, line 184: `model["expressions"]` is assigned unconditionnally here whenever any table needs a generated Direct Lake partition, but this runs before the stash-replay loop `model.setDefault(...)` below, so a preserved model-level expression from a prior import can never win.

If the document mixes a stashed table with one brand-new dataset needing a generated partition, this silently overwrites the shared `DatabaseQuery` M expression for every table, not just the new one.

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
```diff
@@ -0,0 +1,793 @@
...
+    for key, value in stash.items():
+        if key not in _STASH_CONTROL_KEYS:
+            model.setdefault(key, value)
+    model.setdefault("culture", "en-US")
+    _apply_ai_context(model, semantic_model.get("ai_context"))
+
+    bim = {"name": semantic_model.get("name") or "semantic_model"}
+    if description and stash.get("descriptionSource") == "document":
+        bim["description"] = description
+    bim.update(document_properties)
```
jbonofre, line 204: `bim.update(document_properties)` can restore a stashed `compatibilityLevel` before this `setdefault` runs, so `setdefault` silently no-ops even when the current export needs a different level.

Should we do this check before honoring the stashed value, or override it when `generated_partitions` is true?

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
```diff
@@ -0,0 +1,793 @@
...
+            continue
+
+        from_table = relationship.get("from")
+        to_table = relationship.get("to")
+        from_column, to_column = from_columns[0], to_columns[0]
+        if not _endpoints_exist(scope, table_columns, from_table, from_column,
+                                to_table, to_column):
+            continue
+
+        if stash.get("flipped"):
```
jbonofre, line 710: This unconditionally swaps `from`/`to` back to the original orientation whenever `stash.get("flipped")` is true, with no staleness check.

If a user hand-edits the YAML to intentionally reverse a relationship, this silently reverts their edit back to the pre-edit orientation on export.

converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
```diff
@@ -0,0 +1,588 @@
...
+def _convert_metrics(tables):
+    metrics = []
+    seen = set()
+    for table in tables:
+        for measure in table.get("measures") or []:
+            if not isinstance(measure, dict) or not measure.get("name"):
+                continue
+            scope = f"table '{table['name']}' measure '{measure.get('name')}'"
+            warn_unsupported(scope, measure, TMSL_UNSUPPORTED_MEASURE, "Apache Ossie", _PRESERVED)
+            expression = text(measure.get("expression", "")).strip()
```
jbonofre, line 399: Every other exclusion case in this module (private tables, calculation groups, calculated tables, auto-date tables, unresolvable relationships) gets stashed for round-trip preservation, but a blank-expression measure is just warned about and dropped here with no equivalent `excludedMeasures`-style stash.

That makes it permanently unrecoverable rather than round-trippable.

Is it intentional, or should we preserve it like the other exclusions?

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
```diff
@@ -0,0 +1,793 @@
...
+    seen = {}
+    table = dataset["name"]
+    for field in dataset.get("fields") or []:
+        if not isinstance(field, dict) or not field.get("name"):
+            continue
+        aliases = {field["name"]}
+        expressions = dialect_expressions(field.get("expression"))
+        if expressions:
+            _, expression = _preferred_expression(expressions)
+            candidate = expression.strip('"').strip("`").strip("[]")
```
jbonofre, line 514: `_dataset_column_index` only registers unqualified alias keys, but `_column_index` (used for metrics) also registers a `table.column`-qualified key for every column. Since both feed the same `_sql_to_dax` resolver, a calculated column that self-qualifies its own column (let's say `customer.first_name` where the dataset is named `customer`) fails to resolve here even though the identical style works for metrics.

converters/microsoft/src/ossie_microsoft/engine.py
```diff
@@ -0,0 +1,449 @@
...
+def _request(method, url, token, payload=None):
+    data = json.dumps(payload).encode("utf-8") if payload is not None else None
+    request = urllib.request.Request(url, data=data, method=method)  # noqa: S310
+    request.add_header("Authorization", f"Bearer {token}")
+    request.add_header("Content-Type", "application/json")
+    try:
+        with urllib.request.urlopen(request) as response:  # noqa: S310
+            body = response.read().decode("utf-8")
+            return response.status, (json.loads(body) if body else None), dict(response.headers)
+    except urllib.error.HTTPError as exc:
```
jbonofre, line 140: `_request` only catches `HTTPError`, not `URLError` (DNS failure, reset, timeout). Since `deploy()` is called outside the `try`/`finally` that guards the workspace-item `DELETE` cleanup, a transient `URLError` during operation polling propagates straight out of `validate_with_engine`, leaving an already-created Fabric workspace item undeleted even with `keep=False`.

converters/microsoft/src/ossie_microsoft/engine.py
```diff
@@ -0,0 +1,449 @@
...
+        "POST", f"{FABRIC_API}/workspaces/{workspace}/items", token, payload
+    )
+    if status in (200, 201):
+        return body["id"], None
+    if status == 202:
+        operation = headers.get("Location")
+        result = _wait_for_operation(operation, token)
+        if result.get("status") != "Succeeded":
+            return None, json.dumps(result)[:4000]
+        _status, created, _headers = _request("GET", f"{operation}/result", token)
```
jbonofre, line 330: `created["id"]` assumes the follow-up GET succeeded, but `_request` failure branch returns a string, not a dict.

If that GET fails (expired token, transient 5xx) after the create operation itself succeeded, this raises `TypeError: string indices must be integers` instead of `deploy()` documented `(None, "HTTP ...")` return. It crashes `validate_with_engine` rather than surfacing a clean error.

jbonofre (changes requested):

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
eisber: Addressed in `f31b9fd`. Compatibility is now treated as a feature minimum: exports with newly generated Direct Lake partitions use `max(stashed, 1702)`, higher valid levels are retained, malformed levels fail explicitly, and import/generated-M partitions do not force a Direct Lake upgrade. Low/equal/high/no-Direct-Lake cases are covered. The exported round trip also passed TOM structural validation and a real Fabric deploy/refresh/evaluate/delete cycle.

converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
eisber: Addressed in `f015ef2`. Import now extracts only the `OssieAIContext` annotation into `ai_context` at model, dataset, field, metric, and relationship scope; structured JSON is decoded while scalar text remains text, and every unrelated annotation stays in the Power BI stash for replay. Regression coverage exercises all five scopes, string/object forms, and preservation through a full round trip. Validated in the integrated stack: 383 tests, 95.13% branch coverage, Ruff clean, TOM round-trip clean, and live Power BI engine validation clean.

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
eisber: Addressed in `3bf699c`. Import records the normalized endpoint snapshot with orientation/cardinality metadata. Export restores the original Power BI orientation only while the current Ossie endpoints still match that snapshot; edited endpoints win, and stale cardinalities are not replayed. Legacy stashes retain a conservative generated-name freshness check. Tests cover exact unchanged round trip, reversed endpoints, edited columns, stale cardinalities, and legacy payloads.

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
eisber: Addressed in `58ceafa`. `_dataset_column_index` now registers both unqualified and case-insensitive `dataset.alias` keys for field names and physical/source aliases while retaining the existing uniqueness filter. Tests cover qualified model names, qualified source aliases, case variation, and ambiguous/foreign/over-qualified references falling back to the explicit `BLANK()` refusal path.

converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
eisber: Addressed in `a9c45ed`. Expressionless measures are now preserved verbatim in version-2 `excludedMeasures` model stash entries with home table and original position, then restored on export. Current authored metrics are applied first and win on table/name collisions; missing home tables and malformed preserved entries are reported rather than silently guessed. Tests cover exact payload/order round trip, multiple tables, missing homes, malformed entries, and authored-metric replacement.

converters/microsoft/src/ossie_microsoft/engine.py
eisber: Addressed in `df70115`, reconciled with transport handling in `5df6212` and covered further in `8813262`. The async result GET now validates status, body type, and a non-empty string ID before indexing; retries only transient HTTP/transport results; fails immediately for auth errors; and returns `(None, diagnostic)` for malformed successful responses instead of raising `TypeError`. Tests cover 500/string, 401/envelope, missing/empty IDs, transient HTTP recovery, and transient transport recovery. Full integrated suite: 383 passed at 95.13% coverage.

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
eisber: Addressed in `2bf9bf4`. Generated Direct Lake partitions are tracked individually; preserved model expressions are merged rather than replaced, compatible `DatabaseQuery` expressions are reused, and an explicit conflicting source gets a deterministic collision-free name wired only to newly generated partitions. Tests cover mixed preserved/new tables, unrelated expressions, compatible reuse, name collisions, and unchanged old partition references. Integrated validation: 383 tests, 95.13% coverage, TOM and live Power BI round trips clean.

converters/microsoft/src/ossie_microsoft/engine.py
eisber: Addressed in `2acea83`, with integration diagnostics in `5df6212` and coverage in `8813262`. `_request` normalizes URL/DNS/connection/timeout failures; operation and refresh polling retry bounded transient failures; orchestration enters `finally` with the dataset ID whenever one is available; cleanup failures are surfaced as failures; and an unknown ID explicitly reports that cleanup could not be attempted. Regression tests cover transient recovery, persistent transport failure, known-ID cleanup, cleanup failure, and `keep=True`. A real Fabric deploy/refresh/evaluate completed successfully and both validation items were confirmed deleted afterward.

jbonofre (dismissed):
It looks good to me now.

I re-triggered the failed job on CI.

We should fix `converters/microsoft/uv.lock`: we should run `uv lock` and commit the result, and add `--locked` to the `uv sync` step in the CI workflow so a drifted lock fails loudly instead of silently re-resolving.

jbonofre:
I took the liberty to push the fix about `uv` command directly on the PR.

jbonofre:
I'm now investigating the CI failure.

jbonofre:
I updated the PR in order to fix the CI.

jbonofre (approved):

Issue #301, opened 2026-08-04
`_build_relationships` emits relationships whose to_columns is not a key, producing OSI that downstream consumers reject


georgea-tk:
`MSIToOSIConverter._build_relationships` pairs *every* combination of datasets sharing an entity name, including pairs
where **both** sides declare that entity as `FOREIGN`. The resulting relationship has a `to_columns` that is not a
primary or unique key of the `to` dataset, which contradicts the spec's own definition of that field and is rejected by
downstream consumers.

`core-spec/osi-schema.json` defines:

```json
"to_columns": {
  "description": "Primary/unique key columns in the 'to' dataset"
}
```

**Root cause**

`converters/dbt/src/ossie_dbt/msi_to_osi.py:395-421`, `_build_relationships`:

```python
for entity_name, entries in entity_index.items():
    for entry_a, entry_b in combinations(entries, 2):
```

Every pair is emitted. Direction is then chosen in `_relationship_direction` (same file, `:368-392`):

```python
if a_is_one_side and not b_is_one_side:   # FOREIGN -> PRIMARY/UNIQUE   ok
    ...
if b_is_one_side and not a_is_one_side:   # PRIMARY/UNIQUE <- FOREIGN   ok
    ...
# Same cardinality tier — use alphabetical order for determinism.
if ds_a <= ds_b:
    ...
```

The final branch handles two cases that are not equivalent:

- **PRIMARY↔PRIMARY** (or UNIQUE) — `to_columns` *is* a key of the `to` dataset. Valid 1:1 join, and covered by
  `test_same_type_entities_produce_relationship`. This should keep working.
- **FOREIGN↔FOREIGN** — `to_columns` is *not* a key of the `to` dataset. This is the invalid case, and it has no test.

In a star schema every fact table declares the same conformed-dimension entity, so this fires between every pair of
fact tables. For *N* fact models sharing one entity, `combinations` yields N(N+1)/2 relationships of which only N are
valid — 465 vs 30 at N=30.

**Repro** (HEAD `88e0011148283302c9a04cd0287e00e0b9d87354`), using the repo's own test helpers:

```python
import sys
sys.path.insert(0, "src"); sys.path.insert(0, "."); sys.path.insert(0, "../../python/src")

from ossie_dbt.msi_to_osi import MSIToOSIConverter
from tests.helpers import _entity, _manifest
from metricflow_semantic_interfaces.test_utils import semantic_model_with_guaranteed_meta
[...]

PR #330, opened 2026-08-16
Warn when relationship to_columns does not cover a declared key

kayemkim:
## Summary

The spec defines `to_columns` as "Primary/unique key columns in the 'to' dataset" (core-spec/spec.md, osi-schema.json), but the validator only checks that the relationship's datasets exist. A relationship that joins to non-key columns silently breaks many-to-one semantics downstream — joins fan out, and consumers that trust the declared cardinality produce wrong results. #301 shows this class of document being produced in practice and rejected by downstream consumers; a validator-side check catches it at the document level regardless of which tool produced it.

`validate_references` now checks that `to_columns` covers the `to` dataset's `primary_key` or one of its `unique_keys`. Two deliberate softenings, both drawn from how the repo already treats this rule:

- Coverage rather than exact equality: a `to_columns` that is a superset of a key still guarantees the join cardinality (tenant-sharded joins are a common shape), and the Databricks converter's `_covers_unique_key` already applies exactly these semantics.
- Warning rather than error: declared keys can be a partial recovery of the dataset's real keys (the Databricks importer, for instance, records `unique_keys` from a single join's `rely` hint), so a non-covering `to_columns` is suspicious but not provably wrong. The message goes through the validator's existing warning channel and does not fail validation. Happy to tighten this to an error if you'd rather trust key declarations as exhaustive.

Datasets that declare no keys are skipped — both `primary_key` and `unique_keys` are optional, so there is nothing to check against. The check also guards against shapes that already fail schema validation (null `unique_keys`, non-list `to_columns`, flat `unique_keys`) so the semantic pass can't crash or mislead on an invalid document.

Ran against every semantic-model YAML in the repo (examples plus converter fixtures): no regressions, all still pass. A relationship pointing at non-key columns now reports:

```
[Reference] Warning: Relationship 'orders_to_customers' in model 'm': to_columns ['region'] does not cover the primary key or a unique key of dataset 'customers'
```

A couple of adjacent gaps I noticed but left out of scope, flagging in case they're worth issues: no CI workflow currently runs `validation/tests` (this suite included), and the orionbelt converter's mirror validator (which documents itself as mirroring `validate.py`) doesn't have this rule — I can follow up on either. #307 (pending) adds from/to column count validation to the same function and introduces the same test file path; the checks are complementary and I'm happy to rebase if it lands first.

## Related Issues

Related to #301 (validator-side guard for the same class of invalid document; does not fix the converter itself).

### Specification

### Ontology

### Converters

### Validation
- [x] Validation rules in `validation/` are updated if the spec changed

### Documentation

### Examples

### Tests

### Compliance

### DIFF
--- a/validation/tests/test_validate.py
+++ b/validation/tests/test_validate.py
@@ -1,0 +1,163 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from importlib.util import module_from_spec, spec_from_file_location
+from pathlib import Path
+
+import pytest
+
+# validate.py exits at import time when its dependencies are missing, which
+# would abort the whole pytest session during collection — skip instead.
+pytest.importorskip("yaml")
+pytest.importorskip("jsonschema")
+
+_VALIDATE_PATH = Path(__file__).parents[1] / "validate.py"
+_SPEC = spec_from_file_location("ossie_validate", _VALIDATE_PATH)
+assert _SPEC is not None and _SPEC.loader is not None
+_VALIDATE = module_from_spec(_SPEC)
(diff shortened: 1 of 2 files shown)


kayemkim:
Two updates since opening this.

#337 touches the same `validate.py` (rejecting duplicate mapping keys) and adds the validation CI workflow whose absence I flagged in the summary. That workflow runs `uv run validation/test_validate.py` directly, while this PR keeps its tests at `validation/tests/test_validate.py`, so they would not be picked up as-is. If #337 lands first I'll rebase and move the tests to match its layout so the new check actually runs in CI. The `validate.py` changes themselves should not conflict: #337 changes the YAML loading path, this PR changes `validate_references`.

Also a correction to the summary above: "all still pass" overstates it. A few repo files fail validation on main as well, for reasons unrelated to this change: `examples/flights.yaml` is an ontology-layer document (it passes against `ontology/ontology.json`), and the orionbelt fixtures declare `version: 0.2.0.dev0`, which the core schema rejects. Those fail identically with and without this change, so the accurate claim is no regressions, not a clean pass.

khush-bhatia (approved):

PR #288, opened 2026-07-29
Rename remaining OSI references to Ossie

jbonofre:
## Summary

Completes the **OSI → Ossie** rename across the repository, covering the parts that still carried the former name. **135 files changed** (57 file/directory renames, the rest content edits).

Rebased onto current `main` (`1d9ebce`).

### Structural renames
- **core-spec:** `osi-schema.json` → `ossie-schema.json` — including its `$id` and every reference (examples, docs, ROADMAP, python tests, and the Java schema-loader / pom resource copy).
- **Python package:** `OSI*` model classes → `Ossie*` (`OssieDocument`, `OssieDataType`, `OssieSemanticModel`, …); `to_osi_json`/`to_osi_yaml` → `to_ossie_*`.
- **Converters (Python):** package/dir renames (`osi_omni` → `ossie_omni`, `honeydew_osi` → `honeydew_ossie`) plus source, test, and fixture files (`osi_to_msi.py`, `test_osi_converter_*.py`, `tpcds_osi.yaml`, …).
- **Converters (Java):** `Osi*` classes → `Ossie*` (files kept in sync with class names); `ALL_CAPS` constants → `OSSIE` (e.g. `SALESFORCE_TO_OSSIE`); CLI direction flag `toOSI`/`"toosi"` → `toOssie`/`"toossie"`.

### Former-name references outside the `osi` token
- **Pre-donation GitHub org:** 3 links to `github.com/open-semantic-interchange/Ossie` → `github.com/apache/ossie`, in `converters/orionbelt/tests/test_ossie_tpcds_baseline.py` and `converters/orionbelt/ossie_obml_mapping_analysis.md`. Two also had a stale path — `converters/index.md` does not exist here, so they now point at `converters/README.md`, which carries the heading the anchor targets.
- **NVIDIA converter:** dropped the `"Open Semantic Interchange"` PyPI keyword, aligning it with the other nine packages (`"Apache Ossie", "Ossie", "<Vendor>"`).

### Integration with newer `main` commits
Three references in code merged since this branch was opened pointed at names this PR renames:
- `converters/nvidia/tests/test_converter.py` — schema path `core-spec/osi-schema.json` → `ossie-schema.json` (from #247/#299).
- `converters/dbt/tests/test_ossie_to_msi.py` — the SUM_BOOLEAN tests (#292) used `_osi_doc`/`_osi_dataset`/`_osi_field`/`_osi_metric`/`OSIToMSIConverter`, all renamed here; left as-is they would raise `NameError`.
[...]

PR #331, opened 2026-08-18
fix(dbt): use pydantic v1 .json() in ossie-to-msi

### LINKED ISSUE #296: ossie-dbt osi-to-msi crashes: PydanticSemanticManifest has no model_dump_json
`ossie-dbt osi-to-msi` fails on every input, including the repo's own example:

```
$ ossie-dbt osi-to-msi -i examples/tpcds_semantic_model.yaml -o out.json
  File "converters/dbt/src/ossie_dbt/cli.py", line 78, in _cmd_osi_to_msi
    output_path.write_text(result.output.model_dump_json(by_alias=True, exclude_none=True, indent=2))
AttributeError: 'PydanticSemanticManifest' object has no attribute 'model_dump_json'
```

`PydanticSemanticManifest` comes from `metricflow_semantic_interfaces` and subclasses `pydantic.v1.BaseModel`, so the serializer is `.json()`, not `.model_dump_json()`. Calling `.json(by_alias=True, exclude_none=True, indent=2)` on the same object produces valid output.

Reproduced on main at 88e0011. The conversion itself works — only the CLI write fails. No test covers the CLI, so CI passes.

ntoxlut:
## Summary

`ossie-dbt ossie-to-msi` fails on every input, including the repository's own
`examples/tpcds_semantic_model.yaml`, with:

```
AttributeError: 'PydanticSemanticManifest' object has no attribute 'model_dump_json'
```

`PydanticSemanticManifest` (from `metricflow_semantic_interfaces`) subclasses
`pydantic.v1.BaseModel`, whose JSON serializer is `.json()`, not the pydantic v2
`.model_dump_json()`. The conversion itself works — only the CLI's write step
was calling a v2-only method on a v1 object.

**Fix**

- `cli.py`: use `.json(by_alias=True, exclude_none=True, indent=2)`, which takes
  the same options on a pydantic v1 model and produces valid output.
- Add `tests/test_cli.py` — an end-to-end test that drives `ossie-to-msi` and
  asserts a valid semantic manifest is written. No test covered the CLI before,
  which is why CI stayed green through the crash.

**Verification**

```
cd converters/dbt
uv run pytest          # 101 passed (was 100)
uv run ossie-dbt ossie-to-msi -i ../../examples/tpcds_semantic_model.yaml -o /tmp/manifest.json
# -> Written to /tmp/manifest.json  (valid JSON)
```

The new test fails without the fix (`AttributeError`) and passes with it.

## Related Issues

Fixes #296

### Specification
_N/A — bug fix only, no spec change._

### Ontology
_N/A — no ontology change._

### Converters

### Validation
_N/A — no validation change._

### Documentation
_N/A — no user-facing behavior change beyond the command no longer crashing._

### Examples
_N/A — no new spec constructs or converter support._

### Tests

### Compliance

### DIFF
--- a/converters/dbt/src/ossie_dbt/cli.py
+++ b/converters/dbt/src/ossie_dbt/cli.py
@@ -75,7 +75,9 @@ def _cmd_osi_to_msi(args: argparse.Namespace) -> None:
     document = OSIDocument.model_validate(raw)
     result = OSIToMSIConverter().convert(document)
 
-    output_path.write_text(result.output.model_dump_json(by_alias=True, exclude_none=True, indent=2))
+    # PydanticSemanticManifest subclasses pydantic.v1.BaseModel, whose JSON
+    # serializer is .json(), not the pydantic v2 .model_dump_json().
+    output_path.write_text(result.output.json(by_alias=True, exclude_none=True, indent=2))
     print(f"Written to {output_path}", file=sys.stderr)
 
 
--- a/converters/dbt/src/ossie_dbt/cli.py
+++ b/converters/dbt/src/ossie_dbt/cli.py
@@ -74,7 +74,9 @@ def _cmd_ossie_to_msi(args: argparse.Namespace) -> None:
     document = OssieDocument.model_validate(raw)
     result = OssieToMSIConverter().convert(document)
 
-    output_path.write_text(result.output.model_dump_json(by_alias=True, exclude_none=True, indent=2))
+    # PydanticSemanticManifest subclasses pydantic.v1.BaseModel, whose JSON
+    # serializer is .json(), not the pydantic v2 .model_dump_json().
+    output_path.write_text(result.output.json(by_alias=True, exclude_none=True, indent=2))
     print(f"Written to {output_path}", file=sys.stderr)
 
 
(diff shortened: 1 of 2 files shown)


MonkeyCanCode:
Thanks for the PR @ntoxlut . The fix looks good to me, however, we need to change the naming in test where `osi` is now renamed to `ossie`. So the current test would fail on the main branch.

ntoxlut:
Thanks @MonkeyCanCode, good catch. Rebased onto current `main` and renamed everything to match #288 — helpers are `_ossie_*` now, `to_ossie_yaml()`, and the subcommand is `ossie-to-msi`.

The fix itself still applies: `cli.py` on `main` calls `.model_dump_json()` on `PydanticSemanticManifest`, which is a pydantic v1 model.

`uv run pytest` passes (101), and reverting just the `cli.py` change makes the new test fail with the original `AttributeError`.

MonkeyCanCode (approved):

MonkeyCanCode:
Thanks for the fix @ntoxlut. LGTM.

QMalcolm (commented):
LGTM! Thanks @ntoxlut for doing this work 🙂

QMalcolm (approved):

Issue #321, opened 2026-08-12
[databricks] MV→Ossie import aborts on non-equi/filtered join conditions instead of preserving them in custom_extensions


jackstein21:
## Summary
`convert_metric_view_to_ossie` raises `ConversionError` and aborts the entire import when a Metric View join's `on` clause is not a simple equi-join of `alias.column = alias.column` pairs. Real-world metric views routinely use function-wrapped keys and filter predicates in join conditions, so this makes the importer unusable on them — even though the converter already has a custom_extensions mechanism that could preserve such joins losslessly.

## Current behavior
`_decompose_on()` returns the raw `on` string for any clause containing a non-equi operator, a SQL fragment (e.g. `UPPER(...)`, `COALESCE(...)`), or an extra filter predicate (`... AND col NOT IN (...)`). `_convert_join()` then:

    if raw_on is not None:
        raise ConversionError("Join '...' uses a non-equi or unsupported join
        condition ... Cannot import.")

Example (generic Databricks Unity Catalog metric view):
```
    joins:
      - name: accounts
        source: catalog.schema.account
        on: UPPER(source.EXTERNAL_ID) = UPPER(COALESCE(accounts.ID_1,
            accounts.ID_2))
            AND accounts.ACCOUNT_ID NOT IN ('TEST_1', 'TEST_2')
```
→ import aborts; nothing is emitted.

## Why this is inconsistent
The importer already preserves MV-only join attributes (cardinality, rely, …) via `write_stash(rel, stash)` into the relationship's `custom_extensions`, and preserves other MV-only features (filter, window, format) the same way. The spec supports `custom_extensions` on relationships. Aborting on an unrepresentable `on` — rather than stashing it like everything else — is the odd one out.

## Proposal
When `on` cannot be decomposed into equi-join column pairs, preserve the raw clause in `relationship.custom_extensions[DATABRICKS]` (e.g. `{"on": "<raw>"}`) and emit a warning, instead of raising. To avoid misrepresenting a filtered/ non-equi join as a clean equi-join, do NOT populate `from_columns`/`to_columns` in that case (or omit the core relationship and stash the join at dataset/model level). Gate it behind an opt-in (e.g. `on_unsupported="preserve"` / a `--lenient` CLI flag) so the default lossless-or-fail contract is unchanged.

## Impact
Enables `MV → Ossie → MV` round-tripping of production metric views whose joins carry business logic, which the current importer rejects outright.

## Related
[...]

PR #333, opened 2026-08-19
[OSSIE][DATABRICKS] Add Java Converter and Reformat

Haoranli503:
## Summary

Adds a **Java implementation** of the Databricks converter alongside the existing Python
reference, and restructures `converters/databricks/` to host both languages. The Java library is
the implementation maintained going forward; the Python converter becomes a reference copy slated
for deprecation.

Conversion is unchanged in spirit: pure YAML text in, YAML text out, reading and writing the two
formats as parsed maps and lists. The Java and Python implementations target the same Databricks
Unity Catalog Metric View YAML v1.1 contract, and the checked-in fixtures pin the expected output
for both directions (`export`: Ossie -> Metric View, `import`: Metric View -> Ossie).

Changes:

- **Restructure** — the existing flat Python converter (`src/`, `tests/`, `pyproject.toml`,
  `README.md`, `uv.lock`) moves under `converters/databricks/python/` with no content changes,
  making room for the Java tree.
- **Add `java/`** — the bidirectional converter as a Maven module: library, an
  `OssieDatabricksConverter` command-line tool, JUnit 5 tests, and YAML fixtures. Package
  `org.apache.ossie.converter.databricks`, following the conventions of the existing `polaris` and
  `salesforce` Java converters (parent `org.apache:apache:39`; deps SnakeYAML / Jackson / JUnit 5;
  `apache-rat-plugin` license audit; shade plugin for the executable CLI jar).
- **Add a root `README.md`** describing the two-language layout and pointing at each
  implementation's own README.
- **CI** (`.github/workflows/converter-databricks-ci.yml`) — point the Python job at the new
  `python/` working directory, and add a Java job (`mvn -B verify`, Temurin JDK 21) mirroring
  `converter-polaris-ci.yml`.

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance


converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieToMetricView.java
```diff
@@ -0,0 +1,1167 @@
...
+      @SuppressWarnings("unchecked")
+      List<String> joinPath = (List<String>) entry[1];
+      boolean isFact = node == root;
+      String prefix = counts.get(node.dataset) > 1 ? node.alias : null;
+      // Fixed per node, like `prefix`: hoisted so it is not rebuilt for every field.
+      String qualifier = String.join(".", joinPath);
+      if (!isFact) {
+        // A dataset reachable by more than one path (diamond) appears once per path; the first
+        // wins, matching the order dimensions are emitted in.
+        datasetAliasPath.putIfAbsent(node.dataset, qualifier);
```
jbonofre, line 239: A dataset reached via two distinct join paths (a diamond) is silently mapped to only its first-seen alias path via `putIfAbsent`. A measure expression qualified by this dataset's bare name (`SUM(shared.amount)`) will always resolve to one arbitrary branch with no warning (unlike the dimension-path handling of the same ambiguity, which explicitily warns and drops).

For instance, fact `orders` -> `dim1` -> `shared` and `orders` -> `dim2` -> `shared`, plus `SUM(shared.amount)`, silently emits `SUM(dim1.dim1_shared.amount)` and drops the `dim2` branch.

Can we warn-and-drop here the same way the dimension path does?

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
```diff
@@ -0,0 +1,581 @@
...
+      modelStash.put(STASH_SOURCE_KEY, factName);
+    }
+    if (!complexJoins.isEmpty()) {
+      modelStash.put("complex_joins", complexJoins);
+    }
+    writeStash(model, modelStash);
+    return model;
+  }
+
+  private static void walk(String parentName, String parentAlias, List<Object> joins,
```
jbonofre, line 258: The import direction (`walk`/`hasOtm`) never enforces `MAX_JOIN_NODES`, while export checks it before and during recursion. A Metric View with 250+ joins imports cleanly but then fails to re-export.

I believe it breaks the documented guarantee that anything that imports cleanly can be exported again.

Should we add the check on import too?

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieToMetricView.java
```diff
@@ -0,0 +1,1167 @@
...
+  private static void markOtm(String modelName, Node root) {
+    // Each top-level join starts a branch and sets that branch's expected cardinality.
+    for (Node top : root.children) {
+      boolean branchIsOtm = !top.parentIsFrom;
+      top.isOtm = branchIsOtm;
+      markOtmVisit(modelName, top, branchIsOtm);
+    }
+  }
+
+  private static void markOtmVisit(String modelName, Node node, boolean branchIsOtm) {
```
jbonofre, line 559: `markOtmVisit` rejects a join whose cardinality differs from its top-level branch's cardinality in either direction.

The Python reference (`_mark_otm`) only rejects reverting from one-to-many back to many-to-one: it explicitly allows a one-to-many join nested under a many-to-one branch. As written, a normal star-schema-with-fanout model (many-to-one to `customer`, then one-to-many to `customer_notes`) that Python accepts gets rejected here with "is one_to_many but descends from a many_to_one join".

Is the stricter check intentional, or should this match the Python behavior?

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieConverterCommon.java
```diff
@@ -0,0 +1,673 @@
...
+      index++;
+    }
+    return extensions;
+  }
+
+  static Map<String, Object> readStash(Map<String, Object> obj) {
+    Map<String, Object> stash = null;
+    for (Map<String, Object> ext : customExtensions(obj)) {
+      if (VENDOR.equals(str(get(ext, "vendor_name")))) {
+        if (stash != null) {
```
jbonofre, line 556: `readStash` throws when more than one DATABRICKS `custom_extensions` entry is present.

Python's `read_stash` just returns the first match and ignores later duplicates.

An Ossie YAML with two such entries that Python converts fine gets rejected here.

Worth aligning with the Python behavior (first-wins) rather than erroring?

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
```diff
@@ -0,0 +1,581 @@
...
+    List<String> parentCols = (List<String>) decomposed[0];
+    List<String> childCols = (List<String>) decomposed[1];
+    String rawOn = (String) decomposed[2];
+    if (rawOn != null) {
+      // The `on` is not an equi-join of simple `alias.column` pairs (it has a non-equi operator, a
+      // SQL-function-wrapped key, or an extra filter predicate), so an Apache Ossie relationship
+      // cannot represent it (from/to columns are required). Rather than abort the whole
+      // conversion, warn here; the columns-less entry built below is stashed under the model's
+      // custom_extensions (complex_joins) so it round-trips, instead of an invalid relationship.
+      notices.warn("join '" + child + "'", "non-equi or unsupported join condition ('on: " + rawOn
```
jbonofre, line 338: This accepts a non-equi/filtered join `on` clause and preserves it via the new `complex_joins` stash, but the Python reference still raises `ConversionError` on the same input.

Is this divergence intentional/tracked, or should Java match Python's current (stricter) behavior?

jbonofre (changes requested):

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieToMetricView.java
Haoranli503: Good catch. Fixed: a measure that names a diamond dataset by its bare name is now dropped with a notice, the same way the dimension path drops a complex expression it cannot unambiguously qualify. A measure addresses a dataset by name, so it cannot be split per branch the way a dimension column is, which makes warn and drop the consistent choice. Added a test measureNamingADiamondDatasetIsDroppedWithNotice covering the orders to dim1/dim2 to shared example.

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
Haoranli503: Good catch, added the bound. MetricViewToOssie (MV to Ossie) now rejects a Metric View whose fact plus joins exceed MAX_JOIN_NODES, the same limit OssieToMetricView already enforces, so the two directions stay symmetric and the round trip holds. One note on the framing: the failure is actually the reverse. A Metric View with 200 or more joins exports cleanly today, but the emitted model then fails to import again, since the MAX_JOIN_NODES checks live on the Ossie to MV side. The fix rejects it at export instead. Added tests exportRejectsMoreJoinsThanImportCanRebuild and exportAtTheJoinLimitStillSucceeds.

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieToMetricView.java
Haoranli503: This one is intentional. The Java matches the Databricks engine's actual rule. A Metric View validates each top level join branch with Join.validateSubJoinCardinalities, seeded from that branch's top level join cardinality, and fails any descendant whose cardinality differs in either direction (Join.scala, invoked per top level join by SingleSourceMetricView.validateJoins). So your example, many_to_one to customer then one_to_many to customer_notes, is actually rejected by the engine at CREATE: Expected 'customer_notes' to have 'many_to_one' cardinality, but found 'one_to_many'. The Python _mark_otm is the one out of step: it only rejects reverting from one_to_many back to many_to_one, so it would emit that view and hit the engine error at CREATE instead. Its leniency does not actually help, it just moves the same rejection from conversion time to CREATE time. The Python copy is deprecated so we left it as is.

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieConverterCommon.java
Haoranli503: Good point, changed it. readStash no longer throws on a duplicate DATABRICKS custom_extensions entry. It keeps the first, ignores the rest to match first wins, and emits a converter notice so the dropped entry is not lost silently rather than vanishing the way a plain first wins would. So a model with two entries converts instead of failing, and the user still sees that one was ignored. Added a test duplicateDatabricksStashKeepsFirstAndWarns.

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
Haoranli503: Intentional and tracked. A join condition Apache Ossie cannot represent as a relationship (a non equi comparison, a function wrapped key, or an extra filter predicate) has no schema valid relationship form, since an Ossie relationship requires from_columns and to_columns. Rather than abort the whole export, the Java preserves the raw clause in a model level complex_joins stash inside custom_extensions, and the reverse converter rebuilds it. That keeps the round trip lossless for a join that carries real business logic, and the emitted model stays schema valid with no misleading columns less stub relationship, so a non Databricks consumer simply does not see the stashed join rather than seeing a wrong one. The Python reference still raises because it is the deprecated copy.

Haoranli503:
Pushed three commits.

a991e68 (fixes #422): cascade drop matched references to dropped fields case sensitively, so a metric like COUNT(DISTINCT REGION_NAME) survived referencing a dropped region_name as a dangling reference. It now matches case insensitively, in both the containment prefilter and the regex, across the propagation gate and the confirmation. Added a test with the reported repro.

53c59f6: merged main into the branch to bring it current with #396. Resolved the one converters/databricks/README.md conflict by keeping this branch's overview, since the Java and Python implementations live in their own subdirectories here.

416119f: migrated the Java converter to flat documents (the model at the document root) to match #396. Import reads the model at the root and rejects a legacy semantic_model wrapper, a missing string name, and root dialects/vendors; export emits the model at the root. Flattened all test inputs and fixtures and updated the export assertions to read the model at the root.

One note on verification for that last commit: I could not run mvn locally, so I compiled the converter and ran a driver covering flat import, the legacy rejection, flat export, and a round trip, and reconverted all three fixtures. The JUnit assertions still need a mvn test run as the final gate.

converters/databricks/java/src/test/java/org/apache/ossie/converter/databricks/OssieConverterSuite.java
```diff
@@ -0,0 +1,1639 @@
...
+        + "- name: d\n"
+        + "  source: c.s.d\n"
+        + "  fields:\n"
+        + "    - {name: d1, expression: {dialects: "
+        + "[{dialect: DATABRICKS, expression: bad_dim}]}}\n"
+        + "    - {name: d0, expression: {dialects: "
+        + "[{dialect: DATABRICKS, expression: 'measure(m1)'}]}}\n"
+        + "    - {name: bad_dim, expression: {dialects: "
+        + "[{dialect: SNOWFLAKE, expression: bad_dim}]}}\n"
+        + "  metrics:\n"
```
jbonofre, line 1558: This YAML literal indents `metrics:` at 2 spaces, which nests it under the dataset item instead of putting it at the model root.

`schemaMapList(model, "metrics", ...)` returns empty, so `m1`/`m0`/`bad_measure`/`keep` never get parsed. 

Running `mvn test` on this PR fails as a result (1 cascade notice observed vs 4 expected).

`metrics:` needs re-indented to column 0.

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
```diff
@@ -0,0 +1,588 @@
...
+    if (!hasUsing && !hasOn) {
+      throw new ConversionException("Join '" + child + "' has no join condition (empty or "
+          + "absent 'on'/'using'); condition-less (cross) joins have no Apache Ossie "
+          + "relationship representation.");
+    }
+    Object[] decomposed = decomposeOn(join, parentAlias, parentName, child);
+    List<String> parentCols = (List<String>) decomposed[0];
+    List<String> childCols = (List<String>) decomposed[1];
+    String rawOn = (String) decomposed[2];
+    if (rawOn != null) {
```
jbonofre, line 330: This unconditionally warns-and-stashes a non-equi/unsupported join condition rather than rejecting it.

I see two issues:
1. it diverges from the Python reference, which raises `ConversionError` for the same input (`test_non_equi_on_rejected`)
2. I believe issue #321, cited as justification, asked for this to be opt-in, not unconditional.

Should this be gated behind a flag, or should it reject like Python does?

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieToMetricView.java
```diff
@@ -0,0 +1,1210 @@
...
+      rel.put("from", get(pj, "from"));
+      rel.put("to", get(pj, "to"));
+      Map<String, Object> stash = new LinkedHashMap<>(pj);
+      stash.remove("name");
+      stash.remove("from");
+      stash.remove("to");
+      writeStash(rel, stash);
+      relationships.add(rel);
+    }
+    String factHint = explicitSource != null ? explicitSource : str(get(modelStash, STASH_SOURCE_KEY));
```
jbonofre, line 193: `explicitSource` is checked with a plain null-check, so an explicit empty string (e.g. `--source "$SRC"` with `$SRC` unset) is treated as a real override instead of "absent".

That throws `ConversionException("requested source '' is not a dataset")` instead of falling back to the model `source_dataset` hint.

Python uses a truthy check here: this should match (`explicitSource != null && !explicitSource.isEmpty()` or similar).

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
```diff
@@ -0,0 +1,588 @@
...
+      Map<String, Object> view, String modelName, Notices notices) {
+    Object source = get(view, "source");
+    if (source == null || source.toString().isEmpty()) {
+      throw new ConversionException("Metric View is missing required 'source'");
+    }
+    // Same subquery test validateSource applies below, so the two cannot drift.
+    boolean isSql = SELECT_WITH_RE.matcher(source.toString().trim()).find();
+    String lastId = lastIdentifier(source);
+    String factName;
+    if (modelName != null) {
```
jbonofre, line 148: Same pattern as `OssieToMetricView`: `modelName` is checked with `if (modelName != null)` instead of a truthy check, so `--name ""` produces `name: ""` in the output instead of falling back to deriving the name from the source table.

Python falls back for the same input: this should too I guess.

jbonofre (changes requested):

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
Haoranli503: hmmm, @christianeu-db can you tal?

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/OssieToMetricView.java
Haoranli503: Good catch, fixed. `explicitSource` now uses a truthy check (`explicitSource != null && !explicitSource.isEmpty()`), matching Python's `explicit_source` or ..., so an empty `--source` falls back to the model hint instead of failing as `requested source '' is not a dataset`. Added a test.

converters/databricks/java/src/main/java/org/apache/ossie/converter/databricks/MetricViewToOssie.java
Haoranli503: Confirmed offline with Chris, went with rejecting, like Python. Fixed!

converters/databricks/java/src/test/java/org/apache/ossie/converter/databricks/OssieConverterSuite.java
Haoranli503: Fixed. The `metrics:` block was nested under the dataset by mistake; moved it to the model root so the cascade test sees all four notices again.

Haoranli503:
Also ported apache/ossie#446 (OSSIE_SQL_2026) to the Java converter: pickExpression now falls back `DATABRICKS`, then `ANSI_SQL`, then `OSSIE_SQL_2026`, with the two drop warnings and four tests updated to match. Merged latest main in, so the python/ copy picks up #446 too and stays in sync.

PR #335, opened 2026-08-19
fix(cli): reject arguments to plugin list

### LINKED ISSUE #334: cli: plugin list silently ignores unexpected positional arguments
### Description

`ossie plugin list` accepts arbitrary positional arguments, silently ignores them, and exits successfully.

### Steps to reproduce

```bash
ossie plugin list unexpected
echo $?
```

### Actual behavior

```text
no plugins installed
0
```

The argument is ignored and the command reports success. Multiple arguments behave the same way.

### Expected behavior

Because `plugin list` does not accept positional arguments, it should reject them with a non-zero exit status and show command usage, consistent with other CLI commands such as `plugin remove`.

### Cause

The `listCmd` definition does not set an `Args` validator. Cobra treats a nil validator as accepting arbitrary arguments.

### Suggested fix

Add `Args: cobra.NoArgs` to `listCmd` and add tests for zero, one, and multiple positional arguments.

### Environment

- Apache Ossie commit: `88e0011`
- Go: `go1.26.2 darwin/arm64`
- OS: macOS arm64

iamrajatrana:
## Summary

Reject unexpected positional arguments passed to `ossie plugin list` instead of silently ignoring them and returning success. Adds table-driven coverage for zero, one, and multiple arguments.

## Before

```console
$ ossie plugin list unexpected
no plugins installed
$ echo $?
0
```

Unexpected arguments were silently ignored.

## After

```console
$ ossie plugin list unexpected
Error: unknown command "unexpected" for "ossie plugin list"
$ echo $?
1
```

Unexpected arguments are rejected with a non-zero exit code.

## Related Issues

Closes #334

## Tests

- `go test ./...`
- `go vet ./...`
- `go build ./...`
- Black-box verification that `ossie plugin list unexpected` exits with status 1

## Checklist

- [x] New behavior is covered by tests
- [x] All existing CLI tests pass
- [x] ASF license header is present on the new test file
- [x] No third-party dependencies were added

### DIFF
--- a/cli/cmd/plugin/list.go
+++ b/cli/cmd/plugin/list.go
@@ -30,6 +30,7 @@ import (
 var listCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List installed plugins",
+	Args:  cobra.NoArgs,
 	// TODO(P1): cross-reference against the embedded plugin registry to show
 	// latest available versions and update indicators once registry embedding
 	// (F4) is implemented.
--- a/cli/cmd/plugin/list_test.go
+++ b/cli/cmd/plugin/list_test.go
@@ -1,0 +1,50 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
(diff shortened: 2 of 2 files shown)


flyrain (approved):
Thanks @iamrajatrana ! cc @RussellSpitzer @jbonofre

jbonofre (approved):
LGTM, thanks!

I just open #335 as follow up.

PR #337, opened 2026-08-19
fix(validation): reject duplicate mapping keys

### LINKED ISSUE #336: validator: duplicate YAML and JSON keys are silently overwritten
### Description

The validator loads input with `yaml.safe_load`, which silently keeps the last value when a YAML or JSON mapping contains duplicate keys. Schema and semantic validation therefore run against an altered document and may report success after user-authored data has been discarded.

### Reproduction

```yaml
version: 0.2.0.dev0
semantic_model:
  - name: sales
    datasets:
      - name: orders
        source: staging.orders
        source: production.orders
```

```bash
uv run validation/validate.py duplicate.yaml
echo $?
```

Actual result:

```text
Validation PASSED: duplicate.yaml
0
```

The validator only sees `production.orders`. The same behavior occurs for duplicate top-level keys, model keys, entire collections, quoted or explicitly tagged equivalent keys, and JSON object keys. A valid second occurrence can also hide an invalid first value.

### Expected behavior

Reject explicit duplicate mapping keys as invalid input, identify the duplicate key and source locations, and exit non-zero. Legal YAML merge-key overrides should remain supported.

### Suggested implementation

Use a custom loader derived from `yaml.SafeLoader` that detects repeated explicit keys before delegating mapping construction to `SafeLoader`. Add regression coverage for nested duplicates, JSON input, aliases, merge overrides, and valid mappings.

iamrajatrana:
## Summary

Reject duplicate keys in YAML and JSON mappings before schema validation can run against silently overwritten data. The loader remains derived from `yaml.SafeLoader`, adds no dependency, preserves aliases and legal merge-key overrides, and reports both mapping and duplicate-key source locations.

This PR also adds Validation CI so the validator tests and canonical example run on Python 3.11 through 3.14 whenever validation, core specification, or canonical example files change.

## Before

```yaml
source: staging.orders
source: production.orders
```

```console
$ uv run validation/validate.py duplicate.yaml
Validation PASSED: duplicate.yaml
$ echo $?
0
```

## After

```console
$ uv run validation/validate.py duplicate.yaml
Error: Invalid YAML: while constructing a mapping
...
found duplicate key 'source'
...
$ echo $?
1
```

## Edge cases covered

- Top-level and nested duplicate keys
- Quoted and explicitly tagged equivalent keys
- Duplicate JSON object keys
- Duplicate scalar keys whose values are collections
- A valid duplicate hiding an earlier invalid value
- Explicit duplicates alongside YAML merge keys
- Repeated merge keys
- Same key in separate mappings remains valid
- YAML aliases remain valid
- Legal merge-key overrides remain valid
- Merge keys remain distinct from a quoted literal `"<<"` key
- Error marks identify the original mapping and duplicate location
- End-to-end non-zero exit for duplicates and successful validation for valid input

### Merge-key behavior

A mapping may contain one YAML merge key and explicitly override values inherited through that merge; this remains supported. Repeating the special `<<` merge key in the same mapping is rejected because YAML provides the sequence form (`<<: [*first, *second]`) for merging multiple mappings and mapping keys are expected to be unique.

## Scope notes

- This change is intentionally limited to `validation/validate.py`. Several converters load their own vendor-specific YAML with `yaml.safe_load`; applying a shared strict-loading policy across those independent packages requires separate compatibility evaluation.
- Empty documents and top-level sequences can still reach the validator's pre-existing `data.get(...)` assumption and raise `AttributeError`. That behavior is unrelated to duplicate-key detection and is not changed here.
- Tests are invoked as `uv run validation/test_validate.py`, matching their PEP 723 dependency metadata and script-local import path.

## Related Issues

Closes #336

## Tests

- `uv run validation/test_validate.py` (16 tests)
- `uv run validation/validate.py examples/tpcds_semantic_model.yaml`
- Duplicate-key integration matrix across YAML and JSON inputs
- `uv run --with ruff ruff check --ignore EXE001 validation/validate.py validation/test_validate.py`
- Validation CI matrix: Python 3.11, 3.12, 3.13, and 3.14

## Checklist

- [x] New behavior is covered by unit and integration tests
- [x] Validation tests are enforced in GitHub Actions
- [x] Existing canonical semantic model validation still passes
- [x] ASF license headers are present on new files
- [x] No third-party dependency was added

### DIFF
--- a/validation/validate.py
+++ b/validation/validate.py
@@ -80,38 +80,61 @@ SKIP_SQL_VALIDATION = {"MDX", "TABLEAU", "MAQL"}
 class UniqueKeyLoader(yaml.SafeLoader):
     """Safe YAML loader that rejects duplicate explicit mapping keys."""
 
-    def construct_mapping(self, node: yaml.MappingNode, deep: bool = False) -> dict:
-        seen = set()
-        merge_tag = "tag:yaml.org,2002:merge"
-        merge_key = object()
+    MERGE_TAG = "tag:yaml.org,2002:merge"
 
-        for key_node, _ in node.value:
-            if key_node.tag == merge_tag:
-                key = merge_key
-                display_key = "<<"
-            else:
-                key = self.construct_object(key_node, deep=deep)
-                display_key = key
+    def construct_document(self, node: yaml.Node):
+        # Validate the composed node graph before SafeConstructor touches it:
+        # flatten_mapping() rewrites mapping nodes in place while expanding "<<",
+        # so a check that runs during construction sees merged, not authored, keys.
+        self._check_unique_keys(node, set())
+        return super().construct_document(node)
 
-            if not isinstance(key, Hashable):
-                raise ConstructorError(
-                    "while constructing a mapping",
-                    node.start_mark,
-                    "found an unhashable key",
-                    key_node.start_mark,
-                )
+    def _check_unique_keys(self, node: yaml.Node, visited: set) -> None:
+        if id(node) in visited:  # alias or recursive anchor: check the node once
+            return
+        visited.add(id(node))
 
-            if key in seen:
-                raise ConstructorError(
-                    "while constructing a mapping",
-                    node.start_mark,
-                    f"found duplicate key {display_key!r}",
-                    key_node.start_mark,
-                )
-            seen.add(key)
+        if isinstance(node, yaml.MappingNode):
+            seen = set()
+            merge_key = object()
+            for key_node, value_node in node.value:
+                if key_node.tag == self.MERGE_TAG:
+                    key, display_key = merge_key, "<<"
+                elif isinstance(key_node, yaml.ScalarNode):
+                    key = display_key = self.construct_object(key_node, deep=True)
+                else:
+                    # Collection keys are unhashable under SafeLoader. Reject them
+                    # here rather than constructing them, which would run
+                    # flatten_mapping() on the graph this walk must not disturb.
+                    raise ConstructorError(
+                        "while constructing a mapping",
+                        node.start_mark,
+                        "found an unhashable key",
+                        key_node.start_mark,
+                    )
 
-        # Delegate construction (including merge-key flattening) to SafeLoader.
-        return super().construct_mapping(node, deep=deep)
+                if not isinstance(key, Hashable):
+                    raise ConstructorError(
+                        "while constructing a mapping",
+                        node.start_mark,
+                        "found an unhashable key",
+                        key_node.start_mark,
+                    )
+
+                if key in seen:
+                    raise ConstructorError(
+                        "while constructing a mapping",
+                        node.start_mark,
+                        f"found duplicate key {display_key!r}",
+                        key_node.start_mark,
+                    )
+                seen.add(key)
+
+                self._check_unique_keys(key_node, visited)
+                self._check_unique_keys(value_node, visited)
+        elif isinstance(node, yaml.SequenceNode):
+            for child in node.value:
+                self._check_unique_keys(child, visited)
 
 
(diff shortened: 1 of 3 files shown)


validation/validate.py
```diff
@@ -75,6 +77,43 @@
...
 
+class UniqueKeyLoader(yaml.SafeLoader):
+    """Safe YAML loader that rejects duplicate explicit mapping keys."""
+
+    def construct_mapping(self, node: yaml.MappingNode, deep: bool = False) -> dict:
+        seen = set()
+        merge_tag = "tag:yaml.org,2002:merge"
+        merge_key = object()
+
+        for key_node, _ in node.value:
```
flyrain: PyYAML's `flatten_mapping` recursively processes a mapping used directly as the value of `<<` without invoking this `construct_mapping` override, so duplicate keys inside that merge source bypass this loop. For example, `<<: {name: orders, name: customers}` is reduced to `name: customers`; in a schema-valid model the validator exits 0 with `Validation PASSED`. Can we recursively check merge-source mapping nodes before they are flattened, and add a regression test for a duplicate inside an embedded merge mapping?

Alternative, we can deliberately reject `<<` and remove the claims that merge keys are supported. But that seems a bigger discussion. AFAIK, Ossie doesn't enforce the YAML spec version.

validation/validate.py
```diff
@@ -75,6 +77,43 @@
...
+                raise ConstructorError(
+                    "while constructing a mapping",
+                    node.start_mark,
+                    f"found duplicate key {display_key!r}",
+                    key_node.start_mark,
+                )
+            seen.add(key)
+
+        # Delegate construction (including merge-key flattening) to SafeLoader.
+        return super().construct_mapping(node, deep=deep)
```
flyrain: This loader can reject a valid merge override when the same mapping is later reused through an alias.

For example:

```yaml
first:
  <<: &defaults
    <<: &base
      source: staging.orders
    source: production.orders

second: *defaults
```

`source: production.orders` is a legal override of the value inherited from `base`. Both `first` and `second` should therefore contain `source: production.orders`, and the standard `SafeLoader` accepts the document.

However, PyYAML modifies the merged mapping while constructing `first`. When `second` later references that mapping, the new duplicate-key check scans the modified representation and reports `source` as a duplicate.

Could we ensure that each original mapping is checked before merge expansion and is not checked again after PyYAML has modified it? A regression test using the example above would help protect this behavior.

flyrain (commented):
Thanks for the fix, @iamrajatrana ! Left some comments. Both comments are related to the merge feature, which is removed in YAML 1.2 spec. I'm OK if we decide to not support older versions of YAML.

validation/validate.py
iamrajatrana: Confirmed and fixed in 97b4a2e. `flatten_mapping` traverses merge-source mappings without routing them through `construct_mapping`, so their duplicates were never checked. Rather than recursing into merge sources from inside the override, the check now runs as a `construct_document` pre-pass over the composed node graph, so a mapping is inspected in its authored form regardless of how it's reached. Regression tests added for inline and anchored merge sources. Details in the summary comment below.

validation/validate.py
iamrajatrana: Confirmed and fixed in 97b4a2e - same root cause as your other comment. Because the pre-pass runs before any flattening, the aliased mapping is never re-inspected in its mutated form, and your example now loads with both `first` and `second` resolving to `production.orders`. Details in the summary comment below.

flyrain (approved):
LGTM. Thanks @iamrajatrana for the fix! 

Can anyone approve the CI? cc @jbonofre @RussellSpitzer

khush-bhatia (approved):

PR #338, opened 2026-08-21
Add structured file-backed dataset sources

isaac-dasari:
## Summary

This draft adds a backward-compatible structured representation for file-backed datasets in `Dataset.source` while preserving all existing string source forms.


core-spec/spec.md
```diff
@@ -124,14 +124,34 @@ Logical datasets represent business entities or concepts (fact and dimension tab
...
+```
+
+File-backed datasets may use an explicit structured form. `kind` is the discriminator; the initial portable file form requires a physical format and at least one location:
+
+```yaml
+source:
+  kind: file
+  format: parquet
+  locations:
+    - s3://analytics-data/orders/*.parquet
```
khush-bhatia: This is a great enhancement to the spec. But how about we generallize this such that it can support any catalog type (Hive, Polaris etc.. ) and then format can be parquet , table or sql query

Examples:
```
  kind: file
  format: parquet
  locations:
    - s3://analytics-data/orders/*.parquet
```

```
  kind: HIVE_CATALOG
  format: table
  identifier: db.schema.table
 ```

```
  kind: SNOWFLAKE_CATALOG
  format: SQL_QUERY
  query: 'select * from db.schema.table where col1 >10'
```

Comment (khush-bhatia) [reply core-spec/spec.md]: cc @flyrain

File: core-spec/spec.yaml [hunk 2]
```diff
@@ -97,9 +97,29 @@ datasets:
...
+    # source:
+    #   kind: file
+    #   format: parquet
+    #   locations:
+    #     - s3://bucket/path/*.parquet
+    #
+    # Structured catalog table form:
+    # source:
+    #   kind: HIVE_CATALOG
+    #   format: table
```
flyrain, line 114: Do we need a format here?  What would be the expected values? table, view, or UDF?  We may only support table and view. We may not need a format in that case. The identifier for them are unified across multiple systems.

core-spec/spec.yaml
```diff
@@ -97,9 +97,29 @@ datasets:
...
+    #   identifier: database.schema.table
+    #
+    # Structured SQL query form:
+    # source:
+    #   kind: SNOWFLAKE_CATALOG
```
flyrain, line 119: Is this a SQL dialect type?

core-spec/spec.md
flyrain: Thanks for tagging me here. That's a good point. I think 3 catalogs(file/catalog entity/sql) makes sense in general. Have we thought about consolidate SQL_QUERY to catalog entity(e.g., view)?

core-spec/spec.yaml
jbonofre: Yes, that's part of the spec about SQL vendor/dialect.

flyrain: Thanks for the context, @jbonofre ! Is it defined anywhere in the spec? I think an explicit specification would be helpful here. WDYT?

PR #343, opened 2026-08-27
feat: support dataset-scoped metrics

jklahr:
## Summary

A metric may now be declared on a dataset (`datasets[].metrics`) as well as on the model (`semantic_model.metrics`). Both placements use the same `Metric` structure; only name resolution differs.

`datasets[].metrics` is optional and absent from every existing model, and `semantic_model.metrics` is unchanged. Nothing breaks.

Discussed on `dev@ossie.apache.org` as `[DISCUSS] Dataset-scoped metrics`. Related: #287, #342.

## What this adds

Today every metric lives at the model level, so an aggregation over a single dataset has to be written as though it could span several, and named uniquely across the entire model.

Declaring it on the dataset gives three things the current spec cannot express:

- **A dataset that carries its own aggregations.** It can be authored, reviewed and reused across models without each model restating its metrics.
- **Scoped metric names.** Two datasets may each declare `revenue`. Today the second one has to be renamed.
- **Ownership without parsing.** A consumer can tell which dataset an aggregation belongs to from where it is declared, instead of inspecting the expression.

## Scoping rules

| | Model-scoped (`semantic_model.metrics`) | Dataset-scoped (`datasets[].metrics`) |
|---|---|---|
| Expression references | Fields of any dataset, qualified: `dataset.field` | Fields of its own dataset, qualified: `dataset.field`; columns of its `source`, unqualified: `column` |
| Name uniqueness | Unique across the model | Unique within its dataset, and distinct from that dataset's field names |
| Referenced as | `metric_name` | `dataset_name.metric_name` |

1. A dataset-scoped metric's expression MAY reference the declared fields of its dataset and the columns of its `source`. A declared field is written `dataset_name.field_name`; a source column is written unqualified. A qualified reference MUST name a declared field of the declaring dataset. An expression that references another dataset MUST be declared in `semantic_model.metrics`.
2. Dataset-scoped metric names MUST be unique within their dataset. Two datasets MAY each declare a metric with the same name.
3. A dataset-scoped metric name MUST NOT collide with a field name of the same dataset.
4. A model-scoped metric SHOULD NOT reuse the name of a dataset-scoped metric in the same model. Warning, not error.
5. An unqualified metric reference resolves to a model-scoped metric. A dataset-scoped metric is referenced as `dataset_name.metric_name`.

**Two spellings, one per namespace.** `SUM(orders.net_amount)` references the declared field, which is how a metric reuses a field's expression instead of repeating it. `SUM(tax)` references a column of the `source` that is not declared as a field. Because the spellings differ, a field and a column may share a name without ambiguity, so declaring a field named after an existing column does not change the meaning of an expression already using the bare name.

**Placement constrains the expression, not the query.** A dataset-scoped metric is joined and grouped like any other, using the model's relationships, so it can be sliced by dimensions of any dataset the model connects.

The spec defines structure and name resolution only. It says nothing about evaluation, grain or traversal, and this proposal does not change that.

## Errors, warnings, and what is left alone

**Hard error**, where one string resolves two ways or a reference cannot resolve at all:

| Condition | Why |
|---|---|
| Duplicate metric name within a dataset | `dataset.name` resolves two ways |
| Metric name equals a field name of the same dataset | Same qualified namespace, so `orders.amount` resolves two ways |
| Expression references another dataset | Belongs in `semantic_model.metrics` |
| Qualified reference does not name a declared field | `orders.tax` asks for a field the model does not declare |

**Warning, exit code 0:** a model-scoped metric reusing a dataset-scoped metric's name. References stay unambiguous, and a dataset may be authored independently and reused across models, so it must not become invalid because of a name the surrounding model introduces. The warning is attributed to the model and names the model-scoped metric as the thing to rename.

**Not checked**, and stated in spec.md so implementations do not add divergent constraints: a model-scoped metric sharing a name with a field or a dataset, since those are spelled differently; whether a bare name is a real column, since that needs catalog metadata the model does not carry; and a qualifier naming neither a dataset nor a field, since it is a local alias, CTE or subquery source.

## Open questions

1. **Is the qualified-reference check right as a hard error?** `SUM(orders.tax)` is valid SQL if `tax` is a column, so erroring enforces Ossie's reading of the qualifier over SQL's.
2. **Is name shadowing right as a warning?** Should it block validation instead?
3. **Should a model-scoped metric sharing a field name warn?** Currently silent, because it would fire on most real models.
4. **Metrics referencing other metrics** is deliberately out of scope, since it affects the expression language rather than placement. Noted so the two stay compatible: under rule 5, a model-scoped metric could reference any metric, and a dataset-scoped metric only metrics of its own dataset.

## Prior art

Most semantic layers already distinguish the two placements. Listed alphabetically.

| System | Single-entity metric | Cross-entity | Name uniqueness | Reference form |
|---|---|---|---|---|
| **AtScale SML** | `metric` bound to one dataset | `metric_calc` | Global | Bare |
| **Cube** | Measures within cubes | Calculated measures | Per cube | `cube.member` |
| **Databricks UC metric views** | Measures in the view | Joins inside the view | Per view | `MEASURE(name)` |
| **dbt MetricFlow** | Metrics in a semantic model | Top-level `metrics` | Global | Bare |
| **Snowflake semantic views** | `tables[].metrics` | Top-level `metrics` | Per table | `table.metric` |

Scoped uniqueness with qualified references, rather than the flat global namespace AtScale SML and dbt MetricFlow use, follows Ossie's existing convention that field names are scoped to a dataset.

Whether some metrics belong at the dataset level rather than above datasets and relationships is an open question in the Ossie Metrics Working Group as well, argued there from the evaluation side: a metric over one dataset has a fixed dimension domain, while one spanning datasets is evaluated on different join paths depending on the filters and dimensions in play. That reasoning lands on the same boundary this proposal draws structurally.

## Not included

**The TPC-DS example is unchanged.** No converter reads `datasets[].metrics` yet, so moving the example's metrics would silently drop three of them on conversion. Shipping a lossy flagship example is worse than shipping the placement without one; the worked examples live in the `Metric Scoping` section of spec.md meanwhile. Proposed as follow-ups: a shared `iter_metrics(model)` helper so converters get both placements right by default, then updating the converters, after which the example can move. `converters/polaris` carries the same gap the Pydantic model did: its `Dataset` class has no `metrics` field.

**No CI workflow covers the paths this PR touches.** Every workflow is filtered to `cli/**` or `converters/<name>/**`, so nothing runs against `core-spec/`, `validation/`, `python/`, `docs/` or `examples/`, and a green check here means no jobs ran. This also affects the validator suite added in #330. Left alone so this PR stays a specification change.

Run locally instead: the validator suite (31 test functions, the 10 from #330 plus 21 added here), the `python` suite, and 9 of the 11 converter suites. Two Java converters need a JRE I do not have. The one `databricks` failure is a pre-existing Hypothesis case reproducible on unmodified `main`. Both examples validate, every fenced YAML block in spec.md parses, and the two snippets spec.md marks INVALID are confirmed rejected by the validator.

## Changes

- **`core-spec/spec.md`**: `Metric Scoping` promoted to `##` so the TOC entry resolves; rules, namespace model and worked examples added
- **`core-spec/spec.yaml`**, **`core-spec/ossie-schema.json`**: `Dataset.metrics` added and both metric descriptions documented
- **`validation/validate.py`**: scoping and field/metric collision checks added; one shared expression-parsing helper with caching replaces two duplicated parse paths
- **`validation/tests/test_validate.py`**: 21 test functions added to the suite from #330
- **`python/src/ossie/models.py`**: `OssieDataset.metrics` added
- **`python/tests/test_models.py`**: test walking schema `$defs` to catch schema/model drift
- **`docs/index.md`**, **`converters/README.md`**: corrected, both stated that metrics are model-level only

### Specification
- [x] Spec changes have been discussed on the mailing list or in a linked issue
- [x] Breaking changes are called out in the summary. There are none; the addition is optional and backward compatible

### Ontology
- [x] Ontology changes are consistent with spec changes. None required: `ontology.json` references `SemanticModel` by `$ref`, so it picks up `Dataset.metrics` without change
- [x] New or modified terms are defined and documented

### Converters
- [ ] Converter logic is updated. **Deliberately not done**, see "Not included". No converter reads `datasets[].metrics`; because the example is unchanged, none regresses
- [x] New converters include tests. No new converters

### Validation
- [x] Validation rules in `validation/` are updated

### Documentation
- [x] `docs/` is updated to reflect user-facing changes
- [x] New behaviours are documented with examples
- [x] `CONTRIBUTING.md` updated if the process changed. Not applicable

### Examples
- [ ] `examples/` updated for new spec constructs. **Deliberately not done**, so the flagship example is not lossy through converters that do not yet read the new placement

### Tests
- [x] All existing tests pass. Verified locally; note that no CI workflow covers these paths

### Compliance
- [x] ASF license headers present on all new source files. No new files

## AI disclosure

Per the ASF Generative Tooling Guidance, this contribution was prepared with AI assistance. All specification decisions, the scoping rules, and the responses to review are mine. I have reviewed and verified every change, and the verification results were produced by running the suites rather than asserted.

### DIFF
--- a/core-spec/ossie-schema.json
+++ b/core-spec/ossie-schema.json
@@ -220,7 +220,7 @@
           "items": {
             "$ref": "#/$defs/Metric"
           },
-          "description": "Dataset-scoped metrics. Expressions must resolve entirely within this dataset and must not traverse relationships or reference fields of another dataset. Names must be unique within the dataset and must not collide with any model-scoped metric name. Referenced from outside the dataset as dataset_name.metric_name. Metrics that span datasets belong in semantic_model.metrics."
+          "description": "Metrics that aggregate data held by this dataset, at this dataset's grain. The expression must aggregate only fields of this dataset, must not reference a field of another dataset, and must reference fields by unqualified name (SUM(amount), not SUM(orders.amount)). This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
         },
         "custom_extensions": {
           "type": "array",
@@ -343,7 +343,7 @@
           "items": {
             "$ref": "#/$defs/Metric"
           },
-          "description": "Model-scoped metrics. May span multiple datasets and traverse relationships. Names must be unique across the semantic model and must not collide with any dataset-scoped metric name. Metrics that resolve entirely within a single dataset may instead be defined on that dataset (datasets[].metrics)."
+          "description": "Model-scoped metrics. Expressions reference fields by qualified name (dataset.field). Names must be unique across the semantic model; reusing the name of a dataset-scoped metric should produce a warning rather than an error. An unqualified metric reference resolves to a model-scoped metric. See core-spec/spec.md#metric-scoping."
         },
         "custom_extensions": {
           "type": "array",
--- a/core-spec/ossie-schema.json
+++ b/core-spec/ossie-schema.json
@@ -220,7 +220,7 @@
           "items": {
             "$ref": "#/$defs/Metric"
           },
-          "description": "Metrics that aggregate data held by this dataset, at this dataset's grain. The expression must aggregate only fields of this dataset, must not reference a field of another dataset, and must reference fields by unqualified name (SUM(amount), not SUM(orders.amount)). This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
+          "description": "Metrics that aggregate data held by this dataset. The expression must reference fields of this dataset by unqualified name (SUM(amount), not SUM(orders.amount)); an expression referencing fields of more than one dataset is model-scoped and belongs in semantic_model.metrics. This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
         },
         "custom_extensions": {
           "type": "array",
--- a/core-spec/ossie-schema.json
+++ b/core-spec/ossie-schema.json
@@ -220,7 +220,7 @@
           "items": {
             "$ref": "#/$defs/Metric"
           },
-          "description": "Metrics that aggregate data held by this dataset. The expression is written against this dataset's source and must reference its columns by unqualified name (SUM(amount), not SUM(orders.amount)); a column used only inside a metric does not need to be declared as a field. An expression referencing columns of more than one dataset is model-scoped and belongs in semantic_model.metrics. This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
+          "description": "Metrics that aggregate data held by this dataset. The expression may reference this dataset's declared fields, written dataset_name.field_name, and the columns of its source, written unqualified; a column used only inside a metric does not need to be declared as a field. A qualified reference must name a declared field of this dataset. An expression referencing another dataset is model-scoped and belongs in semantic_model.metrics. This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
         },
         "custom_extensions": {
           "type": "array",
--- a/core-spec/ossie-schema.json
+++ b/core-spec/ossie-schema.json
@@ -220,7 +220,7 @@
           "items": {
             "$ref": "#/$defs/Metric"
           },
-          "description": "Metrics that aggregate data held by this dataset. The expression must reference fields of this dataset by unqualified name (SUM(amount), not SUM(orders.amount)); an expression referencing fields of more than one dataset is model-scoped and belongs in semantic_model.metrics. This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
+          "description": "Metrics that aggregate data held by this dataset. The expression is written against this dataset's source and must reference its columns by unqualified name (SUM(amount), not SUM(orders.amount)); a column used only inside a metric does not need to be declared as a field. An expression referencing columns of more than one dataset is model-scoped and belongs in semantic_model.metrics. This constrains the expression only, not the query: the metric is joined and grouped like any other, using the model's relationships. Names must be unique within the dataset and must not collide with a field name of this dataset. Referenced from outside the dataset as dataset_name.metric_name. See core-spec/spec.md#metric-scoping."
         },
         "custom_extensions": {
           "type": "array",
--- a/core-spec/ossie-schema.json
+++ b/core-spec/ossie-schema.json
@@ -215,6 +215,13 @@
             "$ref": "#/$defs/Field"
           }
         },
+        "metrics": {
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/Metric"
+          },
+          "description": "Dataset-scoped metrics. Expressions must resolve entirely within this dataset and must not traverse relationships or reference fields of another dataset. Names must be unique within the dataset and must not collide with any model-scoped metric name. Referenced from outside the dataset as dataset_name.metric_name. Metrics that span datasets belong in semantic_model.metrics."
+        },
         "custom_extensions": {
           "type": "array",
           "items": {
@@ -336,7 +343,7 @@
           "items": {
             "$ref": "#/$defs/Metric"
           },
-          "description": "Quantifiable measures spanning datasets"
+          "description": "Model-scoped metrics. May span multiple datasets and traverse relationships. Names must be unique across the semantic model and must not collide with any dataset-scoped metric name. Metrics that resolve entirely within a single dataset may instead be defined on that dataset (datasets[].metrics)."
         },
         "custom_extensions": {
           "type": "array",
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -432,21 +432,23 @@ A metric may be defined at the semantic model level or on an individual dataset.
 
 | | Model-scoped (`semantic_model.metrics`) | Dataset-scoped (`datasets[].metrics`) |
 |---|---|---|
-| Expression references | Fields of any dataset in the model, qualified: `dataset.field` | Columns of its own dataset's `source`, unqualified: `column` |
+| Expression references | Fields of any dataset in the model, qualified: `dataset.field` | Fields of its own dataset, qualified: `dataset.field`; columns of its `source`, unqualified: `column` |
 | Name uniqueness | Unique across the semantic model | Unique within its dataset, and distinct from that dataset's field names |
 | Referenced as | `metric_name` | `dataset_name.metric_name` |
 
 **Rules**
 
-1. A dataset-scoped metric's expression is written against its dataset's `source` and MUST reference columns by unqualified name: `SUM(ss_ext_sales_price)`, not `SUM(store_sales.ss_ext_sales_price)`. An expression that references columns of more than one dataset MUST be declared in `semantic_model.metrics`.
+1. A dataset-scoped metric's expression MAY reference the declared fields of its dataset and the columns of its `source`. A declared field is written `dataset_name.field_name`; a column of the `source` is written unqualified: `SUM(orders.net_amount) / COUNT(DISTINCT tax_id)`. A qualified reference MUST name a declared field of the declaring dataset. An expression that references another dataset MUST be declared in `semantic_model.metrics`.
 2. Dataset-scoped metric names MUST be unique within their dataset. Two datasets MAY each declare a metric with the same name.
 3. A dataset-scoped metric name MUST NOT collide with the name of a field of the same dataset, which under rule 5 would leave `orders.amount` ambiguous.
 4. A model-scoped metric SHOULD NOT reuse the name of a dataset-scoped metric in the same semantic model. Validators SHOULD warn, and SHOULD attribute the warning to the model rather than to the dataset.
 5. An unqualified metric reference resolves to a model-scoped metric. A dataset-scoped metric is referenced as `dataset_name.metric_name`.
 
-Rule 1 refers to the columns the dataset's `source` exposes, not to its declared fields. A column used only inside a metric does not need to be declared as a field first. This matches how a field's own expression is written: the field `customer_id` with expression `customer_id` refers to the source column, not to itself.
+A column used only inside a metric does not need to be declared as a field. Declaring one is how a metric reuses a field's expression instead of repeating it.
 
-A declared field therefore does not shadow a source column of the same name. If a dataset declares a field `foo` whose expression is not simply `foo`, then `SUM(foo)` in a metric of that dataset still refers to the source column `foo`. Reusing a declared field's expression inside a metric is the same composability question as a metric referencing another metric, and is left to a separate proposal.
+Because the two are spelled differently, a field and a source column MAY share a name without ambiguity. In a metric of `orders`, `orders.foo` is the declared field and `foo` is the source column, so declaring a field named after an existing column does not change the meaning of an expression already using the bare name.
+
+Whether a bare name is a real column of the `source` is not checked, because that requires catalog metadata the model does not carry. A qualified reference is checked, because the field list is in the model.
 
 Rule 1 places no limit on the complexity of the aggregation, and constrains the expression rather than the query. A dataset-scoped metric is joined and grouped like any other using the relationships declared in the model, so a metric declared on `store_sales` as `SUM(ss_ext_sales_price)` may still be grouped by `item.i_brand` or `store.s_state`.
 
@@ -477,20 +479,31 @@ datasets:
             - dialect: ANSI_SQL
               expression: order_id
         description: Order identifier
-      - name: amount
+      - name: net_amount
         expression:
           dialects:
             - dialect: ANSI_SQL
-              expression: amount
-        description: Order amount
+              expression: amount - discount
+        description: Order amount after discount
 
     metrics:
-      - name: total_amount
+      # Qualified, so this references the declared field net_amount
+      - name: total_net_amount
         expression:
           dialects:
             - dialect: ANSI_SQL
-              expression: SUM(amount)
-        description: Total order amount
+              expression: SUM(orders.net_amount)
+        description: Total order amount after discount
+        datatype: Decimal
+
+      # Unqualified, so this references tax, a column of the source that is
+      # not declared as a field
+      - name: total_tax
+        expression:
+          dialects:
+            - dialect: ANSI_SQL
+              expression: SUM(tax)
+        description: Total tax collected
         datatype: Decimal
 
       - name: order_count
@@ -502,7 +515,7 @@ datasets:
         datatype: Integer
 ```
 
-Referenced from a consumer as `orders.total_amount` and `orders.order_count`.
+Referenced from a consumer as `orders.total_net_amount`, `orders.total_tax` and `orders.order_count`.
 
 **Example: invalid dataset-scoped metrics**
 
@@ -510,21 +523,28 @@ Referenced from a consumer as `orders.total_amount` and `orders.order_count`.
 datasets:
   - name: orders
     source: sales.public.orders
+    fields:
+      - name: net_amount
+        expression:
+          dialects:
+            - dialect: ANSI_SQL
+              expression: amount - discount
     metrics:
-      # INVALID (rule 1): references a column of the customers dataset, so this
-      # metric is model-scoped and belongs in semantic_model.metrics
+      # INVALID (rule 1): references the customers dataset, so this metric is
+      # model-scoped and belongs in semantic_model.metrics
       - name: revenue_per_customer
         expression:
           dialects:
             - dialect: ANSI_SQL
-              expression: SUM(amount) / COUNT(DISTINCT customers.id)
+              expression: SUM(net_amount) / COUNT(DISTINCT customers.id)
 
-      # INVALID (rule 1): qualifies a column with the declaring dataset's own name
-      - name: total_amount
+      # INVALID (rule 1): a qualified reference must name a declared field, and
+      # tax is only a column of the source. Write it unqualified as SUM(tax)
+      - name: total_tax
         expression:
           dialects:
             - dialect: ANSI_SQL
-              expression: SUM(orders.amount)
+              expression: SUM(orders.tax)
 ```
 
 **Consumer guidance: flattening to a single metric namespace**
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -432,22 +432,23 @@ A metric may be defined at the semantic model level or on an individual dataset.
 
 | | Model-scoped (`semantic_model.metrics`) | Dataset-scoped (`datasets[].metrics`) |
 |---|---|---|
-| Expression may reference fields from | Any dataset in the model | Only its own dataset |
-| Expression namespace | Qualified: `dataset.field` | Unqualified: `field` |
+| Expression references | Fields of any dataset in the model, qualified: `dataset.field` | Columns of its own dataset's `source`, unqualified: `column` |
 | Name uniqueness | Unique across the semantic model | Unique within its dataset, and distinct from that dataset's field names |
 | Referenced as | `metric_name` | `dataset_name.metric_name` |
 
 **Rules**
 
-1. A dataset-scoped metric's expression MUST reference fields of its declaring dataset by unqualified name: `SUM(ss_ext_sales_price)`, not `SUM(store_sales.ss_ext_sales_price)`. An expression that references fields of more than one dataset MUST be declared in `semantic_model.metrics`.
+1. A dataset-scoped metric's expression is written against its dataset's `source` and MUST reference columns by unqualified name: `SUM(ss_ext_sales_price)`, not `SUM(store_sales.ss_ext_sales_price)`. An expression that references columns of more than one dataset MUST be declared in `semantic_model.metrics`.
 2. Dataset-scoped metric names MUST be unique within their dataset. Two datasets MAY each declare a metric with the same name.
 3. A dataset-scoped metric name MUST NOT collide with the name of a field of the same dataset, which under rule 5 would leave `orders.amount` ambiguous.
 4. A model-scoped metric SHOULD NOT reuse the name of a dataset-scoped metric in the same semantic model. Validators SHOULD warn, and SHOULD attribute the warning to the model rather than to the dataset.
 5. An unqualified metric reference resolves to a model-scoped metric. A dataset-scoped metric is referenced as `dataset_name.metric_name`.
 
-Rule 1 means declared fields, not any column the dataset's `source` exposes: a column used by a dataset-scoped metric must be declared as a field. This is not checked automatically, because an expression may also contain function names, literals, and struct or variant paths.
+Rule 1 refers to the columns the dataset's `source` exposes, not to its declared fields. A column used only inside a metric does not need to be declared as a field first. This matches how a field's own expression is written: the field `customer_id` with expression `customer_id` refers to the source column, not to itself.
 
-Rule 1 places no limit on the complexity of the aggregation, and constrains the expression rather than the query. Any expression that resolves within the declaring dataset is eligible, and a dataset-scoped metric is joined and grouped like any other using the relationships declared in the model, so a metric declared on `store_sales` as `SUM(ss_ext_sales_price)` may still be grouped by `item.i_brand` or `store.s_state`.
+A declared field therefore does not shadow a source column of the same name. If a dataset declares a field `foo` whose expression is not simply `foo`, then `SUM(foo)` in a metric of that dataset still refers to the source column `foo`. Reusing a declared field's expression inside a metric is the same composability question as a metric referencing another metric, and is left to a separate proposal.
+
+Rule 1 places no limit on the complexity of the aggregation, and constrains the expression rather than the query. A dataset-scoped metric is joined and grouped like any other using the relationships declared in the model, so a metric declared on `store_sales` as `SUM(ss_ext_sales_price)` may still be grouped by `item.i_brand` or `store.s_state`.
 
 **Which names may repeat**
 
@@ -510,7 +511,7 @@ datasets:
   - name: orders
     source: sales.public.orders
     metrics:
-      # INVALID (rule 1): references a field of the customers dataset, so this
+      # INVALID (rule 1): references a column of the customers dataset, so this
       # metric is model-scoped and belongs in semantic_model.metrics
       - name: revenue_per_customer
         expression:
@@ -518,7 +519,7 @@ datasets:
             - dialect: ANSI_SQL
               expression: SUM(amount) / COUNT(DISTINCT customers.id)
 
-      # INVALID (rule 1): qualifies a field with the declaring dataset's own name
+      # INVALID (rule 1): qualifies a column with the declaring dataset's own name
       - name: total_amount
         expression:
           dialects:
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -434,32 +434,20 @@ A metric may be defined at the semantic model level or on an individual dataset.
 |---|---|---|
 | Expression may reference fields from | Any dataset in the model | Only its own dataset |
 | Expression namespace | Qualified — `dataset.field` | Unqualified — `field` |
-| Aggregation grain | Determined by the expression | The declaring dataset's grain |
 | Name uniqueness | Unique across the semantic model | Unique within its dataset, and distinct from that dataset's field names |
 | Referenced as | `metric_name` | `dataset_name.metric_name` |
 
 **Rules**
 
-1. A dataset-scoped metric's expression MUST aggregate only fields of the dataset that declares it, and MUST NOT reference a field of another dataset. Model-scoped metrics carry no equivalent restriction.
-2. A dataset-scoped metric's expression MUST reference fields by unqualified name: `SUM(ss_ext_sales_price)`, not `SUM(store_sales.ss_ext_sales_price)`.
-3. Dataset-scoped metric names MUST be unique within their dataset. Two datasets MAY each declare a metric with the same name.
-4. A dataset-scoped metric name MUST NOT collide with the name of a field of the same dataset, which under rule 6 would leave `orders.amount` ambiguous.
-5. A model-scoped metric SHOULD NOT reuse the name of a dataset-scoped metric in the same semantic model. Validators SHOULD warn, and SHOULD attribute the warning to the model rather than to the dataset.
-6. An unqualified metric reference resolves to a model-scoped metric. A dataset-scoped metric is referenced as `dataset_name.metric_name`.
+1. A dataset-scoped metric's expression MUST reference fields of its declaring dataset by unqualified name: `SUM(ss_ext_sales_price)`, not `SUM(store_sales.ss_ext_sales_price)`. An expression that references fields of more than one dataset MUST be declared in `semantic_model.metrics`.
+2. Dataset-scoped metric names MUST be unique within their dataset. Two datasets MAY each declare a metric with the same name.
+3. A dataset-scoped metric name MUST NOT collide with the name of a field of the same dataset, which under rule 5 would leave `orders.amount` ambiguous.
+4. A model-scoped metric SHOULD NOT reuse the name of a dataset-scoped metric in the same semantic model. Validators SHOULD warn, and SHOULD attribute the warning to the model rather than to the dataset.
+5. An unqualified metric reference resolves to a model-scoped metric. A dataset-scoped metric is referenced as `dataset_name.metric_name`.
 
 Rule 1 means declared fields, not any column the dataset's `source` exposes: a column used by a dataset-scoped metric must be declared as a field. This is not checked automatically, because an expression may also contain function names, literals, and struct or variant paths.
 
-Rule 1 places no limit on the complexity of the aggregation. Any expression that resolves within the declaring dataset is eligible.
-
-**Scope describes the aggregation, not the query**
-
-A dataset-scoped metric aggregates data held by its own dataset. That is all the placement asserts. It does not limit how the metric may be queried: the metric is joined and grouped like any other, using the relationships declared in the model, so it can be sliced by dimensions of any dataset the model connects. Rules 1 and 2 constrain what an expression may reference, never what may be joined to the result.
-
-Presenting many datasets and their metrics through one queryable interface is the concern of a layer above this one. This section defines only where an aggregation is anchored.
-
-**Aggregation grain**
-
-A dataset-scoped metric aggregates at its dataset's grain. That grain does not depend on a declared `primary_key`: it is the grain of the rows the `source` produces. `primary_key` and `unique_keys` let consumers reason about fan-out when the dataset participates in relationships, but are not a prerequisite.
+Rule 1 places no limit on the complexity of the aggregation, and constrains the expression rather than the query. Any expression that resolves within the declaring dataset is eligible, and a dataset-scoped metric is joined and grouped like any other using the relationships declared in the model, so a metric declared on `store_sales` as `SUM(ss_ext_sales_price)` may still be grouped by `item.i_brand` or `store.s_state`.
 
 **Which names may repeat**
 
@@ -472,7 +460,7 @@ Names are addressed in three ways, so a name may repeat across them without ambi
 | Field | Qualified — `orders.revenue` |
 | Dataset | Only ever as a qualifier |
 
-A model-scoped metric MAY therefore share a name with a field, or with a dataset. Rule 4 is an error because a field and a metric of one dataset share the same qualified namespace, so `orders.revenue` would resolve two ways. Rule 5 is a warning because the two names remain separately addressable, and because a dataset may be authored independently of the model that includes it.
+A model-scoped metric MAY therefore share a name with a field, or with a dataset. Rule 3 is an error because a field and a metric of one dataset share the same qualified namespace; rule 4 is a warning because the two names remain separately addressable.
 
 **Example — dataset-scoped metrics**
 
@@ -522,42 +510,21 @@ datasets:
   - name: orders
     source: sales.public.orders
     metrics:
-      # INVALID (rule 1): references a field of the customers dataset
+      # INVALID (rule 1): references a field of the customers dataset, so this
+      # metric is model-scoped and belongs in semantic_model.metrics
       - name: revenue_per_customer
         expression:
           dialects:
             - dialect: ANSI_SQL
               expression: SUM(amount) / COUNT(DISTINCT customers.id)
 
-      # INVALID (rule 2): qualifies a field with the declaring dataset's own name
+      # INVALID (rule 1): qualifies a field with the declaring dataset's own name
       - name: total_amount
         expression:
           dialects:
             - dialect: ANSI_SQL
               expression: SUM(orders.amount)
 ```
-
-**Prior art**
-
-Systems in this space differ in where a single-entity aggregation lives, how names resolve, and how strictly scope is enforced.
-
-| System | Single-entity metric | Cross-entity mechanism | Name uniqueness | Reference form |
-|---|---|---|---|---|
-| **AtScale SML** | Standalone `metric` object bound to one `dataset` + `column` | Separate `metric_calc` object type | Global across all repositories | Bare `unique_name` |
-| **Cube** | Measures within cubes | Calculated measures referencing other measures | Per cube | Qualified — `cube.member` |
-| **Databricks UC metric views** | Measures in the view (one flat scope) | Joins declared inside the view | Per metric view | `MEASURE(name)` |
-| **dbt MetricFlow** (v1.12+) | Metrics inside a semantic model | Top-level `metrics` | Global across the project | Bare name |
-| **Snowflake semantic views** | `tables[].metrics` | Top-level `metrics` (derived) | Per logical table | Qualified — `table.metric` |
-
-Four of the five distinguish the two structurally. Ossie has so far provided only the model-level placement, which is the gap this section addresses.
-
-Two points of comparison worth recording. Databricks UC metric views are closest to the namespace rules above: a metric view has one `source` plus optional `joins`, sources are named so a column is addressed as `source_name.column_name`, and the metric view itself exposes a flat schema. Snowflake semantic views let a table-scoped metric's own expression reach through a relationship, so the boundary between their two placements is softer than rule 1.
-
-Scoped uniqueness with qualified references, rather than the flat global namespace used by AtScale SML and dbt MetricFlow, follows Ossie's existing convention: field names are already scoped to a dataset. Rule 2 follows the same convention, since a field's own expression is written without naming its dataset.
-
-Rule 1 is the conservative choice. It keeps a dataset-scoped metric verifiably anchored to one dataset, so a dataset and its metrics can be exchanged without resolving the surrounding join graph. Relaxing it later would be backward compatible; tightening it would not. Whether a dataset-scoped metric's expression should be permitted to reach through an explicitly declared path is left open.
-
-Choosing between multiple relationship paths, where more than one connects the same pair of datasets, is a separate gap that Ossie does not currently address. It applies to model-scoped expressions and consumer queries alike.
 
 **Consumer guidance: flattening to a single metric namespace**
 
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -364,8 +364,8 @@ Quantitative measures defined on business data, representing key calculations li
 
 Metrics may be defined in two placements, using the same structure in both:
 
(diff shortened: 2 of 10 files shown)


core-spec/ossie-schema.json
```diff
@@ -215,6 +215,13 @@
             "$ref": "#/$defs/Field"
           }
         },
+        "metrics": {
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/Metric"
+          },
+          "description": "Dataset-scoped metrics. Expressions must resolve entirely within this dataset and must not traverse relationships or reference fields of another dataset. Names must be unique within the dataset and must not collide with any model-scoped metric name. Referenced from outside the dataset as dataset_name.metric_name. Metrics that span datasets belong in semantic_model.metrics."
```
khush-bhatia: How about clarifying here that the dataset scoped metrics are computed at the grain of the dataset's primary key. Also why restrict relationship traversal ?

core-spec/spec.yaml
```diff
@@ -235,6 +258,10 @@ fields:
 # Metrics Schema
 # Quantitative measures defined on business data
 # Represents key calculations like sums, averages, ratios, etc.
+#
+# The same structure is used in two placements:
+# - semantic_model.metrics  (model-scoped): may span datasets via relationships
```
khush-bhatia, line 271: Let's add here that `semantic_model.metrics` can be used to combine multiple dataset scoped metrics ?

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+| | Model-scoped (`semantic_model.metrics`) | Dataset-scoped (`datasets[].metrics`) |
+|---|---|---|
+| Expression may reference fields from | Any dataset in the model | Only its own dataset |
+| Expression may traverse relationships | Yes | No |
+| Name uniqueness | Unique across the semantic model | Unique within its dataset |
+| Referenced as | `metric_name` | `dataset_name.metric_name` |
+
+**Rules**
+
+1. A dataset-scoped metric's expression MUST only reference fields of the dataset that declares it. It MUST NOT traverse relationships or reference fields belonging to another dataset. A metric that needs to span datasets MUST be model-scoped.
```
christianeu-db: Does the metric's expression reference the dataset's fields, the dataset's source's columns or both (with some disambiguation mechanism)?

In the example below, COUNT(order_id) references a column in the source, not a field in the containing dataset.

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+2. Dataset-scoped metric names MUST be unique within their dataset. Two different datasets MAY each declare a metric with the same name (e.g. `orders.item_count` and `shipments.item_count`).
```
christianeu-db: Should uniqueness apply across both fields & metrics (e.g. can a metric and a field in a dataset have the same name)?

Uniqueness would guarantee that `dataset`.`name` referred to exactly one of a field or metric, not both.

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+4. Dataset-scoped metrics are referenced from outside their dataset using `dataset_name.metric_name`, mirroring how a dataset's fields are already referenced in metric expressions (e.g. `SUM(orders.amount)`).
+
+**Scope restricts the expression, not the query**
+
+Rule 1 constrains only what a metric's *expression* may reference. It does not restrict how the metric may be queried. A dataset-scoped metric can still be grouped by, or filtered on, dimensions from other datasets reached through relationships — grouping dimensions are supplied by the consumer at query time and are not part of the metric definition.
+
+For example, a metric declared on `store_sales` as `SUM(store_sales.ss_ext_sales_price)` is dataset-scoped because its expression touches only `store_sales`, yet it remains valid to group that metric by `item.i_brand` or `store.s_state` via the model's relationships. Only a metric whose own expression must reach into another dataset — such as `SUM(store_sales.amount) / COUNT(DISTINCT customer.id)` — needs to be model-scoped.
```
christianeu-db: Should dataset-scoped metric expressions use a single-layer namespace since they can only reference fields from the dataset and/or source (depending on this [discussion](https://github.com/apache/ossie/pull/343/changes#r3876450874))?

For example, store_sales-scoped metric would have the expression `SUM(ss_ext_sales_price)` instead of  `SUM(store_sales.ss_ext_sales_price)`. This would be more consistent with fields and lines up with some of the [layering](https://docs.google.com/document/d/1EgG5vOlJSubm01S1AGMJ4MKcMcjxBhJHenGKWMr849E/edit?tab=t.0#heading=h.k8nomy2jicfx) framing (i.e. dataset-scoped objects don't reason about other datasets)

If a single-dataset metric references the two level namespace, that seems to be relational metrics layer being aware of concepts (two-layer names) from the layer above.

core-spec/ossie-schema.json
christianeu-db: 1. Agree that dataset-scoped metrics are computed at the dataset's grain. However, this grain should be safely computable without requiring a PK.
2. Agree with the relationship traversal restriction. To follow on to the [layering framing](https://docs.google.com/document/d/1EgG5vOlJSubm01S1AGMJ4MKcMcjxBhJHenGKWMr849E/edit?tab=t.0#heading=h.k8nomy2jicfx), dataset-scoped metrics belong to the relational / metric layer. In that layer, each dataset acts as a self-contained table that can be independently queried / composed in SQL. If databases supported this layer natively, Ossie datasets could be pushed down to the database layer & used by multiple Ossie models. Relationship traversal is more tied to the multi-table semantics within a Ossie model (i.e. it depends on the graph shape), which would mean that datasets now need to reason about each other in the context of some graph

core-spec/spec.yaml
christianeu-db: To Josh's PR review questions, this fits with the question on if / how to build metrics from other metrics. That's probably worth a followup PR / discussion since it touches on how to extend the expression language

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+| Expression may reference fields from | Any dataset in the model | Only its own dataset |
+| Expression may traverse relationships | Yes | No |
+| Name uniqueness | Unique across the semantic model | Unique within its dataset |
+| Referenced as | `metric_name` | `dataset_name.metric_name` |
+
+**Rules**
+
+1. A dataset-scoped metric's expression MUST only reference fields of the dataset that declares it. It MUST NOT traverse relationships or reference fields belonging to another dataset. A metric that needs to span datasets MUST be model-scoped.
+2. Dataset-scoped metric names MUST be unique within their dataset. Two different datasets MAY each declare a metric with the same name (e.g. `orders.item_count` and `shipments.item_count`).
+3. A dataset-scoped metric name MUST NOT collide with the name of any model-scoped metric in the same semantic model. This keeps an unqualified metric reference unambiguous.
```
christianeu-db: Could the rule be that only model-scoped metrics can use unqualified names? 

The requirement that unqualified metric references must be unambiguous seems to conflict with rule 2, which could allow for two datasets to have metrics with the same name. Rule 2 seems more consistent with field behavior / having datasets not reason about name uniqueness across each other.

Then, the model could be:
1. Unscoped names: model-level
2. Scoped names: dataset-level

Some tools, such as Tableau, support cross-dataset dimensions, which also live in the global namespace. In their model, they have both dataset-scoped metrics/dimensions and model-scoped metrics/dimensions

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+
+Separating an aggregation anchored to a single entity from a calculation that spans entities is established practice, though systems differ in where the single-entity metric lives, how names resolve, and how strictly scope is enforced.
+
+| System | Single-entity metric | Cross-entity mechanism | Name uniqueness | Reference form |
+|---|---|---|---|---|
+| **Snowflake semantic views** | `tables[].metrics` | Top-level `metrics` (derived) | Per logical table | Qualified — `table.metric` |
+| **AtScale SML** | Standalone `metric` object bound to one `dataset` + `column` | Separate `metric_calc` object type | Global across all repositories | Bare `unique_name` |
+| **dbt MetricFlow** (v1.12+) | Metrics inside a semantic model | Top-level `metrics` | Global across the project | Bare name |
+| **Cube** | Measures within cubes | Calculated measures referencing other measures | Per cube | Qualified — `cube.member` |
+| **Databricks UC metric views** | Measures in the view (one flat scope) | Joins declared inside the view | Per metric view | `MEASURE(name)` |
```
christianeu-db: For metric views, it's close to named subqueries when building a view. Each source has a name so within a metric view, field names are <source_name>.<field_name> but the metric view itself acts like a table so its schema has a flat scope

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+2. Dataset-scoped metric names MUST be unique within their dataset. Two different datasets MAY each declare a metric with the same name (e.g. `orders.item_count` and `shipments.item_count`).
+3. A dataset-scoped metric name MUST NOT collide with the name of any model-scoped metric in the same semantic model. This keeps an unqualified metric reference unambiguous.
+4. Dataset-scoped metrics are referenced from outside their dataset using `dataset_name.metric_name`, mirroring how a dataset's fields are already referenced in metric expressions (e.g. `SUM(orders.amount)`).
+
+**Scope restricts the expression, not the query**
+
+Rule 1 constrains only what a metric's *expression* may reference. It does not restrict how the metric may be queried. A dataset-scoped metric can still be grouped by, or filtered on, dimensions from other datasets reached through relationships — grouping dimensions are supplied by the consumer at query time and are not part of the metric definition.
```
christianeu-db: Probably worth clarifying that within the layering proposal, the ability to natively query across datasets from a single interface would be in what is called the dimensional layer (Will was also proposing "presentation layer" as an alternate name)

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+
+For example, a metric declared on `store_sales` as `SUM(store_sales.ss_ext_sales_price)` is dataset-scoped because its expression touches only `store_sales`, yet it remains valid to group that metric by `item.i_brand` or `store.s_state` via the model's relationships. Only a metric whose own expression must reach into another dataset — such as `SUM(store_sales.amount) / COUNT(DISTINCT customer.id)` — needs to be model-scoped.
+
+**Choosing a placement**
+
+Prefer dataset-scoped for simple aggregations that belong conceptually to one entity — they keep the metric next to the fields it depends on and make the dataset independently interpretable. Use model-scoped for anything requiring a join.
```
christianeu-db: (nit) The aggregations themselves don't necessarily need to be simple. More complex aggregates (e.g. later window calculation extensions) could be dataset-scoped

christianeu-db (commented):
Thanks for putting this PR together - definitely agree with adding dataset-scoped metrics. Conceptually, this fits in nicely with the [layered model](https://docs.google.com/document/d/1EgG5vOlJSubm01S1AGMJ4MKcMcjxBhJHenGKWMr849E/edit?tab=t.0#heading=h.k8nomy2jicfx) from the working group.

Most of my questions are around making dataset-scoped metrics consistent with fields in terms of name-uniqueness, how they reference fields in their expressions, and the requirement to always be called with a two-level name. Another question is around what can be an input to a dataset-scoped metric.

With respect to your open questions:
1. Relationship traversal information seems like a model-level concept since it spans across datasets.
2. Agreed that defining metric / other reference rules should be a separate discussion since that's a pretty fundamental part of the spec

examples/tpcds_semantic_model.yaml
```diff
@@ -148,6 +148,50 @@ semantic_model:
                 - "profit"
                 - "margin"
 
+        # Dataset-scoped metrics: expressions resolve entirely within
+        # store_sales. These may still be grouped by dimensions of other
+        # datasets (for example item.i_brand or store.s_state) through the
+        # model's relationships. Referenced as store_sales.<metric_name>.
+        metrics:
```
jbonofre: Two tests read this file directly and assert on model-level metrics:
* `converters/omni/tests/test_ossie_to_omni.py` `test_tpcds_export_matches_expected` loads this file and compare the export against `tests/fixtures/tpcds_omni`
* `converters/omni/tests/test_roundtrip.py` `test_ossie_roundtrip_up_to_documented_normalizations`

Both fail due to this change (`cd converters/omni && uv run pytest`). 

I suggest to update the omni converter to host dataset-scoped metrics (updating the fixture in this PR). That would be my preference.

core-spec/ossie-schema.json
```diff
@@ -215,6 +215,13 @@
             "$ref": "#/$defs/Field"
           }
         },
+        "metrics": {
```
jbonofre, line 218: `OssieDataset` in `python/src/ossie/models.py` (the model shipped as the `apache-ossie` package) has no `metrics` field. Pydantic's default `extra='ignore'` means dataset-scoped metrics are discarded on load with no error:

```python
doc = OssieDocument.model_validate(yaml.safe_load(open('examples/tpcds_semantic_model.yaml')))
# store_sales has no `metrics` key; only customer_lifetime_value and
# store_productivity survive to_ossie_yaml(). Three metrics vanish silently.
```

Adding `metrics: list[OssieMetric] | None = None` to `OssieDataset` fixes it.

Worth noting the reason nothing caught this: `python/tests/test_models.py:75` only cross-checks the `DataType` enum against the JSON Schema, so structural drift between the schema and the pydantic model is invisible. A test that walks `$defs` properties and asserts each one exists on the corresponding model would have failed here.

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+
+The rationale for disallowing traversal is that a strict boundary makes the guarantee legible: a dataset-scoped metric is verifiably self-contained, so a dataset together with its metrics can be reasoned about, reused, or exchanged without resolving the surrounding join graph. Relaxing this later would be backward compatible; tightening it would not. Whether Ossie should eventually adopt a `using_relationships` equivalent is left as an open question.
+
+**Consumer guidance: flattening to a single metric namespace**
+
+Consumers whose native model has only model-level metrics do not need to represent the two placements separately. Because a dataset-scoped metric's expression resolves entirely within its declaring dataset, that expression is already valid as a model-scoped metric — hoisting requires no expression rewriting.
```
jbonofre: Every converter reads model-level metrics only:

| Converter | Location |
|---|---|
| omni | `converters/omni/src/ossie_omni/ossie_to_omni.py:187` |
| honeydew | `converters/honeydew/.../converter.py:136` |
| snowflake | `converters/snowflake/.../converter.py:169` |
| nvidia | `converters/nvidia/.../native_converter.py:322` |
| orionbelt | `converters/orionbelt/.../ossie_to_obml.py:170` |
| databricks | `converters/databricks/python/.../ossie_to_metric_view.py:168` |
| polaris | `converters/polaris/.../OssieYamlGenerator.java:72` |

Convert the new TPC-DS example to any target today and three metrics disappear with no warning. Compare `converters/honeydew`, which at least `warnings.warn`s when it has to guess.

"Consumers that read only `semantic_model.metrics` remain valid" is a reasonable spec position, but combined with moving the reference example it means the flagship model is lossy through the entire hub-and-spoke on day one. 

Two things would help:

* A shared hoist helper in the spec tooling (`iter_metrics(model)` yielding both placements with qualified names), so converters get this right by default rather than each reimplementing it.
* Stronger wording here: a consumer that ignores `datasets[].metrics` silently produces an incomplete model, which is a lossy conversion, not a valid one. At minimum it SHOULD warn.

validation/validate.py
```diff
@@ -149,6 +169,75 @@ def validate_references(data: dict) -> list[str]:
...
+                    dialect = dialect_expr.get("dialect", "ANSI_SQL")
+                    expr = dialect_expr.get("expression", "")
+                    if not expr:
+                        continue
+
+                    qualifiers = _expression_qualifiers(expr, dialect)
+                    if qualifiers is None:
+                        continue
+
+                    foreign = sorted(q for q in qualifiers if q != dataset_name)
```
jbonofre: SQL identifiers are case-insensitive, and sqlglot preserves the raw text of unquoted identifiers, so `q != dataset_name` is comparing raw casing against the YAML name.

To reproduce, use dataset `orders`, dialect `SNOWFLAKE`, expression `SUM(ORDERS.AMOUNT)`:

```
[Scope] Dataset-scoped metric 'orders.total' in model 'm' (SNOWFLAKE) references dataset(s) 'ORDERS'.
```

So a correct model is rejected, and the error names a dataset that does not exist. Uppercase is the natural casing for Snowflake semantic views, which the PR body names as the motivating consumer.

Fold case before comparing (`q.casefold() != dataset_name.casefold()`). Strictly correct handling also depends on whether the identifier was quoted, but case-folding unquoted identifiers covers the realistic cases and is a clear improvement on the current behavior.

validation/validate.py
```diff
@@ -149,6 +169,75 @@ def validate_references(data: dict) -> list[str]:
...
+    sqlglot_dialect = DIALECT_MAP.get(dialect)
+
+    for candidate in (expr, f"SELECT {expr}"):
+        try:
+            tree = sqlglot.parse_one(candidate, dialect=sqlglot_dialect)
+        except (ParseError, TokenError):
+            continue
+        if tree is None:
+            continue
+        return {col.table for col in tree.find_all(exp.Column) if col.table}
```
jbonofre: Verified with sqlglot:

```python
qualifiers("SUM(orders.payload.amount)")  # -> {'payload'}   (not 'orders')
qualifiers("SUM(payload.amount)")         # -> {'payload'}
```

For a three-part path sqlglot puts the middle part in `col.table`, so a dataset-scoped metric over a `STRUCT`/`VARIANT` column on `orders` is rejected with `references dataset(s) `payload'`.

There is no way to express such a metric at dataset scope at all, the only escape is moving it to model scope, where the rule does not apply.

`exp.Column` has `catalog`/`db`/`table`/`this` parts, for the three-part case you want to look at `col.db` (or `col.parts[0]`) as the potential dataset qualifier, and treat a bare two-part path whose first element is a declared field name as intra-dataset.

validation/validate.py
```diff
@@ -149,6 +169,75 @@ def validate_references(data: dict) -> list[str]:
...
+                    qualifiers = _expression_qualifiers(expr, dialect)
+                    if qualifiers is None:
+                        continue
+
+                    foreign = sorted(q for q in qualifiers if q != dataset_name)
+                    if foreign:
+                        errors.append(
+                            f"[Scope] Dataset-scoped metric '{dataset_name}.{metric_name}' "
+                            f"in model '{model_name}' ({dialect}) references "
+                            f"dataset(s) {', '.join(repr(f) for f in foreign)}. "
```
jbonofre: Verified with sqlglot:

```python
qualifiers("SUM(o.amount)")
# -> {'o'}
qualifiers("(SELECT SUM(x) FROM other WHERE other.id = orders.id)")
# -> {'orders', 'other'}
```

Both produce `references dataset(s) 'o'` / `'other'` followed by "move this metric to `semantic_model.metrics`": advise that would not fix either case, since neither name is a dataset 😄 

`validate_references` already collects the model's dataset names (around line 137). Cross-checking against that set would let this distinguish three genuinely different situations: a real cross-dataset reference (move it to model scope), a local alias or CTE (fine, ignore), and an unresolvable qualifier (a different error entirely).

validation/validate.py
```diff
@@ -203,6 +292,19 @@ def validate_sql(data: dict) -> list[str]:
                         if error:
                             errors.append(error)
 
+            # Validate dataset-scoped metric expressions
+            for metric in dataset.get("metrics", []):
+                metric_name = metric.get("name", "<unnamed>")
+                expression = metric.get("expression", {})
```
jbonofre: `validate_metric_scoping` uses `metric.get("expression") or {}` (line 213), which handles an explicit null. This line uses the two-arg default, which does not - `.get(k, {})` returns `None` when the key is present with a null value.

To reproduce, we can use `metrics: [{name: total, expression: }]` (a truncated hand-edit) gives:

```
AttributeError: 'NoneType' object has no attribute 'get'
```

so the user gets a traceback instead of the schema error that was already collected.

```suggestion
                expression = metric.get("expression") or {}
```

Line 311 (the pre-existing model-level loop) has the same latent issue; worth fixing both for consistency.

validation/validate.py
```diff
@@ -120,6 +122,24 @@ def validate_unique_names(data: dict) -> list[str]:
...
+        for dataset in model.get("datasets", []):
+            dataset_name = dataset.get("name", "<unnamed>")
+            ds_metric_names = [
+                m.get("name") for m in dataset.get("metrics", []) if m.get("name")
+            ]
+            for dup in find_duplicates(ds_metric_names):
+                errors.append(
+                    f"[Unique] Duplicate metric name '{dup}' in dataset '{dataset_name}'"
+                )
+            for name in set(ds_metric_names) & model_metric_names:
```
jbonofre: With more than one collision the output order changes per process (string hash randomisation). Four runs with datasets and model each declaring `a,b,c,d`:

```
['ds.a','ds.b','ds.d','ds.c']
['ds.b','ds.a','ds.d','ds.c']
['ds.a','ds.d','ds.c','ds.b']
['ds.c','ds.d','ds.a','ds.b']
```

`find_duplicates` (line 88) deliberately preserves input order, so this is inconsistent with the surrounding code as well as unfriendly to golden-output comparison and CI log diffs.

```suggestion
            for name in sorted(set(ds_metric_names) & model_metric_names):
```

validation/validate.py
```diff
@@ -149,6 +169,75 @@ def validate_references(data: dict) -> list[str]:
     return errors
 
 
+def _expression_qualifiers(expr: str, dialect: str) -> set[str] | None:
```
jbonofre: Lines 187–195 repeat lines 162–174 almost exactly: same `(expr, f"SELECT {expr}")` candidates, same `(ParseError, TokenError)` handling, same `SQLGLOT_AVAILABLE`/`SKIP_SQL_VALIDATION`/`DIALECT_MAP` preamble. The two copies already differ in how they treat `tree is None`, which is how this kind of duplication starts to drift.

Since `validate_metric_scoping` and `validate_sql` both run over every dataset-scoped metric expression, each expression is now parsed up to four times.

Extracting one helper: `_parse_expression(expr, dialect) -> Expression | None` — and having both callers use it removes the duplication and halves the parsing.

core-spec/spec.md
jbonofre: Rule 4 makes a dataset-scoped metric reachable as `dataset_name.metric_name`, which is exactly the form a field already uses. Nothing forbids a metric from taking a field's name in the same dataset, and the validator accepts it:

```yaml
- name: orders
  fields:
    - name: amount        # ...
  metrics:
    - name: amount        # expression: SUM(orders.amount)
```

→ `Validation PASSED`.

A consumer resolving `orders.amount` now cannot tell whether it means the row-level field or the aggregate, and the metric's own expression `SUM(orders.amount)` becomes self-referential under rule 4. Suggest a rule 5: a dataset-scoped metric name MUST NOT collide with the name of any field of the same dataset: plus the corresponding check in `validate_unique_names`, which already has both name lists in scope.

core-spec/spec.md
```diff
@@ -130,6 +131,7 @@ Logical datasets represent business entities or concepts (fact and dimension tab
 | `description` | string | No | Human-readable description |
 | `ai_context` | string/object | No | Additional context for AI tools (e.g., synonyms, common terms) |
 | `fields` | array | No | Row-level attributes for grouping, filtering, and metric expressions |
+| `metrics` | array | No | Dataset-scoped metrics whose expressions resolve entirely within this dataset. See [Metric Scoping](#metric-scoping). |
```
jbonofre, line 134: - `docs/index.md:251` (FAQ) — "Metrics are defined at the semantic model level (not within a dataset)"
- `docs/index.md:58` and `docs/index.md:321` (glossary) — same claim
- `converters/README.md:169` — "Metrics are aggregate measures defined at the semantic model level"
- `converters/README.md:97` — the dataset property table omits `metrics`

The converter guide one matters most: an author following it writes a converter that drops dataset-scoped metrics, which is exactly the failure mode already present in all the existing converters.

core-spec/spec.md
```diff
@@ -573,16 +710,14 @@ semantic_model:
         to_columns: [id]
 
     metrics:
-      - name: total_revenue
+      # Model-scoped: spans orders and customers via the relationship
```
jbonofre, line 736: The comment says "spans orders and customers via the relationship", but it applies to a `metrics:` list whose second entry (`customer_count`, line 722) is `COUNT(DISTINCT customers.id)`, single-dataset. By the guidance at line 453 ("prefer dataset-scoped for simple aggregations that belong conceptually to one entity") it belongs on `customers.metrics`.

Commit 4 moved all three single-dataset metrics in the TPC-DS example for exactly this reason, so the two examples in the PR now teach opposite things. Either move `customer_count` onto `customers` here too, or scope the comment to `revenue_per_customer` alone.

core-spec/spec.md
```diff
@@ -415,6 +424,124 @@ expression:
...
+              expression: SUM(orders.amount)
+        description: Total order amount
+        datatype: Decimal
+
+      # Also valid: unqualified reference to a field of the declaring dataset
+      - name: order_count
+        expression:
+          dialects:
+            - dialect: ANSI_SQL
+              expression: COUNT(order_id)
```
jbonofre, line 513: The `orders` dataset above declares exactly one field, `amount`; `order_id` appears only in `primary_key`. So the metric introduced as "unqualified reference to a field of the declaring dataset" references something that is not among the declaring dataset's fields, the example undercuts the point it is making.

Either add `order_id` to `fields`, or use `COUNT(amount)`.

core-spec/spec.md
```diff
@@ -37,7 +37,8 @@
 4. [Relationships](#relationships)
 5. [Fields](#fields)
 6. [Metrics](#metrics)
-7. [Examples](#examples)
+7. [Metric Scoping](#metric-scoping)
```
jbonofre, line 40: nit: every other numbered entry (1–6, 8) maps to a `##` heading, but `Metric Scoping` is authored as `### Metric Scoping` at line 427, nested under `## Metrics`. Rendered docs get a top-level entry that jumps into the middle of the Metrics section.

Either promote it to `## Metric Scoping`, or drop the TOC entry and keep the in-section link at line 370.

core-spec/ossie-schema.json
khush-bhatia: Let's add here that metrics and fields in the dataset share the same namespace. So the names should be unique across fields and metrics in dataset.

willpugh (commented):
I would like us to focus on the spec, before the code changes.  A few issues:
   1). Why not just remove the no-aggregation restriction on fields?
   2) We should probably work on the semantics first.  Particularly around relationship traversal
   3) I'm assuming we are going to want grain locking here, but don't think you covered it.

jklahr:
Thanks all, this was a genuinely useful round. Two things changed materially as a result: the expression namespace is now unqualified, and the name-collision rules are split into errors and warnings rather than all being errors.

The branch is also rebased onto current `main`, which brought in #330. That turned out to overlap: #330 added `validation/tests/test_validate.py`, so the tests here extend that file rather than introduce one. Worth noting that #330's new check emits `[Reference] Warning: ...`, the same convention I had chosen for rule 5 below, arrived at independently.

Every comment is answered below, grouped by reviewer. Verification for each claim is at the end.

---

## @jbonofre

Six of these were bugs with working reproductions. All six are fixed, and each has a test that fails before the change and passes after.

**[`validate.py:137`](https://github.com/apache/ossie/pull/343#discussion_r3880506834), non-deterministic collision ordering.** Applied your `sorted()` suggestion. Verified across six runs in one process that the output is now identical every time.

**[`validate.py:172`](https://github.com/apache/ossie/pull/343#discussion_r3880509505), duplicated parse logic, up to four parses per expression.** Extracted a single `_parse_expression(expr, dialect) -> (tree, error)` used by both callers. This also unified the `tree is None` divergence you spotted: both paths now treat it as a failed candidate and try the next form. I added `lru_cache` on top, so repeated checks over the same expression parse once rather than twice; verified by asserting cache hits increase across a scoping pass followed by a SQL pass. The returned tree is shared, so its docstring states it must not be mutated.

**[`validate.py:194`](https://github.com/apache/ossie/pull/343#discussion_r3880456599), three-part paths.** Fixed using `parts[0]` rather than `col.table`, as you suggested. `SUM(payload.amount)` where `payload` is a declared field is now accepted. I also checked the case you did not raise: `SUM(orders.payload.amount)` now correctly reports the *dataset* qualifier rather than `payload`.

**[`validate.py:228`](https://github.com/apache/ossie/pull/343#discussion_r3880428715), case sensitivity.** Fixed with case folding. Your exact reproduction, dataset `orders` with dialect `SNOWFLAKE` and `SUM(ORDERS.AMOUNT)`, no longer produces a phantom cross-dataset error naming a dataset that does not exist. It now reports the rule 2 violation instead, which is the accurate diagnosis.

**[`validate.py:233`](https://github.com/apache/ossie/pull/343#discussion_r3880477913), aliases and subqueries.** Implemented the three-way distinction you proposed, cross-checking qualifiers against the model's dataset names. `SUM(o.amount)` and your correlated subquery example are now silent rather than advising a move that would not help. A qualifier matching another dataset is a rule 1 error; one matching the declaring dataset is a rule 2 error; anything else is treated as a local alias, CTE, or subquery source and left alone.

**[`validate.py:298`](https://github.com/apache/ossie/pull/343#discussion_r3880504272), null expression traceback.** Applied `.get("expression") or {}`, and to all three loops including the pre-existing model-level one you flagged. `metrics: [{name: total, expression: }]` now surfaces the schema error instead of an `AttributeError`.

**[`ossie-schema.json:218`](https://github.com/apache/ossie/pull/343#discussion_r3880389834), Pydantic silently dropping metrics.** Confirmed and fixed. Before: loading the example yielded only `customer_lifetime_value` and `store_productivity`. After: all five survive, and a load/serialize/reload round-trip is stable and schema-valid.

Your diagnosis of *why* nothing caught it was the more valuable part. Added a test walking every `$defs` entry with `properties` and asserting each is representable on the corresponding model, comparing against aliases too since `from` is a keyword. I then verified the test is not vacuous by simulating the pre-fix model and confirming it reports `['metrics']`.

One note on how the fix landed: adding the field needed `OssieMetric` visible to `OssieDataset`, and the file uses no forward references anywhere, so I moved `OssieMetric` above `OssieDataset` rather than introduce the first one. That is why `models.py` shows ~30 changed lines for one new field. Happy to switch to `Optional[list["OssieMetric"]]` if you would prefer the smaller diff.

**[`spec.md:40`](https://github.com/apache/ossie/pull/343#discussion_r3880528900), TOC level.** Promoted `Metric Scoping` to `##`. TOC entry 7 now resolves to a top-level section.

**[`spec.md:134`](https://github.com/apache/ossie/pull/343#discussion_r3880518395), stale docs.** All five fixed: `docs/index.md` at the component table, the FAQ, and the glossary; `converters/README.md` at the metrics section and the dataset property table, which now lists `metrics` and warns that reading only `semantic_model.metrics` drops them.

**[`spec.md:442`](https://github.com/apache/ossie/pull/343#discussion_r3880515498), metric/field name collision.** Confirmed, and it was accepted by the validator exactly as you showed. Added as rule 4, a hard error, with a check. @christianeu-db raised the same thing independently. See the warn-vs-error section below for why this one is hard while the model-level clash is not.

**[`spec.md:485`](https://github.com/apache/ossie/pull/343#discussion_r3880525839), example undercut its own point.** Right. Added `order_id` to the `orders` fields, which also lets the example demonstrate the fields-not-source-columns rule properly.

**[`spec.md:713`](https://github.com/apache/ossie/pull/343#discussion_r3880522553), the two examples taught opposite things.** Moved `customer_count` onto `customers.metrics` as `COUNT(DISTINCT id)` and scoped the model-level comment to `revenue_per_customer` alone.

**[`spec.md:539`](https://github.com/apache/ossie/pull/343#discussion_r3880397593), all seven converters, and the lossy-conversion wording.** Took your stronger wording: the spec now says a consumer reading only `semantic_model.metrics` produces an incomplete representation, that this is a lossy conversion rather than a valid one, that it SHOULD warn naming the dropped metrics, and MUST NOT present the result as faithful.

The `iter_metrics(model)` helper is a good idea and I have not done it, to keep this PR reviewable. Happy to open it as a follow-up.

**[`tpcds_semantic_model.yaml:155`](https://github.com/apache/ossie/pull/343#discussion_r3880362878), broken omni tests.** Reproduced: 2 failed, 81 passed. I went the other way from your preference and reverted the example, so it is byte-identical to `main` and no converter is touched. All seven suites now pass.

The reasoning is your own point about the flagship model being lossy through the whole hub-and-spoke on day one. Rather than fix one converter and leave six lossy, the example stays as-is until converters support the placement, which makes converter work a prerequisite for moving the example rather than a race against it. spec.md carries the worked examples meanwhile. If you would still rather see omni updated here, say so and I will do it.

Unrelated, found while running the suites: `converters/wisdom` has no committed `uv.lock` while every other converter does.

---

## @christianeu-db

**[`spec.md:449`](https://github.com/apache/ossie/pull/343#discussion_r3876543659), single-level namespace.** Adopted as rule 2, with a validator check and rewritten examples throughout. This is the largest change in the revision.

It also simplified the validator, which no longer compares a qualifier against the declaring dataset's name to decide whether a reference is local.

One correction it forced: the first revision claimed hoisting to a flat namespace needed no expression rewriting. It now does, and the consumer guidance says so.

**[`spec.md:440`](https://github.com/apache/ossie/pull/343#discussion_r3876450874), fields, source columns, or both.** Fields, specifically. Rule 1 said "fields" but only implied it, so that is now stated outright: a column used by a metric must be declared as a field. Not machine-enforced, because an expression also contains function names, literals, and struct or variant paths. See question 3 below.

**[`spec.md:441`](https://github.com/apache/ossie/pull/343#discussion_r3876459026), uniqueness across fields and metrics.** Yes, added as rule 4, a hard error. Your framing was the deciding one: uniqueness is what guarantees `dataset.name` refers to exactly one thing.

**[`spec.md:442`](https://github.com/apache/ossie/pull/343#discussion_r3876689763), unqualified names for model-level only.** Adopted. Rule 6 is now: unqualified resolves to a model-scoped metric, dataset-scoped is always `dataset.metric`. Rule 3 stays, so two datasets may each declare the same local name.

This had a knock-on effect. The old rule forbidding a dataset-scoped name from colliding with a model-scoped one existed *only* to keep unqualified references unambiguous. Under rule 6 that justification disappears, so the constraint had to be re-argued, which is how it became a warning.

The spec's prior-art table now records reference form per system, so the Tableau comparison is visible.

**[`spec.md:447`](https://github.com/apache/ossie/pull/343#discussion_r3877764369), layering terminology.** Added a note that presenting many datasets and their metrics through one queryable interface belongs to the layer above. I have not hard-coded "dimensional" or "presentation" while the working group is choosing between them; happy to use whichever once settled.

**[`spec.md:453`](https://github.com/apache/ossie/pull/343#discussion_r3877795697), aggregations need not be simple.** Reworded. The spec now says rule 1 places no limit on the complexity of the aggregation, and that any expression resolving within the declaring dataset is eligible.

**[`spec.md:517`](https://github.com/apache/ossie/pull/343#discussion_r3877669350), metric views detail.** Rewrote that entry with your description: one `source` plus optional `joins`, sources named so a column is `source_name.column_name`, with the view itself exposing a flat schema. The spec now notes it as the closest analogue to the namespace rules here.

**[`ossie-schema.json:223`](https://github.com/apache/ossie/pull/343#discussion_r3876575945), grain without a PK, and the expression boundary.** Took both points. There is now a grain paragraph stating that a dataset-scoped metric aggregates at its dataset's grain, and that this does not depend on a declared `primary_key`. `primary_key` matters for reasoning about fan-out but is not a prerequisite.

On the expression boundary, your layering rationale is stronger than the one I had, so I added it: letting an expression reach through relationships makes it depend on the shape of the relationship graph, which would require a dataset to reason about the datasets around it. The schema description carries the grain point too.

**[`spec.yaml:263`](https://github.com/apache/ossie/pull/343#discussion_r3876597056), metric references as a separate proposal.** Agreed, and not attempted here. The spec.yaml comment says so rather than leaving a silent gap, and the PR body records the direction rules that follow from rule 6 so the two proposals stay compatible. The same question is live in #342.

**Your review summary** also asked what can be an input to a dataset-scoped metric. Two halves: fields rather than source columns, answered above; and whether another metric can be an input, deferred as agreed.

---

## @khush-bhatia

**[`ossie-schema.json:223`](https://github.com/apache/ossie/pull/343#discussion_r3876181384), grain, and why the expression boundary is where it is.**

On grain: added, with @christianeu-db's qualification that it must be computable without requiring a PK. So the spec states the grain is that of the rows the `source` produces, and does not tie it to `primary_key`.

On the expression boundary: I had this framed badly in the first revision, and the wording is now fixed throughout. A dataset-scoped metric simply aggregates data held by its dataset. It is then joined and grouped exactly like any other metric, using the relationships declared in the model, so it can be sliced by dimensions of any dataset the model connects. The placement records where an aggregation is anchored; it does not fence off part of the model from the metric, and nothing that Ossie models express today becomes unexpressible.

What rule 1 constrains is only the *expression*: it must aggregate fields of its own dataset rather than combine fields from several. An expression that genuinely needs to combine columns across datasets, such as `SUM(orders.amount) / COUNT(DISTINCT customers.id)`, is what `semantic_model.metrics` is for.

As for why the boundary sits there, the honest answer is that it is the conservative choice rather than the obviously correct one. It makes the anchor point checkable, so a dataset and its metrics can be reasoned about or exchanged without resolving the join graph. And the asymmetry matters: relaxing this later is backward compatible, tightening it is not.

Snowflake semantic views do let a table-scoped metric's own expression reach through a relationship, so the boundary between their two placements is softer than the one proposed here. If you think that is the better model, this is the place to push, and the spec records it as an open question rather than settling it.

One thing I should separate out, since I previously ran these together: choosing between multiple relationship paths when more than one connects the same pair of datasets is a different problem entirely. Ossie has no mechanism for it, it affects model-scoped metric expressions and consumer queries regardless of this PR, and nothing here addresses it. I raised it separately in #342.

**[`spec.yaml:263`](https://github.com/apache/ossie/pull/343#discussion_r3876192153), combining dataset-scoped metrics at model level.** This is the metric-references-metric question, and @christianeu-db is right that it belongs in its own proposal since it changes the expression language rather than metric placement.

Not silent about it, though. The spec.yaml comment now states that composing model-scoped metrics from dataset-scoped ones is a frequently requested capability being handled separately, and the PR body records the direction rules so whichever proposal lands second does not have to reopen this one. Worth noting the gap is ambiguous rather than absent: `SUM(orders.amount) / total_orders` parses today with `total_orders` read as an unqualified column, so a producer could write it and consumers could disagree on its meaning. That is an argument for settling it explicitly rather than leaving it implicit.

---

## Warn vs error, and where I would like feedback

Review made clear that not every constraint deserves the same severity, so there are now three levels. The full table is in the PR body; the short version and the open questions:

**Hard error** where one string resolves two ways, or the expression is not self-contained: duplicate metric name in a dataset; metric name equal to a field name of the same dataset (rule 4); expression referencing another dataset (rule 1); expression qualifying a field with its own dataset name (rule 2).

**Warning, exit 0** for one case: a model-scoped metric reusing a dataset-scoped metric's name (rule 5). References stay unambiguous, and a dataset may be authored independently and reused across models, so it must not become invalid because of a name the surrounding model introduces. The warning is attributed to the **model**, and names the model-scoped metric as the thing to rename. The first revision reported it against the dataset, which asks the wrong party to change.

**Deliberately unchecked**, now stated in spec.md rather than left silent: a model-scoped metric name equal to a field name, or to a dataset name; whether a bare name is a declared field; and qualifiers that are local aliases or CTEs.

The rule 4 versus rule 5 split is the part I most want checked. The line I drew is that rule 4 prevents one string resolving two ways, while rule 5 guards against confusion between names that remain separately addressable. Four specific questions:

1. **Is rule 2 right as a hard error?** `SUM(orders.amount)` inside dataset `orders` is valid, unambiguous SQL, so erroring enforces a spelling. The case for erroring is that it breaks dataset portability; the case for warning is that nothing is actually wrong.
2. **Is rule 5 right as a warning?** Should shadowing block validation instead?
3. **Should the undeclared-column case be enforced?** Rule 1 says fields, but checking it needs heuristics to avoid false positives on functions and struct paths.
4. **Should model-metric-vs-field warn?** Currently silent. I judged a warning would be noise, since a model-scoped metric usually takes the name of the column it aggregates, and noisy warnings would devalue rule 5's. Reasonable to disagree.

---

## Verification

The reproductions are now committed. #330 landed a validator test suite while this branch was in review, so these extend it rather than introduce one:

- **`validation/tests/test_validate.py`**: 28 tests, the 10 from #330 plus 18 added here, one per defect reported above plus the rule behaviours. Confirmed non-vacuous by reverting only `validate.py` to the pre-review commit, where **10 of the 18 new cases fail** and all 10 pre-existing ones still pass.
- **`python`**: 10 tests, including the schema/model drift test, which likewise fails against a simulated pre-fix model.
- **9 Python converter suites pass**: databricks, dbt, gooddata, honeydew, nvidia, omni, orionbelt, snowflake, wisdom. I had previously said "seven", which was wrong; my sweep used a hardcoded list and had skipped dbt and gooddata. It now auto-discovers.
- **2 Java converters not run**: polaris and salesforce need a JRE unavailable on this machine.
- `examples/tpcds_semantic_model.yaml` validates against the core schema; `examples/flights.yaml` against the ontology schema.
- All 24 fenced YAML blocks in spec.md parse, and the Complete Example validates against the schema.
- The two snippets spec.md marks INVALID are actually rejected by the validator, so the documented rules and the tooling cannot drift apart.

Four things found while verifying that are unrelated to this change and **not addressed here**. Each looks worth its own issue, and I am happy to open them:

- **No CI workflow covers the paths this PR touches.** Every workflow is path-filtered to `cli/**` or `converters/<name>/**`, so nothing watches `core-spec/`, `validation/`, `python/`, `docs/` or `examples/`. A green check on this PR would mean no jobs ran. This affects #330's new validator suite too, which no workflow executes. I have deliberately not added a workflow here so this stays a specification change.
- **`converters/databricks` has a pre-existing test failure** in `test_roundtrip_properties.py::test_mv_to_ossie_to_mv`. Hypothesis generates a metric view whose `source` table and a join share the name `j0`, tripping the converter's duplicate-name guard. Confirmed reproducible on unmodified `main`.
- **`converters/polaris` has the same latent gap the Pydantic model had**: its `Dataset` class has no `metrics` field and its parser reads model-level metrics only. Latent because no model in the repo currently declares dataset-scoped metrics.
- **`converters/wisdom` has no CI workflow and no committed `uv.lock`**, unlike every other converter.

One process note for anyone re-running the validator tests: they must run under `uv`. Bare `python3` has no sqlglot, so `_parse_expression` returns "skip" and the scoping checks silently pass. My first attempt produced four false passes that way, which is why `test_validate.py` opens by asserting sqlglot is importable.

core-spec/spec.md
jklahr: Proposal for data set metrics -- SUM(col_name) references a source column. SUM(dataset.field_name) can reference a field in the same data set.

core-spec/spec.md
jklahr: yes

khush-bhatia (approved):

core-spec/spec.md
christianeu-db: Having differentiation makes a lot sense. Let's chat at the working group as well. This one is quite tricky to get right. For metric views, we have:

1. source.field: from the table
2. field: a dimension (if it exists), otherwise it's from source

The wrinkle is that you could have a struct-typed column called source so that source.field is ambiguous. This been a tricky area for development.

To be honest, 1 pager on the name resolution precedence rules (e.g. source, dataset, struct, model) could be valuable in of itself.

core-spec/spec.md
```diff
@@ -417,6 +426,133 @@ expression:
...
+
+| | Model-scoped (`semantic_model.metrics`) | Dataset-scoped (`datasets[].metrics`) |
+|---|---|---|
+| Expression references | Fields of any dataset in the model, qualified: `dataset.field` | Fields of its own dataset, qualified: `dataset.field`; columns of its `source`, unqualified: `column` |
+| Name uniqueness | Unique across the semantic model | Unique within its dataset, and distinct from that dataset's field names |
+| Referenced as | `metric_name` | `dataset_name.metric_name` |
+
+**Rules**
+
+1. A dataset-scoped metric's expression MAY reference the declared fields of its dataset and the columns of its `source`. A declared field is written `dataset_name.field_name`; a column of the `source` is written unqualified: `SUM(orders.net_amount) / COUNT(DISTINCT tax_id)`. A qualified reference MUST name a declared field of the declaring dataset. An expression that references another dataset MUST be declared in `semantic_model.metrics`.
```
justint-db: It is somewhat clunky that a declared field is written `dataset_name.field_name`. It introduces some circularity that makes programmatic handling difficult: a dataset is defined by its list of fields and metrics, but those are themselves defined in terms of the dataset.

An alternative that is semantically simpler would be:
- declared fields of the dataset are referenced unqualified.
- columns from the source are referenced either (1) unqualified (if there is no name conflict), or (2) qualified by a `source_name` to disambiguate. (The `source_name` could be fixed, e.g. always `"source"`, or could be something we allow model authors to specify.

This approach would avoid the naming circularity.

core-spec/spec.md
```diff
@@ -417,6 +426,133 @@ expression:
...
+2. Dataset-scoped metric names MUST be unique within their dataset. Two datasets MAY each declare a metric with the same name.
+3. A dataset-scoped metric name MUST NOT collide with the name of a field of the same dataset, which under rule 5 would leave `orders.amount` ambiguous.
+4. A model-scoped metric SHOULD NOT reuse the name of a dataset-scoped metric in the same semantic model. Validators SHOULD warn, and SHOULD attribute the warning to the model rather than to the dataset.
```
justint-db: While I think starting with this restriction is fine; we can always relax it later, it doesn't seem strictly necessary since (5) below says that model-scoped and dataset-scoped names are qualified differently anyway, so can't practically conflict.

core-spec/spec.md
```diff
@@ -417,6 +426,133 @@ expression:
...
+5. An unqualified metric reference resolves to a model-scoped metric. A dataset-scoped metric is referenced as `dataset_name.metric_name`.
```
justint-db: It is unclear where these references appear.

Can a reference to a model-scoped metric appear in a dataset-scoped definition?

It would be useful to list out explicitly:
1. what can a dataset-scoped field definition refer to, using what name resolution rules?
2. what can a dataset-scoped metric definition refer to, using what name resolution rules?
3. what can a model-scoped metric definition refer to, using what name resolution rules?
4. what can Layer 2 or Layer 3 query languages over a model refer to?

For the purposes of this doc, we probably only need to define (1) and (2). The others can live in docs that explicitly about those other parts of the model.

justint-db (changes requested):

PR #347, opened 2026-08-29
fix(gooddata): preserve grain for aliased primary-key fields

### LINKED ISSUE #346: GoodData export drops grain for aliased primary-key fields
## Summary

The Ossie-to-GoodData converter silently drops a dataset's `grain` when an Ossie field name differs from the physical primary-key column named by its ANSI SQL expression.

Ossie defines `primary_key` as physical column names. The converter correctly derives each GoodData attribute's `sourceColumn` from the ANSI SQL expression, but it decides grain membership by comparing the logical field name with the physical primary-key columns.

## Minimal reproduction

From `converters/gooddata`:

```bash
uv run python - <<'PY'
from ossie_gooddata import ossie_to_gooddata

document = {
    "version": "0.2.0.dev0",
    "semantic_model": [{
        "name": "m",
        "datasets": [{
            "name": "customers",
            "source": "db.s.customers",
            "primary_key": ["customer_id"],
            "fields": [{
                "name": "customer_key",
                "expression": {
                    "dialects": [{
                        "dialect": "ANSI_SQL",
                        "expression": "customer_id",
                    }],
                },
                "dimension": {},
            }],
        }],
    }],
}

dataset = ossie_to_gooddata(document).ldm.datasets[0]
print(dataset.attributes[0].source_column)
print([grain.id for grain in dataset.grain])
PY
```

Actual output:

```text
customer_id
[]
```

Changing only the field name to `customer_id` produces the expected grain entry, so the failure is specific to valid logical-name/physical-column aliases.

## Expected behavior

The GoodData dataset should contain:

```text
attr.customers.customer_key
```

in its grain because that attribute represents physical primary-key column `customer_id`.

## Impact

The generated GoodData LDM loses the dataset's declared row grain even though the source Ossie model supplies a primary key. Consumers therefore receive a model without the intended uniqueness/grain semantics.

## Root cause

`_convert_ossie_dataset` compares `field_name` with `pk_columns`, while `_convert_to_attribute` separately resolves the physical source column through `_get_source_column(field_def)`.

## Proposed fix

Compare the converted attribute's `source_column` with `pk_columns` when deciding whether to add its attribute ID to `grain_ids`. Preserve the existing behavior when logical and physical names are identical.

Add a focused regression test next to `test_grain_from_primary_key` covering an aliased ANSI SQL source column. No dependency, specification, or public API changes are needed.

## Validation

```bash
cd converters/gooddata
uv run pytest
```

mattfaltyn:
## Summary

- Preserve GoodData dataset grain when an Ossie attribute uses a logical field name that differs from its physical primary-key column.
- Match `primary_key` entries against the converted attribute's `source_column`, which is already derived from the ANSI SQL expression.
- Cover both explicit dimensions and MAQL-detected attributes with regression cases.

This keeps the existing behavior for fields whose logical and physical names match while preventing silent loss of key/grain semantics for aliased fields.

## Related Issues

Closes #346

## Validation

```text
Python 3.12: uv run --python 3.12 --frozen pytest -q  -> 76 passed
Python 3.13: uv run pytest -q                         -> 76 passed
Python 3.14: uv run --python 3.14 --frozen pytest -q  -> 76 passed
uv run ruff check src tests                           -> passed
git diff --check                                      -> passed
```

### Specification
- [x] No specification changes

### Converters
- [x] Regression coverage is included under the GoodData test directory

### Documentation
- [x] Existing documentation was reviewed; no update is needed because this restores the documented primary-key-to-grain mapping without changing the interface

### Tests
- [x] All GoodData tests pass across the CI Python matrix
- [x] New behavior is covered by tests

### Compliance
- [x] No new source files or dependencies

### DIFF
--- a/converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
+++ b/converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
@@ -153,13 +153,12 @@ def _convert_ossie_dataset(
     pk_columns = set(ds.get("primary_key", []))
 
     for field_def in fields:
-        field_name = field_def["name"]
         is_dimension = field_def.get("dimension") is not None
 
         if is_dimension:
             attr = _convert_to_attribute(field_def, ds_name)
             attributes.append(attr)
-            if field_name in pk_columns:
+            if attr.source_column in pk_columns:
                 grain_ids.append(attr.id)
         else:
             # Check MAQL expression to determine if fact or attribute
@@ -167,7 +166,7 @@ def _convert_ossie_dataset(
             if maql_type == "attribute":
                 attr = _convert_to_attribute(field_def, ds_name)
                 attributes.append(attr)
-                if field_name in pk_columns:
+                if attr.source_column in pk_columns:
                     grain_ids.append(attr.id)
             else:
                 facts.append(_convert_to_fact(field_def, ds_name))
--- a/converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
+++ b/converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
@@ -91,6 +91,10 @@ def _build_target_info(sm: dict[str, Any]) -> dict[str, dict[str, Any]]:
         if not is_date:
             for f in ds.get("fields", []):
                 src = _get_source_column(f)
+                if src in col_to_attr:
+                    raise ValueError(
+                        f"Dataset '{ds_name}': source column '{src}' maps to multiple fields."
+                    )
                 col_to_attr[src] = f"attr.{ds_name}.{f['name']}"
         info[ds_name] = {"is_date": is_date, "col_to_attr": col_to_attr}
     return info
@@ -148,9 +152,6 @@ def _convert_ossie_dataset(
     # Regular dataset
     attributes: list[GdAttribute] = []
     facts: list[GdFact] = []
-    grain_ids: list[str] = []
-
-    pk_columns = set(ds.get("primary_key", []))
 
     for field_def in fields:
         is_dimension = field_def.get("dimension") is not None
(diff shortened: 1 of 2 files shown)


converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
```diff
@@ -153,21 +153,20 @@ def _convert_ossie_dataset(
...
 
     for field_def in fields:
-        field_name = field_def["name"]
         is_dimension = field_def.get("dimension") is not None
 
         if is_dimension:
             attr = _convert_to_attribute(field_def, ds_name)
             attributes.append(attr)
-            if field_name in pk_columns:
+            if attr.source_column in pk_columns:
```
jbonofre: This matches PK columns by comparing `attr.source_column` against `pk_columns`, which is a string/value match rather than a unique field match. If two fields alias the same physical column (e.g. a natural key `customer_id` and a display alias `customer_key` that both resolve to `ANSI_SQL: customer_id)`, both get added to `grain_ids`, producing a bogus multi-attribute composite grain for what's actually a single-column key.

I suggest keyring grain membership off the field's identity (e.g. matching source columns to specific field defs and taking the first match, or building grain from `primary_key` entries directly rather than scanning all attributes) instead of plain "is this source_column in pk_columns" check.

converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
jbonofre, line 98: I found it related while reviewing 😄 

`col_to_attr` is a dict keyed by resolved `source_column`, with no collision check. If two fields resolve to the same physical column, whichever is processed last silently overwrites the earlier entry. Since `_convert_relationship` looks up target attributes through this map, a relationship meant to reference one attribute can get silently rewired to a different one (no error, just a wrong `GdReferenceTarget` in the output).

Given `_convert_relationship` already raises `ValueError` for an unresolved column, this should probably raise (or at least warn) on a `source_column` collision here too, rather than overwriting silently.

jbonofre (changes requested):

converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
mattfaltyn: > building grain from primary_key entries directly rather than scanning all attributes

Great catch. Grain now follows primary_key entries directly, so each key column resolves to one attribute.

converters/gooddata/src/ossie_gooddata/ossie_to_gooddata.py
mattfaltyn: > this should probably raise on a source_column collision here too

Agreed. The converter now raises ValueError on duplicate resolved source columns, with regression coverage for ambiguous grain and relationship targets.

PR #349, opened 2026-08-31
fix(cli): reject invalid arguments on plugin, plugin install, and convert

### LINKED ISSUE #345: cli: argument validation gaps remain after #334 (plugin, plugin install, convert)
### Description

#334 / #335 fixed `ossie plugin list`, but the same "silently ignore positional arguments and exit 0" defect remains in three other places, including one level up on the `plugin` parent command itself.

The parent case is the most likely to be hit, since a typo'd subcommand is more common than a stray argument to a valid one, and it is easy to assume it was covered by #334.

**Important:** the one-line fix from #335 does *not* work on the parent command. See "Cause" below before attempting it.

### Steps to reproduce

Against `main` with #335 merged:

```bash
ossie plugin bogus;               echo $?   # unknown subcommand
ossie plugin help list;           echo $?   # wrong help, still succeeds
ossie plugin install a b c;       echo $?   # extra positional args ignored
ossie plugin install;             echo $?   # no name and no --all
ossie convert --from x --input y extra; echo $?   # trailing arg ignored
```

### Actual behavior

All five exit `0`. `ossie plugin bogus` prints the `plugin` help to STDOUT and reports success; the rest print `not yet implemented` and report success, discarding the arguments.

### Expected behavior

Each should reject the invocation with a non-zero exit status and show usage, consistent with `plugin list` after #335 and with `plugin remove`.

### Cause

Two different causes, which is why one fix does not cover all of them:

**`plugin install` (`cli/cmd/plugin/install.go:25`) and `convert` (`cli/cmd/convert.go:25`)** simply declare no `Args` validator. Cobra treats a nil validator as accepting arbitrary arguments — the same root cause as #334.

**`plugin` (`cli/cmd/plugin/plugin.go:23`) is different.** Adding `Args: cobra.NoArgs` to `Cmd` has no effect, verified by building it: because the parent has no `Run`/`RunE`, cobra's `execute()` returns `flag.ErrHelp` at the `if !c.Runnable()` check (`command.go:954`) and never reaches `c.ValidateArgs` (`command.go:968`). The `Args` field is dead code on a non-runnable command.

### Suggested fix

- `plugin install`: `Args: cobra.MaximumNArgs(1)`, plus a check that exactly one of the plugin name or `--all` is supplied.
- `convert`: `Args: cobra.NoArgs` (it is flag-only).
- `plugin`: give it a `RunE` that prints help and returns an error, so an unknown or missing subcommand exits non-zero. `Args: cobra.NoArgs` alone will not do it.

To keep this from recurring as commands are added, consider one table-driven test that walks `rootCmd.Commands()` recursively and asserts every command either declares a non-nil `Args` or is a parent with an explicit help-and-error `RunE`. That would have caught all three of these in a single pass.

Note that a test asserting this by calling `cmd.ValidateArgs(...)` directly will pass on the `plugin` parent while the real CLI still exits 0. These need to be exercised through `rootCmd.SetArgs(...)` / `Execute()` with captured output.

### Related, likely a separate issue

The unimplemented stubs print `not yet implemented` and `return nil`, so they exit 0 while doing nothing: `convert` (`convert.go:46`), `validate` (`validate.go:38`), `plugin install` (`install.go:36`), `plugin remove` (`remove.go:33`). A pipeline running `ossie validate model.yaml && deploy` would deploy an unvalidated model. These should write to `cmd.ErrOrStderr()` and return an error until implemented. Happy to split this out if preferred.

### Environment

- Apache Ossie: PR #335 branch `fix/plugin-list-reject-args` @ `5947c69`
- Go: `go1.27.0 darwin/arm64`
- OS: macOS arm64

kayemkim:
## Summary

All five reproductions from #345 now exit non-zero with usage shown:

```
$ ossie plugin bogus
Error: unknown command "bogus" for "ossie plugin"
$ ossie plugin help list
Error: unknown command "help" for "ossie plugin"
$ ossie plugin install a b c
Error: accepts at most 1 arg(s), received 3
$ ossie plugin install
Error: requires a plugin name or --all
$ ossie convert --from x --input y extra
Error: unknown command "extra" for "ossie convert"
```

As the issue's cause analysis lays out, this took two different fixes:

- `plugin install` and `convert` simply lacked an `Args` validator. `install` now takes at most one positional argument and requires exactly one of the plugin name or `--all` (so `install --all foo` is also rejected); `convert` is flag-only and takes none.
- The `plugin` parent had to become runnable first: without a `Run`/`RunE`, cobra bails out with `flag.ErrHelp` before reaching `ValidateArgs`, so `Args` alone is dead code there. It now has a `RunE` that returns an error for a bare `ossie plugin`, and once the command is runnable, `NoArgs` produces the "unknown command" error for a typo'd subcommand.

Tests follow the issue's suggestion on both points. They drive the real entry point through `rootCmd.SetArgs()` / `Execute()`, since a direct `ValidateArgs` call passes on the non-runnable parent while the CLI still exits 0; covered are the five reproductions, the valid invocations, and the help paths (`--help`, `-h`, and `help plugin` all still exit 0). A second test walks `rootCmd.Commands()` recursively and fails if any command lacks an `Args` validator or any parent is not runnable, so the next command added without validation fails the suite.

The "stubs print not yet implemented and exit 0" part at the end of the issue is left out here, since it's marked there as likely a separate issue.

## Related Issues

Fixes #345.

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance

### DIFF
--- a/cli/cmd/args_test.go
+++ b/cli/cmd/args_test.go
@@ -1,0 +1,106 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// execute drives the CLI through rootCmd.Execute. Calling ValidateArgs
+// directly is not enough: it passes on the non-runnable plugin parent
+// while the real CLI exits 0 (#345).
+func execute(t *testing.T, args ...string) error {
+	t.Helper()
+	out := &bytes.Buffer{}
+	rootCmd.SetOut(out)
+	rootCmd.SetErr(out)
+	rootCmd.SetArgs(args)
+	err := rootCmd.Execute()
+	resetFlags(rootCmd)
+	return err
+}
+
+// Flag values persist across Execute calls, so reset them between cases.
+func resetFlags(cmd *cobra.Command) {
+	cmd.Flags().Visit(func(f *pflag.Flag) {
+		_ = f.Value.Set(f.DefValue)
+		f.Changed = false
+	})
+	for _, sub := range cmd.Commands() {
+		resetFlags(sub)
+	}
+}
+
+func TestArgumentValidation(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+
+	tests := []struct {
+		name    string
+		args    []string
+		wantErr bool
+	}{
+		// The five reproductions from #345.
+		{name: "plugin unknown subcommand", args: []string{"plugin", "bogus"}, wantErr: true},
+		{name: "plugin help routed as args", args: []string{"plugin", "help", "list"}, wantErr: true},
+		{name: "plugin install extra args", args: []string{"plugin", "install", "a", "b", "c"}, wantErr: true},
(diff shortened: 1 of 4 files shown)


QMalcolm (approved):
Looks good to me, thank you for doing this work !

kayemkim:
@QMalcolm friendly ping, when you get a chance. This still merges clean against current main and the CI run is green.

jbonofre:
@kayemkim I'm doing a quick pass and I will merge it. Thanks for the reminder!

jbonofre (approved):
LGTM thanks!

PR #350, opened 2026-08-31
docs(spec): align spec.yaml ai_context with the schema at every node

### LINKED ISSUE #340: core-spec/spec.yaml omits `ai_context` on Relationship, while the JSON schema defines it
## Summary

`Relationship` is the only node where the two core-spec files disagree on whether a field exists at all:

| Source | `Relationship` properties |
|---|---|
| `core-spec/osi-schema.json` (`$defs/Relationship`) | `name`, `from`, `to`, `from_columns`, `to_columns`, **`ai_context`**, `custom_extensions` |
| `core-spec/spec.yaml` (`relationships:`) | `name`, `from`, `to`, `from_columns`, `to_columns`, `custom_extensions` |

Every other node (`semantic_model`, `datasets`, `fields`, `metrics`) carries `ai_context` in both files.

## Why it matters

Implementations that vendor `spec.yaml` — rather than the JSON schema — as their prompt-facing or documentation contract never learn that a relationship can be annotated. A join whose meaning is not obvious from column names (a bridge table, a type-2 dimension, a deliberately filtered relationship) has no place to say so, even though validators accept it.

## Suggested fix

Add `ai_context` to the `relationships:` block in `spec.yaml`, matching how it appears under `datasets:`.

## Related

This is the third reported drift between the two files:

- #141 — `spec.yaml` says `ai_context: string`; the JSON schema and the canonical example use a structured object
- #322 — the document root carries neither `ai_context` nor `custom_extensions`

All three share a root cause: `spec.yaml` and `osi-schema.json` are maintained by hand with nothing checking that they agree. A CI check asserting that the property set of each node matches between the two files would catch this class of drift, and would likely surface more than these three.

kayemkim:
## Summary

`spec.yaml` understated `ai_context` in two different ways. On `Relationship` the field was missing outright, even though the JSON schema's `$defs/Relationship` carries it and the spec.md property table lists it. On the other four nodes it was written as `ai_context: string`, while the schema defines AIContext as string-or-object with `additionalProperties: true`, spec.md documents it as `string/object` at all five nodes, and the canonical TPC-DS example uses the object form.

The notation is the one suggested in #141 and asked for during the review of #221: an open map as the value, the two accepted forms spelled out once where the field first appears, and a pointer to AIContext in the JSON schema at each site. Nothing normative changes. The JSON schema is untouched and already accepts both forms, as do the Python models.

The first commit adds the field to `Relationship`, the second aligns the remaining four nodes. `spec.yaml` still parses.

## Related Issues

Fixes #340. Also covers #141, which asked for exactly this alignment; #221 proposed it as well but has had changes requested since July without an update.

### DIFF
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -171,6 +171,10 @@ relationships:
     #   - [order_id, line_number]    # Composite key
     to_columns: []  # Array of column names
 
+    # Optional: Additional context for AI tools (e.g., synonyms, business context)
+    # Helps LLMs understand when and why the datasets should be joined
+    ai_context: string
+
     # Optional: Vendor-specific attributes for extensibility
     custom_extensions:
       - vendor_name: string  # Free-form string identifying the vendor
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -173,6 +173,11 @@ relationships:
 
     # Optional: Additional context for AI tools (e.g., synonyms, business context)
     # Helps LLMs understand when and why the datasets should be joined
+    # Either a free-form string, or a structured object:
+    # ai_context:
+    #   instructions: string  # Instructions for AI on how to use this entity
+    #   synonyms: []          # Alternative names and terms
+    #   examples: []          # Sample questions or use cases
     ai_context: string
 
     # Optional: Vendor-specific attributes for extensibility
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -67,7 +67,14 @@ semantic_model:
     description: string
 
     # Optional: Additional context for AI tools (e.g., custom prompts, instructions)
-    ai_context: string
+    # Can be either:
+    #   - A free-form string, or
+    #   - A structured object with optional keys:
+    #       instructions: string  # how AI should use this entity
+    #       synonyms:     []      # alternative names / terms
+    #       examples:     []      # sample questions or use cases
+    #     (additionalProperties: true, so vendors may add more keys)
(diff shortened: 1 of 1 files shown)


core-spec/spec.yaml
```diff
@@ -171,6 +171,10 @@ relationships:
     #   - [order_id, line_number]    # Composite key
     to_columns: []  # Array of column names
 
+    # Optional: Additional context for AI tools (e.g., synonyms, business context)
+    # Helps LLMs understand when and why the datasets should be joined
+    ai_context: string
```
jbonofre: The schema, `spec.md`, the Python models, and the TPC-DS example all treat relationship `ai_context` as string-or-object. This PR fixes "field is missing" and introduces "field has the wrong type" (corresponding to #141). 

I think it's worth changing to reflect both forms before merge, since `spec.yaml` handles nested shapes elsewhere (`expressions`, `dimensions`).

jbonofre (changes requested):

core-spec/spec.yaml
kayemkim: Agreed. Pushed a change that documents the object form (instructions/synonyms/examples) in a comment above the field, the way primary_key and unique_keys show their alternate shapes. I kept the live value as `ai_context: string` to stay in line with the sibling nodes, since reworking the string-only notation across the file is what #141 tracks; or I can switch the live value to the nested form if you'd rather lead with it here.

jbonofre (approved):
Good one! Thanks!

Issue #285, opened 2026-07-29
Proposal: ThoughtSpot converter (bidirectional, TML ↔ Ossie)


djwaldo:
ThoughtSpot would like to contribute a converter between ThoughtSpot's semantic model format (Model TML + Table TML, YAML) and the Ossie semantic model, following the established converter pattern.

## Scope

- **Bidirectional**: `tml_to_ossie` (import) and `ossie_to_tml` (export)
- **File-to-file, Python**, mirroring the `converters/databricks` package shape (fixtures including a TPC-DS pair, roundtrip + property-based tests, per-converter CI workflow)
- **`custom_extensions` with `vendor_name: THOUGHTSPOT`** for lossless roundtrip of ThoughtSpot-only concepts; one row added to the well-known vendor examples table in `core-spec/spec.md`
- **Mapping docs before code**: a construct-mapping document and an expression/function-mapping document (each function in `core-spec/expression_language.md` classified as direct ThoughtSpot formula equivalent, requiring a SQL passthrough function, or unmappable → converter issue) will be posted on this issue for community review **before** the code PR

## Proposed mapping sketch (high level)

| Ossie | ThoughtSpot |
|---|---|
| `datasets` | Model `model_tables` + Table TML (physical `source`) |
| `fields` | table columns / ATTRIBUTE formulas |
| `metrics` | MEASURE columns / formulas |
| `relationships` | Model `joins` |
| `ai_context` | synonyms, Spotter instructions |
| `custom_extensions[THOUGHTSPOT]` | join types, aggregation defaults, formats, other ThoughtSpot-only metadata |

Happy to adjust scope/shape based on feedback — the mapping documents will follow as comments on this issue.

— Damian Waldron (ThoughtSpot); damian.waldron@thoughtspot.com on the dev list.

djwaldo:
Cross-linking #269 — @utkarshparekh opened it on 26 July, three days before this issue, asking the same question. I did not find it before filing this one. Suggesting we **consolidate here**, with #269 closed as the earlier duplicate once @utkarshparekh and the maintainers are happy with that; my full reply to the points raised there is on #269 and I won't repeat it.

[...]

Issue #269, opened 2026-07-26
Evaluate a ThoughtSpot TML converter


utkarshparekh:
Hi all,

I would like to ask whether the community has existing plans for, or would welcome, a converter between Apache Ossie and ThoughtSpot Modeling Language (TML).

## Why this may be useful

ThoughtSpot represents metadata objects as TML in YAML or JSON. Tables, worksheets, and views can carry columns, formulas, joins, descriptions, and modeling properties that overlap with Ossie datasets, fields, metrics, and relationships. ThoughtSpot also exposes supported TML import/export operations through its API, so static TML is a possible interchange boundary.

Official references:

- https://developers.thoughtspot.com/codespot/tml-python-library
- https://developers.thoughtspot.com/
- https://github.com/thoughtspot/thoughtspot_rest_api_v1_python

I searched the open and closed Ossie issues and pull requests and did not find an existing ThoughtSpot or TML converter proposal.

## Possible v1 boundary

If the project considers this integration appropriate, I suggest investigating a dependency-free static converter first:

- Read and write static TML YAML or JSON without requiring access to a ThoughtSpot instance.
- Initially limit the supported TML object types to the smallest set that represents semantic modeling, likely tables plus either worksheets or models/views, based on representative exports.
- Map columns and formulas to Ossie fields, reusable aggregate definitions to metrics where unambiguous, and representable joins to Ossie relationships.
- Preserve TML-specific properties in a versioned `THOUGHTSPOT` custom extension.
- Report unsupported or ambiguous constructs explicitly.
- Validate generated Ossie documents against `core-spec/osi-schema.json`.
- Keep live ThoughtSpot REST API access outside the static converter's runtime scope.

## Licensing question

The official `thoughtspot_tml` Python repository is public but is distributed under a ThoughtSpot Development Tools EULA rather than an OSI-approved open-source license. Its restrictions also require careful review for a format converter. I therefore do **not** propose depending on, copying from, or distributing that library.

Before implementation, we would need to confirm:

1. Whether a converter based only on publicly documented TML and independently written parsing is appropriate for Apache Ossie.
2. Whether representative TML fixtures can be contributed under Apache-compatible terms.
3. Whether any ASF licensing or legal review is required before proceeding.

## Questions for maintainers and contributors

- Is a ThoughtSpot TML converter something the Ossie community would welcome?
- Is anyone already planning or prototyping this integration so that work can be coordinated rather than duplicated?
[...]

PR #351, opened 2026-08-31
Add THOUGHTSPOT to the Dialect enum and SKIP_SQL_VALIDATION

djwaldo:
## Summary

Registers `THOUGHTSPOT` as an expression-language dialect, mirroring exactly what `MDX`, `TABLEAU` and `MAQL` already do.

ThoughtSpot's formula language is not SQL. It has its own syntax and its own column-reference form:

```
concat ( [ORDERS::Region] , ' - ' , [ORDERS::Segment] )
if ( [Revenue] > 0 ) then 'positive' else 'zero or below'
```

Two things follow. `Dialect` is a closed enum, so a `THOUGHTSPOT` entry fails schema validation outright. And `validate_sql` falls through to a default-dialect sqlglot parse for anything not in `SKIP_SQL_VALIDATION`, so such an expression would raise a parse error rather than being left alone — which is precisely why `MDX`, `TABLEAU` and `MAQL` are in that set.

### Evidence

A semantic model carrying ThoughtSpot expressions on both a field and a metric, against `validation/validate.py`:

**Before**
```
Validation FAILED with 2 error(s):
  [Schema] ... fields -> 0 -> expression -> dialects -> 0 -> dialect:
      'THOUGHTSPOT' is not one of ['ANSI_SQL', 'SNOWFLAKE', 'MDX', 'TABLEAU', 'DATABRICKS', 'MAQL', 'BIGQUERY']
  [Schema] ... metrics -> 0 -> expression -> dialects -> 0 -> dialect:
      'THOUGHTSPOT' is not one of [...]
```

**After**
```
Validation PASSED
```

Existing `python/` and `validation/` suites: 19 passed.

### Impact on existing implementations

None. No behaviour changes for any other dialect — the change is additive to a vocabulary. A consumer that does not know `THOUGHTSPOT` treats it as it would any dialect it does not implement.

### Why five files

The dialect vocabulary is stated in five places, and they have to agree:

| File | Change |
|---|---|
| `core-spec/ossie-schema.json` | the `Dialect` enum |
| `core-spec/spec.md` | the dialect table |
| `core-spec/spec.yaml` | the `dialects` list |
| `validation/validate.py` | `DIALECT_MAP` entry (`None`) + `SKIP_SQL_VALIDATION` |
| `python/src/ossie/models.py` | the `OssieDialect` enum |

The last is easy to overlook — without it the JSON schema accepts the dialect while the shared Python models package still rejects it.

The `DIALECT_MAP` entry is strictly redundant, since `SKIP_SQL_VALIDATION` short-circuits before the lookup, but `MDX`/`TABLEAU`/`MAQL` are each in both and I have followed that pattern rather than diverge from it.

### Possibly relevant beyond this dialect

In discussion #342 on shared filters, part of the argument turns on a parse-based join-path resolution strategy being unworkable for any dialect in `SKIP_SQL_VALIDATION`. That set currently has three members; this would make it four. Better known while that design is under discussion than after.

## Related Issues

Context: #285 and #269 (ThoughtSpot converter). Announced on `dev@ossie.apache.org`: [[DISCUSS] ThoughtSpot converter: scope, licensing, and where it should live](https://lists.apache.org/thread/j186q87txcrml5nygjq0zf225oz3vnb6), with a dedicated thread for this change.

Per CONTRIBUTING's specification-change process this is opened alongside the dev@ announcement, for the 7-day discussion window and a `[VOTE]`.

### Specification
- [x] Spec changes have been discussed on the mailing list or in a linked issue
- [x] Breaking changes to the spec are clearly called out in the summary — there are none; the change is additive

### Validation
- [x] Validation rules in `validation/` are updated
- [ ] New validation cases are covered by tests — no existing test references any specific dialect member; happy to add one if reviewers would like

### Tests
- [x] All existing tests pass (19 passed)

### Compliance
- [x] ASF license headers are present on all new source files — no new files

### DIFF
--- a/core-spec/ossie-schema.json
+++ b/core-spec/ossie-schema.json
@@ -23,7 +23,7 @@
   "$defs": {
     "Dialect": {
       "type": "string",
-      "enum": ["ANSI_SQL", "SNOWFLAKE", "MDX", "TABLEAU", "DATABRICKS", "MAQL", "BIGQUERY"],
+      "enum": ["ANSI_SQL", "SNOWFLAKE", "MDX", "TABLEAU", "DATABRICKS", "MAQL", "BIGQUERY", "THOUGHTSPOT"],
       "description": "Supported SQL and expression language dialects"
     },
     "Vendor": {
--- a/core-spec/spec.md
+++ b/core-spec/spec.md
@@ -58,6 +58,7 @@ Supported SQL and expression language dialects for metrics and field definitions
 | `DATABRICKS` | Databricks SQL |
 | `MAQL` | GoodData MAQL (Metric Analysis and Query Language) |
 | `BIGQUERY` | Google BigQuery (GoogleSQL) |
+| `THOUGHTSPOT` | ThoughtSpot formula language |
 
 ### Data types
 
--- a/core-spec/spec.yaml
+++ b/core-spec/spec.yaml
@@ -37,6 +37,7 @@ dialects:
   - "DATABRICKS"            # Databricks SQL
   - "MAQL"                  # GoodData MAQL (Multi-Dimensional Analytical Query Language)
   - "BIGQUERY"              # Google BigQuery GoogleSQL
+  - "THOUGHTSPOT"           # ThoughtSpot formula language (not SQL)
 
 # Supported logical data types for fields and metrics
 # TODO: Generate this list from the authoritative DataType enum in
--- a/python/src/ossie/models.py
+++ b/python/src/ossie/models.py
@@ -32,6 +32,7 @@ class OssieDialect(str, Enum):
     TABLEAU = "TABLEAU"
     DATABRICKS = "DATABRICKS"
     BIGQUERY = "BIGQUERY"
+    THOUGHTSPOT = "THOUGHTSPOT"
 
 
 class OssieDataType(str, Enum):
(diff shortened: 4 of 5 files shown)


core-spec/ossie-schema.json
```diff
@@ -23,7 +23,7 @@
   "$defs": {
     "Dialect": {
       "type": "string",
-      "enum": ["ANSI_SQL", "SNOWFLAKE", "MDX", "TABLEAU", "DATABRICKS", "MAQL", "BIGQUERY"],
+      "enum": ["ANSI_SQL", "SNOWFLAKE", "MDX", "TABLEAU", "DATABRICKS", "MAQL", "BIGQUERY", "THOUGHTSPOT"],
```
jbonofre, line 26: That's ok, but worth to note that some converters will "fail".

For instance in nvidia GSF, `_pick_expression` finds no `ANSI_SQL` and silently fails through to `return expression[0]["expression"]`, writing the ThoughtSpot formula into a GSF model as a SQL expression (no warning).
I believe DBT and Honeydew converters will behave the same.

Before this PR, the schema gate rejected such a document. Now it is accepted and produces broken output two layers down. The root cause is that no converter has any machine-readable way to ask "is this dialect SQL?".

Nothing to consider soon.

jbonofre (approved):

PR #353, opened 2026-09-02
fix(dbt): substitute DERIVED metric references in a single pass

ayushtkn:
## Summary

`MSIToOssieConverter._resolve_derived` inlined each input metric of a DERIVED metric
with its own `re.sub` call, run over the string produced by the previous iteration.
Because every pass re-scanned text that earlier passes had inserted, and because the
resolved expression was passed as `re.sub`'s *replacement template*, the emitted Ossie
expression could be silently corrupted in two ways.

**1. A later reference matched text an earlier one inserted.** MetricFlow metrics are
commonly named after the column they aggregate, so this is easy to hit. With measures
`gross = SUM(net)` and `net = SUM(net_amount)` and a DERIVED metric `expr = "gross - net"`:

    before:  SUM(orders.SUM(orders.net_amount)) - SUM(orders.net_amount)
    after:   SUM(orders.net) - SUM(orders.net_amount)

**2. Backslashes in the resolved SQL were read as replacement escapes.** A backslash
surviving from a metric filter was reinterpreted on inlining. With a SIMPLE metric
filtered on `{{ Dimension('order__path') }} LIKE 'a\b'`, inlined into `expr = "revenue * 2"`:

    before:  SUM(CASE WHEN order__path LIKE 'a<backspace>' THEN orders.amount END) * 2
    after:   SUM(CASE WHEN order__path LIKE 'a\b' THEN orders.amount END) * 2

The `\b` became a literal backspace character. Other sequences (e.g. `\d`) raised
`re.error: bad escape` and aborted the conversion instead.

**The change.** Collect the `{reference: resolved expression}` pairs first, then
substitute them in a single pass using an alternation pattern with a callback
replacement. A callback is not template-expanded, so one change fixes both problems.
This is the same approach already used in the Databricks converter
(`metric_view_to_ossie.py`, `re.sub(r"\bsource\.", lambda _m: ...)`).

## Related Issues

NA

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance

### DIFF
--- a/converters/dbt/src/ossie_dbt/msi_to_ossie.py
+++ b/converters/dbt/src/ossie_dbt/msi_to_ossie.py
@@ -338,8 +338,16 @@ class MSIToOssieConverter:
         """Resolve a DERIVED metric by substituting each input metric's expression into the expr string.
 
         Compound sub-expressions (DERIVED/RATIO) are wrapped in parentheses to preserve operator precedence.
+
+        All references are substituted in a single pass. Substituting them one at a
+        time would re-scan text inserted by an earlier reference, so a metric named
+        after a column appearing in an already-inlined expression would be expanded
+        twice. The replacement is a callback rather than a string so that backslashes
+        in the resolved SQL (e.g. from a `LIKE 'a\\b'` filter) are inserted verbatim
+        instead of being interpreted as `re.sub` template escapes.
         """
         expr = metric.type_params.expr or ""
+        replacements: Dict[str, str] = {}
         for input_metric in metric.type_params.metrics or []:
             ref = input_metric.alias if input_metric.alias else input_metric.name
             dep_metric = self._lookup_metric(metric_index, input_metric.name, f"DERIVED metric '{metric.name}'")
@@ -347,8 +355,18 @@ class MSIToOssieConverter:
             resolved = self._resolve_metric_expression(dep_metric, metric_index, cache, input_filter)
             if dep_metric.type in (MetricType.DERIVED, MetricType.RATIO):
                 resolved = f"({resolved})"
-            expr = re.sub(rf"\b{re.escape(ref)}\b", resolved, expr)
-        return expr
+            replacements[ref] = resolved
+
+        if not replacements:
+            return expr
+
+        # The `\b` anchors already stop a short reference from matching inside a
+        # longer identifier; sorting by length keeps the alternation order stable
+        # and independent of the order metrics happen to be declared in.
+        pattern = re.compile(
+            r"\b(" + "|".join(re.escape(ref) for ref in sorted(replacements, key=len, reverse=True)) + r")\b"
+        )
+        return pattern.sub(lambda match: replacements[match.group(0)], expr)
 
     @staticmethod
     def _build_entity_index(
--- a/converters/dbt/src/ossie_dbt/msi_to_ossie.py
+++ b/converters/dbt/src/ossie_dbt/msi_to_ossie.py
@@ -361,10 +361,10 @@ class MSIToOssieConverter:
             return expr
 
         # The `\b` anchors already stop a short reference from matching inside a
-        # longer identifier; sorting by length keeps the alternation order stable
+        # longer identifier; sorting by length (then name) keeps the alternation order stable
         # and independent of the order metrics happen to be declared in.
         pattern = re.compile(
-            r"\b(" + "|".join(re.escape(ref) for ref in sorted(replacements, key=len, reverse=True)) + r")\b"
+            r"\b(" + "|".join(re.escape(ref) for ref in sorted(replacements, key=lambda ref: (-len(ref), ref))) + r")\b"
         )
         return pattern.sub(lambda match: replacements[match.group(0)], expr)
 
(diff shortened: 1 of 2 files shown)


converters/dbt/src/ossie_dbt/msi_to_ossie.py
```diff
@@ -338,17 +338,35 @@ def _resolve_derived(
...
+            replacements[ref] = resolved
+
+        if not replacements:
+            return expr
+
+        # The `\b` anchors already stop a short reference from matching inside a
+        # longer identifier; sorting by length keeps the alternation order stable
+        # and independent of the order metrics happen to be declared in.
+        pattern = re.compile(
+            r"\b(" + "|".join(re.escape(ref) for ref in sorted(replacements, key=len, reverse=True)) + r")\b"
```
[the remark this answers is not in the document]
ayushtkn: Done.

converters/dbt/src/ossie_dbt/msi_to_ossie.py
```diff
@@ -338,17 +338,35 @@ def _resolve_derived(
...
         for input_metric in metric.type_params.metrics or []:
             ref = input_metric.alias if input_metric.alias else input_metric.name
             dep_metric = self._lookup_metric(metric_index, input_metric.name, f"DERIVED metric '{metric.name}'")
             input_filter = _merge_filter_sqls(filter_sql, _collect_filter_sql(input_metric.filter))
             resolved = self._resolve_metric_expression(dep_metric, metric_index, cache, input_filter)
             if dep_metric.type in (MetricType.DERIVED, MetricType.RATIO):
                 resolved = f"({resolved})"
-            expr = re.sub(rf"\b{re.escape(ref)}\b", resolved, expr)
-        return expr
+            replacements[ref] = resolved
```
jbonofre, line 373: I think that collecting into `replacements[ref] = resolved` changes the collision behavior from "first entry wins" to "last entry wins". I'm not sure it's addressed anywhere.

if a derived metric lists the same input metric twice with different per-input filters and no alias, this dict collapses to one entry and both occurrences get F2's SQL. This input shape isn't rejected upstream: MetricFlow's `DerivedMetricRule._validate_alias_collision` only compares entries that have an `alias` set, so two unaliased duplicates sail through validation. 

Worth either erroring on a duplicate unaliased ref here, or confirming this silent overwrite is intentional (and so it should be documented).

converters/dbt/tests/test_msi_to_ossie.py
```diff
@@ -644,6 +644,80 @@ def test_derived_metric_uses_alias_for_substitution(self) -> None:
         profit_ossie = next(m for m in _ossie_metrics(result) if m.name == "profit")
         assert profit_ossie.expression.dialects[0].expression == "SUM(orders.amount) - SUM(orders.cost_amount)"
 
+    def test_derived_metric_does_not_re_expand_an_inlined_reference(self) -> None:
```
jbonofre, line 647: I suggest to add a test for a DERIVED metric whose `type_params.metrics` contains two entries resolving to the same reference (same name, same alias) with different filters?

That's the case where the dict-based `replacements` collapses to one entry and silently picks whichever occurrence was declared last (there's no coverage for that ordering behavior right now).

converters/dbt/src/ossie_dbt/msi_to_ossie.py
ayushtkn: Thanx @jbonofre for the review. Good catch — silent last-wins wasn’t intentional. `MetricFlow` doesn’t reject this shape (`DerivedMetricRule._validate_alias_collision` only compares aliased entries), and with a single token in expr neither resolution is more correct. Raised a `ValueError` when the same reference resolves differently, and pointed at distinct aliases as the fix. Identical duplicates stay accepted since they’re redundant, not ambiguous. Covered in 6952a43.

converters/dbt/tests/test_msi_to_ossie.py
ayushtkn: Added two tests in 6952a43: one that rejects the same reference listed twice with differing filters, and one that accepts a redundant identical duplicate (revenue + revenue). Went with reject rather than asserting last-wins ordering.

ayushtkn:
Hi @jbonofre does the latest changes look good? Let me know if it requires further changes or investigation :-)

jbonofre:
@ayushtkn I'm doing a new pass. Thanks!

jbonofre (commented):
Thanks for the fix, I think we are very close!

The single-pass substitution change itself is solid and well covered by the new tests. 

Before merging, I propose:
1. to fix the two false positives at `msi_to_ossie.py:367`: a duplicated input metric with no per-input filter vs one that restates the parent own filter currently render to different text (`"(X)"` vs `"(X) AND (X)"`) and trips the check even though they are logically identical. And a duplicated-but-unused reference (never appearing in `expr`) should not be checked at all.
2. to fail soft instead of hard-raising: catch this case the same way the converter handles every other unsupported shape (`CONVERSION_METRIC_DROPPED`, etc), append a `ConverterIssue` and drop that one metric, rather than letting `ValueError` propagate uncaught through `cli.py` and kill the whole run.

I would be happy to pair on this or take a stab at the fix myself if that's faster.
Once those two are addressed I'm good to approve and merge.

ayushtkn:
Thanx @jbonofre for helping with the review. I have quickly tried and pushed a commit to address the review comment. I am good with if you want to collaborate or wanna take it over as well. I can help with testing and reviewing the changes :-)

jbonofre (approved):

PR #363, opened 2026-09-06
fix(examples): use d_moy for the TPC-DS date_dim month field

### LINKED ISSUE #357: core-spec: version is a const, but converter fixtures declare 0.1.1 and fail the schema
`core-spec/ossie-schema.json` pins the document version to a single value:

```json
"version": {
  "type": "string",
  "const": "0.2.0.dev0",
  "description": "Apache Ossie specification version"
}
```

But `converters/orionbelt/tests/fixtures/tpcds_ossie.yaml` begins:

```yaml
version: "0.1.1"
```

So that fixture fails the project's own schema.

## Reproduction

```python
import json, yaml, urllib.request
from jsonschema import Draft202012Validator

raw = "https://raw.githubusercontent.com/apache/ossie/main/"
schema = json.load(urllib.request.urlopen(raw + "core-spec/ossie-schema.json"))
doc = yaml.safe_load(urllib.request.urlopen(raw + "converters/orionbelt/tests/fixtures/tpcds_ossie.yaml"))

print([e.message for e in Draft202012Validator(schema).iter_errors(doc)])
# ["'0.2.0.dev0' was expected"]
```

That is the only validation error the fixture produces, so the version is the sole mismatch.

## Why this came up

We build a DuckDB implementation that reads Ossie models and executes queries against them. We
vendored several converter fixtures as a conformance corpus — checking both that our reader accepts
them and that they validate against the published schema. That is how the mismatch surfaced.

## The underlying question

A `const` means exactly one spec version is ever valid, so every document has to be rewritten in
lockstep with the spec and no implementation can accept a range. That may well be intended during
incubation, but it does mean:

- documents in the wild (`0.1.1`, `0.1.0`) become schema-invalid the moment the const moves
- an implementation cannot distinguish "written against an older spec" from "malformed"

If a range is wanted eventually, `pattern` or `enum` would express it. If `const` is deliberate, then
the orionbelt fixture is simply out of date and could be bumped.

Happy to send a PR either way — just let me know which you would prefer.

*(Edited to correct the file paths: these were `osi-schema.json` and `tpcds_osi.yaml` when I first
looked, and have since been renamed to `ossie-`. Re-verified against `main` today; both defects are
unchanged.)*

### LINKED ISSUE #358: converters/orionbelt fixture declares date_dim.d_month_name, which is not a TPC-DS column
`converters/orionbelt/tests/fixtures/tpcds_ossie.yaml` declares a field on the `date_dim` dataset
that TPC-DS does not have:

```yaml
205:          - name: d_month_name
209:                  expression: d_month_name
```

The real columns are `d_moy` (month of year), `d_month_seq`, and `d_current_month`. There is no
month-*name* column in the TPC-DS schema at all.

## Reproduction

Any DuckDB, no extra tooling:

```sql
INSTALL tpcds; LOAD tpcds;
CALL dsdgen(sf = 0.01);

SELECT column_name FROM duckdb_columns()
WHERE table_name = 'date_dim' AND column_name ILIKE '%mo%';
```

```
┌─────────────────┐
│   column_name   │
├─────────────────┤
│ d_month_seq     │
│ d_moy           │
│ d_current_month │
└─────────────────┘
```

No `d_month_name`. Anything that resolves that field against a real TPC-DS table fails; in our case:

```
Binder Error: Referenced column "d_month_name" not found in FROM clause!
Candidate bindings: "d_month_seq"
```

## How it was found

We build a DuckDB implementation that *executes* Ossie models rather than converting them, so we can
run every metric against every dimension and diff the results against hand-written TPC-DS SQL. Doing
that at `sf=1` surfaced it. Our own fixtures had inherited the same mistake from the same source and
are now corrected to `d_moy`; yours is left untouched since it is your file.

## Why it survives in a converter test suite

Nothing in a conversion round-trip touches the physical table, so a field naming a column that does
not exist converts perfectly and never fails. It only shows up when something tries to *run* a query
that groups by it.

That is not a criticism of the fixture so much as an observation that execution catches a class of
error conversion cannot — which may be worth keeping in mind for the `compliance/` suite.

Suggested fix is a one-line rename to `d_moy`, or dropping the field. Happy to send the PR.

*(Edited to correct the file path: this was `tpcds_osi.yaml` when I first looked and has since been
renamed. Re-verified against `main` today; the field is still there at lines 205 and 209.)*

kayemkim:
## Summary

`examples/tpcds_semantic_model.yaml` declares a `d_month_name` field on `date_dim` with `expression: d_month_name`. TPC-DS has no such column. The month column is `d_moy` (1-12); `d_day_name` and `d_quarter_name` do exist, which is probably where the name came from. A conversion round-trip never touches the physical table, so the field converts cleanly in every suite and only fails when something executes the model against real TPC-DS data, as reported in #358.

This renames the field and its expression to `d_moy` and updates the description. Several converter fixtures are copies or conversions of the canonical example, so the same rename is applied to them, both to keep them consistent and because the omni tests export the canonical example and compare it against the `tpcds_omni` fixture:

- `converters/orionbelt/tests/fixtures/`: `tpcds_ossie.yaml`, `tpcds_semantic_model.yaml`, `tpcds_as_obml.yaml`
- `converters/omni/tests/fixtures/tpcds_omni/views/date_dim.view.yaml`
- `converters/snowflake/tests/example_converted_tpcds_semantic_model.yaml`

The field keeps `is_time: true` with no `datatype`, since the comment above it uses this field to show that the two annotations are independent.

Separately, two orionbelt fixtures (`tpcds_ossie.yaml`, `obml_as_ossie.yaml`) still declared `version: 0.1.1`, which the schema's `version` const rejects. They now declare `0.2.0.dev0`. No test loads either file, so this only changes standalone validation. Whether `version` should remain a `const` is a separate question raised in #357; this PR only brings the fixtures in line with the schema as it stands.

Checked:

- DuckDB `tpcds` extension at `sf=0.01`: `date_dim` has 28 columns, none named `d_month_name`; `d_moy` resolves.
- `validation/validate.py` passes on the canonical example and on all three orionbelt Ossie-format fixtures. Before this change `tpcds_ossie.yaml` failed with `'0.2.0.dev0' was expected`.
- Converter tests: orionbelt 154 passed, omni 83 passed / 3 skipped, snowflake 107 passed, honeydew TPC-DS round-trip passed.
- `grep -rn d_month_name` over the repo returns nothing.

## Related Issues

Fixes #357. Fixes #358.

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples
- [x] `examples/` are added or updated for any new spec constructs or converter support

### Tests

### Compliance

### DIFF
--- a/converters/omni/tests/fixtures/tpcds_omni/views/date_dim.view.yaml
+++ b/converters/omni/tests/fixtures/tpcds_omni/views/date_dim.view.yaml
@@ -57,8 +57,8 @@ dimensions:
     - month
     - quarter
     - year
-  d_month_name:
-    description: Month name
+  d_moy:
+    description: Month of year (1-12)
     synonyms:
     - month
     timeframes:
--- a/converters/orionbelt/tests/fixtures/obml_as_ossie.yaml
+++ b/converters/orionbelt/tests/fixtures/obml_as_ossie.yaml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-version: '0.1.1'
+version: '0.2.0.dev0'
 semantic_model:
 - name: sales_analytics
   description: Sales, Returns, Shipments and Purchases Analytics
--- a/converters/orionbelt/tests/fixtures/tpcds_as_obml.yaml
+++ b/converters/orionbelt/tests/fixtures/tpcds_as_obml.yaml
@@ -137,10 +137,10 @@ dataObjects:
         customExtensions:
         - vendor: Ossie
           data: '{"synonyms": ["quarter", "fiscal quarter"]}'
-      d_month_name:
-        code: d_month_name
+      d_moy:
+        code: d_moy
         abstractType: date
-        comment: Month name
+        comment: Month of year (1-12)
         customExtensions:
         - vendor: Ossie
           data: '{"synonyms": ["month"]}'
@@ -299,9 +299,9 @@ dimensions:
     dataObject: date_dim
     column: d_quarter_name
     resultType: date
-  d_month_name:
+  d_moy:
     dataObject: date_dim
-    column: d_month_name
+    column: d_moy
     resultType: date
   c_customer_id:
     dataObject: customer
--- a/converters/orionbelt/tests/fixtures/tpcds_ossie.yaml
+++ b/converters/orionbelt/tests/fixtures/tpcds_ossie.yaml
@@ -21,7 +21,7 @@
 # This example demonstrates the Ossie Core Metadata Spec using the TPC-DS benchmark schema
 # TPC-DS is a decision support benchmark with a realistic retail business model
 
-version: "0.1.1"
+version: "0.2.0.dev0"
 
 semantic_model:
   - name: tpcds_retail_model
--- a/converters/orionbelt/tests/fixtures/tpcds_ossie.yaml
+++ b/converters/orionbelt/tests/fixtures/tpcds_ossie.yaml
@@ -202,12 +202,12 @@ semantic_model:
                 - "quarter"
                 - "fiscal quarter"
 
-          - name: d_month_name
+          - name: d_moy
             expression:
               dialects:
                 - dialect: ANSI_SQL
-                  expression: d_month_name
-            description: Month name
+                  expression: d_moy
+            description: Month of year (1-12)
             dimension:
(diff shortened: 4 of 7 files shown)


jbonofre (approved):
Good catch! Thanks!

PR #364, opened 2026-09-07
feat(thoughtspot): bidirectional ThoughtSpot TML <-> Ossie converter

djwaldo:
Adds `converters/thoughtspot/` — a bidirectional converter between ThoughtSpot's semantic model format (Model TML plus its Table and SQL View documents) and the Ossie semantic model.

Proposed in #285 (consolidating #269); `THOUGHTSPOT` was added to the `Dialect` enum in #351.

## Scope

| Ossie | ThoughtSpot |
|---|---|
| `datasets` | Model `model_tables[]` + a Table or SQL View document |
| `fields` | Model `columns[]` with `column_type: ATTRIBUTE` |
| `metrics` | Model `columns[]` with `column_type: MEASURE`, always via `formulas[]` |
| `relationships` | Model inline joins and Table `joins_with[]` |
| `custom_extensions[THOUGHTSPOT]` | everything with no Ossie home |

One structural note that shapes the whole design: **one Ossie semantic model corresponds to 1 + N TML documents**, not one file. The converter reads and writes the set.

## Expression handling — the design decision most likely to draw questions

ThoughtSpot's formula language is not SQL. It has its own syntax (`concat ( [a] , [b] )`, `[TABLE::Column]` references, `{ }` grouping) which no SQL dialect parses.

**A ThoughtSpot formula is carried verbatim under a `THOUGHTSPOT` dialect entry. A portable `ANSI_SQL` sibling is added only where the whole expression is a bare column reference — the one shape where portability is certain. No SQL dialect is ever re-rendered into another, in either direction.**

That is deliberate, and the reasoning is in `converters/thoughtspot/README.md`:

- The specification's own stated default is to pass unknown values through, and `dialects[]` exists precisely so an untranslatable expression can still travel tagged with the dialect it is valid in.
- No converter in this repository re-renders an expression across dialects. The Databricks converter carries expressions verbatim under its own vendor dialect and reads that first.
- Round-trip fidelity is therefore exact rather than best-effort: the return leg reads the verbatim entry.

`expressions/catalog.py` maps all 146 constructs the Ossie expression language defines to a ThoughtSpot rendering (108 direct, 37 via `sql_*` passthrough, 1 unmappable) and is used for `Ossie → TML`. `expressions/reverse.py` records which ThoughtSpot-native functions compose into a portable Ossie expression; it is present for testing and future use and is **not yet wired into the shipped `TML → Ossie` path** — the README says so explicitly rather than leaving it to be inferred.

## Reference documentation is generated from the code

`converters/thoughtspot/docs/` is produced by `tools/generate_reference_docs.py` from `CATALOG`, the reverse inventory, the datatype map and the vendor-payload key classification. A test compares the committed documents byte-for-byte against fresh generator output, so they cannot drift from the implementation.

## Testing

804 tests, offline, on Python 3.10–3.14.

- Both directions, unit-level.
- **Round-trip both ways** over two fixture sets, with preservation and translation asserted *separately* — the return leg reads the verbatim dialect entry, so a round trip alone proves preservation and says nothing about translation quality.
- The issue log is asserted as a first-class output: every declared loss must be reported *before* it happens, and nothing may be reported that did not occur.
- **Property-based** round trips (`hypothesis`, test-only) over deliberately awkward identifiers — YAML 1.1 boolean tokens, names colliding only after normalisation, non-ASCII, names containing `::`. This found two crashes the example fixtures never produced.
- Fixtures mirror `examples/tpcds_semantic_model.yaml` so this converter is comparable to its siblings; both expected documents validate against `core-spec/ossie-schema.json` and `validation/validate.py`.

## Dependencies

`PyYAML>=6.0` and nothing else at runtime. `hypothesis` and `jsonschema` are test-only.

## Provenance

All TML fixtures and all code are original, authored for this contribution. Nothing derives from ThoughtSpot's `thoughtspot_tml` library or any other existing converter.

## A note on relationship orientation

Ossie's `Relationship` has no cardinality field — direction carries it, with `from` documented
as "the logical dataset on the many side" and `to` as "the one side". ThoughtSpot's joins carry
an explicit `cardinality`, including `ONE_TO_MANY`, which has no Ossie representation in the
direction TML declares it.

The converter therefore **swaps the endpoints** for a `ONE_TO_MANY` join, so the emitted
relationship is always many-side-to-one-side as the specification requires; `MANY_TO_ONE` and
`ONE_TO_ONE` pass through unchanged. Both TML spellings of the same relationship consequently
produce the same Ossie relationship, and the vendor extension records the swap so the return
trip reproduces the original TML join — same direction, same target, same cardinality.

Worth flagging for reviewers: `validation/validate.py` does **not** catch an inverted
relationship, because its key-coverage check skips a dataset that declares no keys. An earlier
revision of this branch emitted the un-swapped form and validated clean.

## Note on history

81 commits, conventional-commit style. Happy to squash to a smaller set of logical commits if that is preferred for review.


djwaldo:
i've Rebased onto current `main` and migrated to the single-model document format from #383, the model's fields now sit at the document root, and a document still carrying the `semantic_model` wrapper is rejected by name rather than unwrapped, matching #396.

823 tests pass on 3.10–3.14, offline. Both fixtures and the emitted output validate against the current `core-spec/ossie-schema.json`, and `to-ossie` → `to-tml` round-trips the TPC-DS fixture end to end.

Two things I can't do from this side:

- **CI has never run.** Every workflow on the branch sits at `action_required`, so it needs a
  committer to approve the run.
- **No reviewer yet.** Open since 7 September.

Could a committer approve CI and take a look?

.github/workflows/converter-thoughtspot-ci.yml
```diff
@@ -0,0 +1,69 @@
...
+      - 'converters/thoughtspot/**'
+      - '.github/workflows/converter-thoughtspot-ci.yml'
+      - 'core-spec/expression_language.md'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```
QMalcolm: Worth noting, python 3.10 goes EOL at the end of this month. Maybe not a problem of this PR, but we should drop 3.10 shortly.

converters/thoughtspot/docs/expression-mapping.md
```diff
@@ -0,0 +1,208 @@
...
+Every construct the Ossie expression language specification defines, mapped to its ThoughtSpot rendering (`Ossie -> ThoughtSpot`, the direction `expressions/catalog.py` drives). Rows follow the source's own definition order, which groups related constructs together (aggregates, then type conversion, date/time, string, math/conditional, operators, window functions) — that grouping exists only as source comments, not as data the code carries, so it is not reproduced as separate sections here.
+
+## Coverage
+
+| Classification | Count | Share |
+|---|---|---|
+| direct | 108 | 74% |
+| passthrough | 37 | 25% |
+| unmappable | 1 | 1% |
+| **Total** | **146** | **100%** |
```
QMalcolm: Am I understanding it this correct that 1% of things being unmappable means that 1% of things can't be converted and thus dropped? If so, that by itself is fine. However, in this section called "coverage" we're putting `100%` in bold, which reads like there being 100% coverage. Perhaps we should format this differently to remove the possible ambiguity/misreading.

djwaldo:
Rebased onto current `main` (now includes #396) and widened this converter's CI path filters to `core-spec/**` + `examples/**`, matching the fourteen siblings #396 updated — this one wasn't in that PR, so it was the only converter whose CI didn't re-run on a spec change.

823 tests pass locally. Apologies — the push reset the run you approved on the 17th, so it needs approving once more.

jbonofre:
@djwaldo thanks! Let me do the review.

converters/thoughtspot/src/ossie_thoughtspot/expressions/catalog.py
```diff
@@ -0,0 +1,1931 @@
...
+# way: also no native function, but their compositions use only native
+# functions (strpos/substr/strlen), so they stay direct. There is no
+# regular-expression support of any kind, so every REGEXP_* row is passthrough
+# with no native fallback.
+# --------------------------------------------------------------------------
+CATALOG.update(
+    {
+        "CONCAT(str1, str2, ...)": Construct(
+            "CONCAT(str1, str2, ...)", Classification.DIRECT,
+            template="concat ( {0} , {1} , ... )",
```
jbonofre: The `DIRECT` template for `CONCAT`/`GREATEST`/`LEAST` embed a literal `...` token that `emit_direct` `str.format` never fills. Every real call (including the base 2-argument case) renders with a stray literal in the output, e.g. `concat ( [T::First] , [T::Last] , ... )`, which is invalid ThoughSpot syntax (I believe, but you know better than me 😄 ). The catalog-wide sweep test only checks that `emit_direct` doesn't raise, so this has no coverage that would catch it.

I think it needs a template that actually varies with arg count, or a dedicated N renderer instead of a fixed `DIRECT` string.

Thoughts?

converters/thoughtspot/src/ossie_thoughtspot/expressions/catalog.py
```diff
@@ -0,0 +1,1931 @@
...
+            ),
+        ),
+        "BETWEEN": Construct(
+            "BETWEEN", Classification.DIRECT,
+            template="{0} between {1} and {2}",
+            note="Inclusive on both sides.",
+        ),
+        "IN": Construct(
+            "IN", Classification.DIRECT,
+            template="{0} in {{ {1} , {2} , ... }}",
```
jbonofre: Same `...` literal defect as the `CONCAT` template, plus this one is hardcoded to exactly two values.

`emit_direct(CATALOG["IN"], [...])` with any list other than 2 times either renders the stray `...` or doesn't match the template arity at all: there's no correct rendering path for real IN lists of 1, 3, or 4+ values.

converters/thoughtspot/src/ossie_thoughtspot/expressions/catalog.py
```diff
@@ -0,0 +1,1931 @@
...
+            template="if ( {0} ) then {1} else {2}",
+            note="Specification alias for IF.",
+        ),
+        "NULLIF(expr1, expr2)": Construct(
+            "NULLIF(expr1, expr2)", Classification.DIRECT,
+            template="nullif ( {0} , {1} )",
+        ),
+        "COALESCE(expr1, expr2, ...)": Construct(
+            "COALESCE(expr1, expr2, ...)", Classification.DIRECT,
+            template="ifnull ( {0} , ifnull ( {1} , {2} ) )",
```
jbonofre: COALESCE's DIRECT template hardcodes exactly 3 arguments, but the comment above it notes 2 is the common case. `emit_direct(CATALOG["COALESCE(expr1, expr2, ...)"], ["a", "b"])` raises `ValueError: ...expects 3 arguments, got 2`. Same failure for 4+. The two most common real-world arities both crash.

converters/thoughtspot/src/ossie_thoughtspot/tml_to_ossie.py
```diff
@@ -0,0 +1,2347 @@
...
+
+    Either fallback always differs from `display_name`, so a caller that
+    already stashes the original display name whenever the identifier
+    differs from it (metrics do; fields carry it in `label` instead, which
+    is populated independently of this call and needs no stash) picks this
+    case up for free — nothing here writes a stash entry itself, only logs
+    the WARNING naming what happened.
+    """
+    try:
+        return identifiers.normalise(display_name)
```
jbonofre, line 534: `_field_or_metric_identifier` common-case path returns `identifiers.normalise(display_name)` directly with no collision check against identifiers already minted in this scope, unlike the Ossie -> TML allocator, which always checks.

Two ATTRIBUTE columns display-named `"Order Date"` and `"Order-Date"` both normalise to `order_date` and end up as two fields sharing one identifier in the same dataset, with nothing logged.

Same gap applied to Metrics, which are unique model-wide rather than per-table. This path needs to go through the same collision checked allocator the other direction uses.

converters/thoughtspot/src/ossie_thoughtspot/tml.py
```diff
@@ -0,0 +1,247 @@
...
+    out = []
+    for document in documents:
+        suffix = _SUFFIX[document.kind]
+        max_stem_bytes = _MAX_FILENAME_BYTES - len(f".{suffix}") - counter_reserve
+        stem = _safe_filename_component(document.body.get("name", document.kind))
+        stem = _truncate_utf8(stem, max_stem_bytes).rstrip(" .") or "_unnamed"
+
+        candidate = f"{stem}.{suffix}"
+        counter = 1
+        while candidate in used:
```
jbonofre: `dump_document_set()` filename collision avoidance compares names case-sentivitively, but the default target filesystem (macOS, Windows) are case-insensitive. Two documents named `Store_Sales` and `STORE_SALES` get distinct looking filenames with no counter suffix, and the second write silently overwrites the first on disk (exit code 0, nothing logged). 

I think collision check needs to normalise case (or otherwise account for case-insensitive filesystems) before deciding two names are distinct.

jbonofre (changes requested):

djwaldo:
Thanks — all five reproduce, and all five are fixed in 4ace101. Your `CONCAT` instinct was
right: the `...` is not valid ThoughtSpot syntax, it's the spec's own ellipsis notation copied
into the template.

I fixed 1–3 as a class rather than row by row, which turned up a sixth you didn't hit:
**`NOT IN` has the same ellipsis defect** as `IN`.

`Construct` now declares `variadic=Variadic(style, min_args, tail_from)`. `JOIN` substitutes the
argument tail at a `{*}` marker; `FOLD` applies a binary template right-associatively, which is
what `COALESCE` needs since `ifnull` has no n-ary form:

```
CONCAT   2→ concat ( a , b )              4→ concat ( a , b , c , d )
COALESCE 2→ ifnull ( a , b )              3→ ifnull ( a , ifnull ( b , c ) )
IN       2→ a in { b }                    4→ a in { b , c , d }
```

**On why the sweep missed it** — you're right that it only checked `emit_direct` doesn't raise,
and the deeper problem is that it derived each row's arity *from* its own template. Guard and
test read the same string, so no template could ever disagree with itself. It now sweeps
variadic rows across a range of arities and asserts on the **output**: no literal `...`, no
unfilled placeholder, balanced parens and braces. A literal `...` in any template is also
refused at construction now, so a new row can't reintroduce it. I mutation-checked both guards
against the original `CONCAT` template — each fails on it independently.

**Finding 4 — one judgment worth your eye.** I resolved collisions where the owning scope is
first known (per dataset for fields, the flat `metrics[]` for metrics) rather than inside
`_field_or_metric_identifier`. That function's allocator is deliberately model-wide, so routing
the common case through it would rename `orders.amount` and `customers.amount` to `amount` and
`amount_2` — legitimately distinct fields in different datasets. Renames are reported at
WARNING rather than applied quietly. Say if you'd rather they were an error.

**Finding 5** reproduced as real data loss on macOS — 3 documents emitted, 2 on disk. Comparison
now folds case; emitted names keep theirs.

824 tests pass. PTAL when you get a chance.

jbonofre:
@djwaldo thank you so much! I will do a new pass.

djwaldo:
Your review prompted me to run a max-effort pass over the whole converter rather than just the
five rows. It found **nine more defects of the same character**, all now fixed in 7e7309b and
a05e3b4. 851 tests pass, up from 824.

The three worth your attention, because they produced wrong data rather than wrong syntax:

- **Silent double aggregation.** `_AGGREGATE_CALL_NAMES` knew `group_aggregate` alone — above a
  comment asserting "there is exactly one such construct" — while this package's own reverse
  inventory registered four shorthands and two more `*_aggregate_op` variants. A MEASURE column
  using any of them got its own `aggregation` composed on top: `sum ( group_sum ( ... ) )`, no
  issue logged.
- **A parenthesised join condition lost its relationship.** `( [A::x] = [B::y] and [A::p] =
  [B::q] )` put every `and` at paren-depth 1, so nothing split, no equality pair matched, and the
  whole relationship became an unrepresentable-join stash entry — datasets disconnected, no key
  candidates.
- **Moving-window frame bounds were inverted at one end.** `moving_average(m, 2, 3, d)` emitted
  `BETWEEN 2 PRECEDING AND 3 PRECEDING` — valid SQL over the wrong five rows.

Plus: an aliased self-join emitted two Table documents of the same name; `zip` truncated a join
condition when the column arrays disagreed; `synonym_type` was overwritten after being faithfully
stashed; shared list objects put YAML anchors in every generated document; `_safe_target_path`'s
condition let through the one case it existed to catch; and two input shapes escaped the CLI as
bare tracebacks.

**Two things I'd flag as process rather than code.** Several of these were *asserted by existing
tests* — the frame-bound test pinned the defect and its comment recorded the misreading as fact,
so I settled the convention against the Tableau and Looker mappings instead and cited both in the
test. And your point about the catalog sweep generalised further than the five rows: it derived
each row's arity *from its own template*, so no template could ever disagree with itself. It now
asserts on the rendered output.

Also fixed the 11 passthrough rows that bake a caller-supplied value in as a literal —
`NTILE(4)` was being published as the mapping for `NTILE(n)`. They now declare it, the generated
doc marks them "example only", and construction refuses a new undeclared one.

CI needs approving again on the new head, if you would. PTAL.

converters/thoughtspot/src/ossie_thoughtspot/ossie_to_thoughtspot.py
```diff
@@ -0,0 +1,2216 @@
...
+    order: list[str] = []
+    for table in tables:
+        name = table.body.get("name")
+        first = by_name.get(name)
+        if first is None:
+            by_name[name] = table
+            order.append(name)
+            continue
+
+        seen_columns = {c.get("name") for c in first.body.get("columns") or []}
```
jbonofre: This merge only read `columns`, but SQL View documents store output columns in `sql_view_columns`. When two aliases datasets share a SQL View and surface different fields, the second set is discarded.

`build_model` then checks both aliases against the first document, so fields only present in the second set are omitted with an error.

Could this merge and comparison use the column key for the document kind?

converters/thoughtspot/src/ossie_thoughtspot/cli.py
```diff
@@ -0,0 +1,241 @@
...
+
+    try:
+        texts = [(path, Path(path).read_text(encoding="utf-8")) for path in args.tml_files]
+        document_set = tml.load_document_set(texts)
+        result = tml_to_ossie.convert(document_set)
+    except (ConversionError, OSError, UnicodeDecodeError) as e:
+        print(f"Error: {e}", file=sys.stderr)
+        return 1
+
+    targets = [output_path] + ([issues_path] if issues_path else [])
```
jbonofre: If `--issues` and `-o` resolve to the same path on a fresh run, this check passes, then `_write_issues` replaces the converted document with JSON.

The same collision can happen between `--issues` and a generated `to-tml` file.

Could the CLI reject colliding paths before writing?

djwaldo:
Your two reviews prompted a much wider pass. Seventeen commits since, and the most useful
thing in them did not come from more code review — it came from running converted models
against a live ThoughtSpot cluster.

**Round-tripped 30 real customer models** (161 tables, 1,471 columns, 202 formulas, 146 joins),
export → Ossie → TML. All 30 pass `validation/validate.py`, and no model loses a table, column
or join without saying so.

**Then imported one back into the cluster and compared the numbers.** A three-table joined
model, converted and re-imported, answers identically to its source — totals, five regions, a
formula-derived attribute, three distinct counts, to the last decimal.

That exercise found four defects that five rounds of adversarial review and 900+ tests had all
missed, because every one of them is invisible to document-level validation:

- **`with` was omitted from a referencing join.** Mandatory on every `model_tables[].joins[]`
  entry. The document passes the Ossie schema, passes `validation/validate.py`, and ThoughtSpot
  refuses it: *"Compulsory Field worksheet->model_tables(2nd)->joins(1st)->with is not
  populated."* Nothing caught it because the hand-authored fixtures omitted `with` too — code
  and tests shared one misunderstanding and agreed with each other.
- **A bare `group_aggregate(...)` lost its column `aggregation`.** For `sum(...)` and the
  `group_sum` shorthands that property is genuinely inert, and the converter discarded it for
  all of them. A bare `group_aggregate` is the exception: like a raw column, ThoughtSpot applies
  it. A changed answer, silently.
- **Formulas that no `columns[]` entry surfaces were dropped.** These are hidden helpers other
  formulas reference — a date-parameter model is the usual case. 32 of 41 formulas in one real
  model, and the visible ones referencing them came back with dangling references.
- **`obj_id` was discarded** along with `guid` and `fqn`. It is not the same kind of thing: it
  is the handle that survives import and lets a re-import update the object it came from instead
  of duplicating it.

Also fixed since your review: a formula cross-reference rebinding to a *different* formula;
duplicate relationship names producing a document `validation/validate.py` rejects; a renamed
dataset dangling every reference to itself; a Table merged with a SQL View; display-name
collisions invented between names ThoughtSpot treats as distinct. 824 tests then, 936 now, every
fix pinned by a test that fails when it is reverted.

**Two things worth saying plainly.** Several of these defects were *asserted by our own tests* —
one sweep derived each row's expected arity from the row's own template, so nothing could ever
disagree with itself; another pinned its defect and wrote the misreading into a comment as fact.
And four of the defects fixed in later rounds were introduced by earlier rounds' fixes. The rate
is dropping and the suite is much harder to fool than it was, but I would not describe this as
finished, and I would rather say so than present seventeen commits as convergence.

CI needs approving again on the new head.

jbonofre:
@djwaldo amazing work! Thanks for the complete pass! I don't have ThoughtSpot cluster and I'm not ThoughtSpot expert as you are 😄 I just checked the code.

Let me do another pass, but I propose to merge quickly (we can always improve the converter later).

Thanks again!

converters/thoughtspot/src/ossie_thoughtspot/tml_to_ossie.py
```diff
@@ -0,0 +1,2567 @@
...
+                    f"name {prefix!r}; only the first is converted"
+                ),
+                object_ref=object_ref,
+            )
+            continue
+        table_doc = document_set.table_by_name(table_ref) if table_ref else None
+        if table_doc is None:
+            log.add(
+                code="TS-DATASET-TABLE-MISSING",
+                severity=Severity.WARNING,
```
jbonofre: Could this resolve the physical column to its Ossie field identifier before emitting the ANSI_SQL sibling?

For example, if the Ossie field is `total_revenue` but its warehouse column is `NET_AMOUNT`, this emits `orders.NET_AMOUNT`. Ossie's portable expression language resolves `dataset.field` against logical fields, so consumers may not be able to evaluate this reference.

We could emit the logical field name, or omit the sibling when that mapping cannot be made safely.

converters/thoughtspot/src/ossie_thoughtspot/cli.py
```diff
@@ -0,0 +1,283 @@
...
+            return 1
+
+    try:
+        output_path.parent.mkdir(parents=True, exist_ok=True)
+        output_path.write_text(_yaml.dump(result.model), encoding="utf-8")
+        _write_issues(result.issues, issues_path)
+    except OSError as e:
+        print(f"Error: {e}", file=sys.stderr)
+        return 1
+
```
jbonofre: This only checks whether targets already exists: it doesn't detect duplicate paths among the generated files and `--issues`.

If `--issues` is equal to a new output path, the conversion writes the document and then overwrites it with issue JSON.

I think (as a follow-up) we should reject colliding paths before writing. The same problem exists in to-ossie.

jbonofre (approved):

jbonofre (approved):
I left a couple of comments as follow-up, but this converter is a great work.

Thanks again for your contribution! Great one!

PR #371, opened 2026-09-09
[DBT] Emit OFFSET_SEMANTICS_LOSS when a DERIVED metric input uses offset_window/offset_to_grain

### LINKED ISSUE #302: msi_to_osi silently drops offset_window / offset_to_grain on DERIVED metric inputs, producing constant-zero expressions
`MSIToOSIConverter._resolve_derived` ignores `offset_window` and `offset_to_grain` on DERIVED metric inputs. The offset is dropped silently — no `ConverterIssue` is emitted — so a period-over-period metric converts into an expression that is algebraically constant, and looks like a successful conversion.

**Root cause**

`converters/dbt/src/ossie_dbt/msi_to_osi.py:337-351`, `_resolve_derived`:

```python
expr = metric.type_params.expr or ""
for input_metric in metric.type_params.metrics or []:
    ref = input_metric.alias if input_metric.alias else input_metric.name
    dep_metric = self._lookup_metric(metric_index, input_metric.name, f"DERIVED metric '{metric.name}'")
    input_filter = _merge_filter_sqls(filter_sql, _collect_filter_sql(input_metric.filter))
    resolved = self._resolve_metric_expression(dep_metric, metric_index, cache, input_filter)
    if dep_metric.type in (MetricType.DERIVED, MetricType.RATIO):
        resolved = f"({resolved})"
    expr = re.sub(rf"\b{re.escape(ref)}\b", resolved, expr)
```

`input_metric.alias`, `.name` and `.filter` are all read; `input_metric.offset_window` and `input_metric.offset_to_grain` are not read anywhere in the file. Two inputs referencing the same metric — one current, one offset — therefore resolve to byte-identical SQL, and the difference between them collapses to zero.

The year-over-year shape this breaks is the canonical use of `offset_window`, and the one in the MetricFlow docs.

```python
import sys

from tests.helpers import _manifest, _simple_metric
from metricflow_semantic_interfaces.implementations.elements.measure import PydanticMeasure
from metricflow_semantic_interfaces.implementations.metric import (
    PydanticMetric, PydanticMetricInput, PydanticMetricTimeWindow, PydanticMetricTypeParams,
)
from metricflow_semantic_interfaces.test_utils import default_meta, semantic_model_with_guaranteed_meta
from metricflow_semantic_interfaces.type_enums import AggregationType, MetricType

sm = semantic_model_with_guaranteed_meta(
    name="bookings",
    measures=[PydanticMeasure(name="booking_count", agg=AggregationType.SUM, expr="1")],
)
base = _simple_metric("booking_count", "booking_count")

yoy = PydanticMetric(
    name="bookings_yoy_growth",
    description=None,
    type=MetricType.DERIVED,
    type_params=PydanticMetricTypeParams(
        expr="(this_year - last_year) / NULLIF(last_year, 0) * 100",
        metrics=[
            PydanticMetricInput(name="booking_count", alias="this_year"),
            PydanticMetricInput(name="booking_count", alias="last_year",
                                offset_window=PydanticMetricTimeWindow(count=1, granularity="year")),
        ],
    ),
    filter=None, metadata=default_meta(), config=None,
)

res = MSIToOSIConverter().convert(_manifest(semantic_models=[sm], metrics=[base, yoy]))
out = [m for m in res.output.semantic_model[0].metrics if m.name == "bookings_yoy_growth"][0]
print("EXPRESSION:", out.expression.dialects[0].expression)
print("ISSUES:", res.issues)
```

Run from `converters/dbt/`.

Observed:

```
EXPRESSION: (SUM(1) - SUM(1)) / NULLIF(SUM(1), 0) * 100
ISSUES: []
```

The expression is algebraically `0` for every input row, and nothing signals that anything was lost.

Expected: either the offset represented in the output, or — since Ossie expressions have no window semantics to represent it with — a `ConverterIssue` reporting the loss, as `CUMULATIVE` metrics already do.

**Why this is worse than the cumulative case**

`CUMULATIVE` metrics already emit `ConverterIssueType.CUMULATIVE_SEMANTICS_LOSS` (`msi_to_osi.py:95-98`), so a consumer can see the semantics didn't survive and decide what to do. `offset_window` has no equivalent, so the failure mode is different in kind: the cumulative metric comes out *approximate*, the offset metric comes out *wrong*, and both look identical to anything downstream.

Concretely, converting through `converters/snowflake` and creating the semantic view in Snowflake produces a metric that runs successfully and returns `0` for every period — indistinguishable from genuinely flat year-on-year growth. It is the only failure I hit in an end-to-end dbt → Ossie → Snowflake conversion that produced a plausible wrong number rather than an error or an obvious gap.

**Suggested fix**

At minimum, emit a `ConverterIssue` when an input metric carries `offset_window` or `offset_to_grain`, matching the existing `CUMULATIVE_SEMANTICS_LOSS` pattern — e.g. `OFFSET_SEMANTICS_LOSS`. That turns a silent wrong answer into a visible gap, and lets consumers gate on it.

Beyond that it's a spec question rather than a converter one: representing a time-offset input needs something in the Ossie spec to carry it, which I don't think exists today (related to #290's shape, though not the same issue). Happy to raise that separately if it's useful.

`tests/test_msi_to_osi.py` has DERIVED coverage but no case where an input carries `offset_window`, which is why this isn't caught.

Happy to send a PR for the `ConverterIssue` part.

georgea-tk:
## Summary

`MSIToOssieConverter._resolve_derived` reads `alias`, `name` and `filter` off each DERIVED metric input but never `offset_window` or `offset_to_grain`. A current-vs-offset comparison (e.g. a YoY metric with one input at the current period and one offset by a year) resolves both inputs to byte-identical SQL, so the derived expression becomes something like `(SUM(x) - SUM(x)) / NULLIF(SUM(x), 0) * 100` — always zero, with `ConverterResult.issues` empty. A plausible-looking wrong number with no signal that anything was dropped.

This PR adds the minimum fix requested in #302: a new `OFFSET_SEMANTICS_LOSS` `ConverterIssueType`, emitted whenever a DERIVED metric declares an input with `offset_window` or `offset_to_grain` set, following the existing `CUMULATIVE_SEMANTICS_LOSS` pattern. It does not attempt to represent the offset semantics in the emitted expression — OSI/Ossie has no concept of a time offset today, so full representation is a separate spec question, not a converter one.

## Related Issues

Fixes #302

### Specification
- [ ] Spec changes are included in `core-spec/` and follow the existing structure — N/A, no spec change
- [ ] Spec changes have been discussed on the mailing list or in a linked issue — N/A
- [ ] Breaking changes to the spec are clearly called out in the summary — N/A, no breaking change

### Ontology
- [ ] Ontology changes in `ontology/` are consistent with spec changes — N/A
- [ ] New or modified terms are defined and documented — N/A

### Converters
- [x] New converters include tests under the converter's test directory — N/A (existing converter), but new tests added under `converters/dbt/tests/`

### Validation
- [ ] Validation rules in `validation/` are updated if the spec changed — N/A
- [ ] New validation cases are covered by tests — N/A

### Documentation
- [ ] `docs/` is updated to reflect any user-facing changes — N/A, this is an internal `ConverterIssue` signal, no user-facing doc surface
- [ ] New features or behaviors are documented with examples where appropriate — N/A
- [ ] `CONTRIBUTING.md` is updated if the contribution process changed — N/A

### Examples
- [ ] `examples/` are added or updated for any new spec constructs or converter support — N/A

### Tests
- [x] All existing tests pass (`pytest` / CI green) — ran full `converters/dbt` suite locally, 102 passed
- [x] New functionality is covered by tests — 3 new tests: `offset_window`, `offset_to_grain`, and a negative case confirming an ordinary DERIVED metric with no offset does not emit the issue

### Compliance
- [x] ASF license headers are present on all new source files — no new files, only edits to existing licensed files


converters/dbt/src/ossie_dbt/msi_to_ossie.py
```diff
@@ -110,6 +110,13 @@ def convert(
                 issues.append(
                     ConverterIssue(issue_type=ConverterIssueType.CUMULATIVE_SEMANTICS_LOSS, element_name=metric.name)
                 )
+            if metric.type is MetricType.DERIVED and any(
+                input_metric.offset_window or input_metric.offset_to_grain
+                for input_metric in metric.type_params.metrics or []
+            ):
+                issues.append(
+                    ConverterIssue(issue_type=ConverterIssueType.OFFSET_SEMANTICS_LOSS, element_name=metric.name)
+                )
```
christianeu-db, line 119: I just found https://github.com/apache/ossie/pull/318/changes by @fanruan-tuco, which has some elements that could be adopted. I like the stronger-typed message from this PR and the recursive approach from the other PR, which could capture more cases.

Perhaps, this would be a good opportunity to collaborate & merge the two PRs?

georgea-tk: Fair — #318 is more complete (recursive, and catches window/grain_to_date/join_to_timespine/fill_nulls_with too). Closing this one in favor of it.

One suggestion on top: split `offset_window`/`offset_to_grain` out of `NON_PORTABLE_METRIC_SEMANTICS` into their own issue type. They're the one case in that bucket that's a *silent wrong number* (two inputs collapse to identical SQL) rather than an honest, visible-from-the-metric's-shape limitation. Worth letting consumers gate on that specifically.

@fanruan-tuco happy to send that as a small follow-up on #318 if useful.

fanruan-tuco: It's really a good suggestion, I will apply it in https://github.com/apache/ossie/pull/318

PR #372, opened 2026-09-09
Correct ontology schema's $id field

Ostrzyciel:
## Summary

Change the `$id` of the ontology schema from `https://github.com/apache/ossie/core-spec/ossie-schema.json` to `https://github.com/apache/ossie/core-spec/ontology.json`.

ossie-schema.json is already taken by the main schema, which is also referenced in the ontology schema (`properties.ai_context.$ref`). This reference resolves incorrectly to the ontology schema (this JSON file), breaking validators. The `semantic_model` ref is broken in the same way.

## Related Issues

N/A

### Ontology
- [x] Ontology changes in `ontology/` are consistent with spec changes
- [x] New or modified terms are defined and documented

### Tests

### Compliance

### DIFF
--- a/ontology/ontology.json
+++ b/ontology/ontology.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/apache/ossie/core-spec/ossie-schema.json",
+  "$id": "https://github.com/apache/ossie/core-spec/ontology.json",
   "title": "Apache Ossie Ontology Metadata Specification",
   "description": "JSON Schema for validating Apache Ossie ontology definitions",
   "type": "object",
--- a/ontology/ontology.json
+++ b/ontology/ontology.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/apache/ossie/core-spec/ontology.json",
+  "$id": "https://github.com/apache/ossie/ontology/ontology.json",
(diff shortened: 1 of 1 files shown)


ontology/ontology.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/apache/ossie/core-spec/ossie-schema.json",
+  "$id": "https://github.com/apache/ossie/core-spec/ontology.json",
```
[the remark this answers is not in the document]
Ostrzyciel: Can I get a review from a human?

jbonofre: Very good catch!

What you do actually fix a real `$id` collision with `core-spec/ossie-schema.json`, which was causing the `ai_context`/`semantic_model` `$ref`s below to be ambiguous for any tool that loads both schemas into one registry.

Just a minor comment (nit): the new `$id` uses `core-spec/ontology.json`, but this file actually lives at `ontology/ontology.json`. `ossie-schema.json` `$id` mirrors its real path (`core-spec/ossie-schema.json`), so for consistency this might read better as:

```
"$id": "https://github.com/apache/ossie/ontology/ontology.json",
```

To be fair, it's clearly not required since `$id` here is just an opaque identifier rather than a fetchable URL, but it would match the repo's existing convention 😄

jbonofre (dismissed):

ontology/ontology.json
Ostrzyciel: Ah, fair point, thanks. Fixed :)

jbonofre (approved):

PR #373, opened 2026-09-09
fix(validation): classify warnings independently of message text

flyrain:
This fix marks warnings explicitly by using a typed error instead of a text prefix, so user-provided text cannot disguise errors as warnings. It adds regression tests and includes SQLGlot in CI to cover SQL errors.

## Validation

Seven regression cases failed before the fix and pass afterward. All 54 validator tests pass locally on Python 3.14, and the canonical example validates successfully:

- `uv run validation/test_validate.py` — 34 passed
- `uv run --with pytest --with pyyaml --with jsonschema --with sqlglot -m pytest validation/tests/` — 20 passed
- `uv run validation/validate.py examples/tpcds_semantic_model.yaml` — passed
- `git diff --check` — passed

### Validation

### Tests
- [x] All existing validator tests pass locally

### Compliance
- [x] ASF license headers are preserved
- [x] Only existing third-party dependencies are used

### DIFF
--- a/.github/workflows/validation-ci.yml
+++ b/.github/workflows/validation-ci.yml
@@ -60,7 +60,7 @@ jobs:
         run: uv run validation/test_validate.py
 
       - name: Run validation pytest tests
-        run: uv run --with pytest --with pyyaml --with jsonschema -m pytest validation/tests/
+        run: uv run --with pytest --with pyyaml --with jsonschema --with sqlglot -m pytest validation/tests/
 
       - name: Validate canonical example
         run: uv run validation/validate.py examples/tpcds_semantic_model.yaml
--- a/validation/tests/test_validate.py
+++ b/validation/tests/test_validate.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import json
 from importlib.util import module_from_spec, spec_from_file_location
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


kayemkim (commented):
Ran this locally against a merge with current main (28365cd) on Python 3.11 through 3.14, following the steps in validation-ci.yml: the duplicate-key tests, the pytest suite with sqlglot, and the canonical example. All green, and the sqlglot test executed on every version rather than being skipped.

I also reproduced the bug on main. A relationship pointing at an unknown dataset named `Warning: customers` currently prints the reference error and then `Validation PASSED` with exit 0, because severity was decided by searching the message text. With this patch it fails with exit 1. The key-coverage warning from #330 (which I wrote) stays nonfatal, and the case where a genuine warning and a real error appear together still exits 1.

Validating every semantic-model YAML in the repo (examples and converter fixtures, 11 files) gives identical results before and after, so nothing that passes today changes status.

One small observation, not a request: the `Warning:` prefix in the two messages is now purely cosmetic, since severity comes from the type. Keeping it seems right for the humans reading the output.

LGTM.

MonkeyCanCode (dismissed):

MonkeyCanCode:
@flyrain mind fix the conflict when you get a chance?

flyrain:
Rebased, PTAL, @MonkeyCanCode.

jbonofre (approved):
LGTM.

During the review, I found other issues (unrelated to this PR), so I created #408 as follow up.

PR #377, opened 2026-09-10
fix(snowflake): recognise query sources with leading comments

### LINKED ISSUE #376: Snowflake converter turns a query source with a leading comment into a fake table
If a dataset's `source` is a SQL query that starts with a comment, the Snowflake converter doesn't notice it's a query. It splits the text on dots, uppercases the pieces, and emits them as a physical table. No error, no warning.

Here's the smallest example. This is valid Ossie; the spec says `source` can be a table name or a query.

```yaml
- name: orders
  source: |
    -- revenue source
    SELECT amount FROM db.schema.orders
```

Running it through `convert_ossie_to_snowflake` gives:

```yaml
base_table:
  database: "-- REVENUE SOURCE\nSELECT AMOUNT FROM DB"
  schema: SCHEMA
  table: ORDERS
```

The expected output is `base_table: {definition: <the query>}`, which is what Snowflake documents for query-backed tables.

The same thing happens with a `/* */` comment, with a query wrapped in parentheses, with `SELECT*FROM` (no space), and with `\r\n` after `SELECT` or `WITH`. Reproduced on `main` at 50457d3.

**Why**

`_parse_source` in `converters/snowflake/src/ossie_snowflake/converter.py` decides "query or table" with one check:

```python
upper.startswith(("SELECT ", "SELECT\n", "SELECT\t", "WITH ", "WITH\n", "WITH\t"))
```

Anything that doesn't match is assumed to be a table name. That path accepts any three dot-separated chunks without checking they're identifiers, so SQL text gets chopped up and emitted as `database.schema.table`.

**Where else**

The Honeydew converter has the same check and labels the query as a table. The NVIDIA converter has it too; its "must be a physical table" guard is bypassed by the comment, and it emits the same fake table. The Omni converter uses a word-boundary regex and raises a clear error instead, which is the right behavior when unsure.

**Proposed fix**

Two small changes, per converter:

1. Skip leading whitespace, `--` and `/* */` comments, and opening parentheses before looking for `SELECT` or `WITH`, and match the keyword as a whole word. Keep the query text exactly as written.
2. On the table path, require each of the three parts to be a real identifier (plain or double-quoted). If not, raise instead of uppercasing it.

This is independent of the structured `source` work in #109 / #173 / #338, which all keep string sources supported.

I have a patch for the Snowflake converter with tests (all existing tests pass, 18 added) and can follow up for Honeydew and NVIDIA. Happy to open the PR if this sounds right.

barveP:
## Summary

`_parse_source` in the Snowflake converter decided "query or table" with a fixed prefix check (`SELECT `, `SELECT\n`, `SELECT\t`, `WITH ` …). A query that starts with a `--`, `//` or `/* */` comment, uses `\r\n` after the keyword, is wrapped in parentheses, or has no whitespace after `SELECT` fell through to the relation path, which split the text on dots and emitted an uppercased `database/schema/table` with no warning:

```yaml
source: |
  -- revenue source
  SELECT amount FROM db.schema.orders
```

became `base_table: {database: "-- REVENUE SOURCE\nSELECT AMOUNT FROM DB", schema: SCHEMA, table: ORDERS}`.

This PR:

- adds `_is_query_source`, which skips leading whitespace, SQL comments and opening parentheses, then matches `SELECT`/`WITH` only when the next character cannot continue an unquoted identifier. `SELECT_RESULTS` and `SELECT$ARCHIVE` stay relations; `SELECT*FROM`, `SELECT/*c*/` and CRLF after the keyword are queries. Query text is emitted verbatim as `base_table.definition`, which Snowflake documents for query-backed tables.
- makes the relation path require three valid quoted or unquoted identifiers and raise `OssieConversionError` otherwise, so unrecognised text can no longer become a database name.

No behaviour change for existing physical-table or plain-query inputs: all 107 existing tests pass unchanged, and 23 tests are added (query shapes, relation-shaped garbage rejection, a keyword-prefixed table name and a `$` identifier as controls, and one end-to-end conversion). The change applies cleanly alongside #338. The Honeydew and NVIDIA converters have the same prefix check and will get follow-up PRs referencing #376.

## Related Issues

Fixes #376

### Specification

### Ontology

### Converters

### Validation

### Documentation

### Examples

### Tests

### Compliance

### DIFF
--- a/converters/snowflake/src/ossie_snowflake/converter.py
+++ b/converters/snowflake/src/ossie_snowflake/converter.py
@@ -25,6 +25,7 @@ Usage:
 """
 
 import argparse
+import re
 import sys
 import warnings
 
@@ -421,6 +422,30 @@ def _normalize_identifier(identifier):
         return stripped
     return stripped.upper()
 
+# A dataset source is a SQL query when, after leading whitespace and SQL
+# comments (`-- ...` and `/* ... */`) and any opening parentheses, it starts
+# with SELECT or WITH as a whole word. The word boundary keeps table names such
+# as SELECT_RESULTS on the relation path, while `SELECT*FROM`, `SELECT/*c*/`,
+# and CRLF line endings after the keyword are still recognised as queries.
+_LEADING_SQL_TRIVIA = re.compile(r"^(?:\s+|--[^\n]*(?:\n|$)|/\*.*?\*/)+", re.DOTALL)
+_QUERY_KEYWORD = re.compile(r"(?:SELECT|WITH)\b", re.IGNORECASE)
+_UNQUOTED_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_$]*$")
+_QUOTED_IDENTIFIER = re.compile(r'^"(?:[^"]|"")+"$')
+
+
+def _is_query_source(source_stripped):
+    """True if the source text is a SQL query rather than a relation name."""
+    body = _LEADING_SQL_TRIVIA.sub("", source_stripped, count=1)
+    while body.startswith("("):
+        body = _LEADING_SQL_TRIVIA.sub("", body[1:], count=1)
+    return _QUERY_KEYWORD.match(body) is not None
+
+
+def _is_identifier(part):
+    """True if `part` is a valid quoted or unquoted Snowflake identifier."""
+    return bool(_UNQUOTED_IDENTIFIER.match(part) or _QUOTED_IDENTIFIER.match(part))
+
+
 def _split_identifiers(source_str):
     """Split a dot-separated identifier string while respecting double quotes."""
     parts = []
@@ -451,15 +476,16 @@ def _parse_source(source):
     if not source_stripped:
         return None
 
-    # Detect subqueries — require whitespace after the keyword to avoid false
-    # positives on table names like WITH_TABLE or SELECT_RESULTS.
-    upper = source_stripped.upper()
-    if upper.startswith(("SELECT ", "SELECT\n", "SELECT\t",
-                          "WITH ", "WITH\n", "WITH\t")):
+    # Queries keep their exact text (including leading comments) so the
+    # emitted definition is what the author wrote.
+    if _is_query_source(source_stripped):
         return {"definition": source_stripped}
 
+    # Anything else must be a relation name whose three parts are real
+    # identifiers; otherwise SQL text that was not recognised as a query would
+    # silently become a bogus database/schema/table.
     parts = _split_identifiers(source_stripped)
-    if len(parts) == 3:
+    if len(parts) == 3 and all(_is_identifier(part) for part in parts):
         # Only uppercase unquoted identifiers; preserve quoted ones as-is.
         return {
             "database": _normalize_identifier(parts[0]),
@@ -468,7 +494,8 @@ def _parse_source(source):
         }
 
     raise OssieConversionError(
-        f"Source '{source}' must be a fully qualified db.schema.table or a subquery"
+        f"Source '{source}' must be a fully qualified db.schema.table "
+        "(quoted or unquoted identifiers) or a SELECT/WITH query"
     )
 
 
--- a/converters/snowflake/src/ossie_snowflake/converter.py
+++ b/converters/snowflake/src/ossie_snowflake/converter.py
@@ -422,13 +422,17 @@ def _normalize_identifier(identifier):
         return stripped
     return stripped.upper()
 
-# A dataset source is a SQL query when, after leading whitespace and SQL
-# comments (`-- ...` and `/* ... */`) and any opening parentheses, it starts
-# with SELECT or WITH as a whole word. The word boundary keeps table names such
-# as SELECT_RESULTS on the relation path, while `SELECT*FROM`, `SELECT/*c*/`,
-# and CRLF line endings after the keyword are still recognised as queries.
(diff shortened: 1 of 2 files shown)


converters/snowflake/src/ossie_snowflake/converter.py
```diff
@@ -421,6 +422,34 @@ def _normalize_identifier(identifier):
         return stripped
     return stripped.upper()
 
+# A dataset source is a SQL query when, after leading whitespace, SQL comments
+# (`-- ...`, `// ...` and `/* ... */`) and any opening parentheses, it starts
+# with SELECT or WITH followed by something that cannot continue an unquoted
```
mattsenicksigma: NIT: can probably just condense this comment to a single line

converters/snowflake/src/ossie_snowflake/converter.py
```diff
@@ -421,6 +422,34 @@ def _normalize_identifier(identifier):
...
+# identifier. That keeps names such as SELECT_RESULTS or SELECT$ARCHIVE on the
+# relation path (Snowflake allows `$` in unquoted identifiers, so `\b` would be
+# wrong), while `SELECT*FROM`, `SELECT/*c*/` and CRLF after the keyword are
+# still recognised as queries.
+_LEADING_SQL_TRIVIA = re.compile(
+    r"^(?:\s+|--[^\n]*(?:\n|$)|//[^\n]*(?:\n|$)|/\*.*?\*/)+", re.DOTALL
+)
+_QUERY_KEYWORD = re.compile(r"^(?:SELECT|WITH)(?![A-Za-z0-9_$])", re.IGNORECASE)
+_UNQUOTED_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_$]*$")
+_QUOTED_IDENTIFIER = re.compile(r'^"(?:[^"]|"")+"$')
```
christianeu-db, line 437: Probably a question for @khush-bhatia but I see sqlglot used in other parts of Ossie & it supports the Snowflake dialect. I wonder if directly parsing would be a bit easier for handling edge cases when compared to regex

barveP: Thanks, that could make this simpler. Happy to try SQLGlot here. 
@khush-bhatia, what do you think?

barveP: tried sqlglot and pushed an update. lmk what you think.

converters/snowflake/src/ossie_snowflake/converter.py
```diff
@@ -421,6 +425,30 @@ def _normalize_identifier(identifier):
         return stripped
     return stripped.upper()
 
+_UNQUOTED_IDENTIFIER = re.compile(r"^[A-Za-z_][A-Za-z0-9_$]*$")
+_QUOTED_IDENTIFIER = re.compile(r'^"(?:[^"]|"")+"$')
```
christianeu-db, line 429: You might be able to avoid these regex as well with an approach like the following:
 
def try_parse_source_relation(source_stripped):
    try:
        table = parse_one(source_stripped, read="snowflake", into=exp.Table)
    except ParseError:
        return None
    if table is None or len(table.parts) != 3 or not all(isinstance(p, exp.Identifier) for p in table.parts):
      return None

   return {
    "database": _render_identifier(table.parts[0]),
     "schema": _render_identifier(table.parts[1]),
       "table": _render_identifier(table.parts[2]),
   }

There is one nuance that the parse_one might be permissive to trailing semi-colons (so that might be a case to check in try_parse_source_relation => return None).

This helps keep the control flow a little cleaner:

if is_source(source_stripped):
  return ...

parsed_relation = try_parse_relation(source_stripped)
if parsed_relation is not None:
   return parsed_relation

throw_error


(ideas from this [blog](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) on the parse/don't validate pattern)

christianeu-db (commented):
Looking a lot simpler - thanks for taking the feedback! I left one other comment (with some pseudocode) about the remaining regex could be removed / the parsing logic could be consolidated.

once that's in, it's worth re-pinging one of the committers (esp. one with snowflake expertise). I like the encapsulation of _is_query_source that you added - this makes the query detection logic easier to extend in the futrue

converters/snowflake/src/ossie_snowflake/converter.py
barveP: thanks. took the control-flow shape in the latest commit.

i also tried `parse_one(..., into=exp.Table)` for the relation path. it works with four guards: trailing `;` (as you noted), comments, extra args like an alias or `AT (OFFSET => ...)` that would otherwise be dropped silently, and `isinstance(exp.Table)`, since `a.b.c; d.e.f` parses to a Block. with those it passes every test except `from.public.orders` and `db.public.qualify`

those two are the real question. sqlglot decides which keywords can be identifiers from its own `ID_VAR_TOKENS` list ([parser doc](https://github.com/tobymao/sqlglot/blob/main/sqlglot/parser.py)), so it rejects from and `qualify` but accepts `order`, `table`, `group`, which snowflake reserves too ([reserved-keywords](https://docs.snowflake.com/en/sql-reference/reserved-keywords)). the regexes are snowflake's identifier grammar verbatim ([identifiers](https://docs.snowflake.com/en/sql-reference/identifiers-syntax)): they check shape only and leave reserved words to snowflake

so it's partial enforcement via sqlglot, or none. i lean none because it's consistent but i'm fine either way and have the `parse_one` version ready (with a bump to sqlglot 30.13.0, since 30.12.0 misparses `"@".public.orders`).

which do we prefer?

cc: @khush-bhatia since you know the snowflake side best

Issue #111, opened 2026-04-20
Follow up on OSI ai_context and custom_extensions mapping in snowflake yaml


sfc-gh-tgao:
As discussed in the initial [PR](https://github.com/open-semantic-interchange/OSI/pull/98) for OSI->Snowflake YAML converter, there are some concepts in the OSI ai_context and custom_extensions that do not have a direct mapping in Snowflake YAML (e.g., field-level instruction and examples, ai_context on relationships, etc.). We should follow up with the Snowflake team on how we should add support and properly convert these in Snowflake YAML so that there is no metadata loss.

PR #378, opened 2026-09-10
[DBT] Map config.meta onto Ossie custom_extensions instead of dropping it

### LINKED ISSUE #303: msi_to_osi silently discards config.meta instead of mapping it to custom_extensions
`MSIToOSIConverter` silently discards `config.meta` from metrics, dimensions and semantic models. MSI carries the metadata, Ossie has a place to put it (`custom_extensions`), but the converter never reads one or writes the other — and unlike other lossy paths in this converter, no `ConverterIssue` is emitted.

**Root cause**

`converters/dbt/src/ossie_dbt/msi_to_osi.py`. Each OSI object is built from a fixed handful of attributes:

```python
OSIMetric(
    name=metric.name,
    expression=self._make_expression(expr),
    description=metric.description,
)
```

```python
return OSIField(
    name=dim.name,
    expression=self._make_expression(expr),
    dimension=OSIDimension(is_time=is_time),
    label=dim.label,
    description=dim.description,
)
```

`grep -n '\.config\|meta' converters/dbt/src/ossie_dbt/msi_to_osi.py` returns nothing — `.config` is never accessed anywhere in the converter.

Both sides of the mapping already exist:

- **Source:** `PydanticMetric`, `PydanticSemanticModel` and `PydanticDimension` all expose a `config` field of type `PydanticSemanticLayerElementConfig`, whose `meta` is an arbitrary dict. dbt writes user-supplied `config.meta` into `semantic_manifest.json`, so it is present in the object the converter is handed.
- **Destination:** `custom_extensions` is defined on `Field`, `Metric`, `Dataset`, `Relationship` and `SemanticModel` in `core-spec/osi-schema.json`, and `Vendor` is documented as "Any string value is accepted", with `DBT` given as an example.

```python
import sys

from tests.helpers import _entity, _manifest, _simple_metric
from metricflow_semantic_interfaces.implementations.element_config import PydanticSemanticLayerElementConfig
from metricflow_semantic_interfaces.implementations.elements.dimension import PydanticDimension
from metricflow_semantic_interfaces.implementations.elements.measure import PydanticMeasure
from metricflow_semantic_interfaces.test_utils import semantic_model_with_guaranteed_meta
from metricflow_semantic_interfaces.type_enums import AggregationType, DimensionType, EntityType

meta = {"value_format": '[>=1000000]$#,##0.00,,"M";$#,##0.00', "group_label": "Booking Costs"}

sm = semantic_model_with_guaranteed_meta(
    name="bookings",
    entities=[_entity("booking", EntityType.PRIMARY, "booking_id")],
    measures=[PydanticMeasure(name="amount", agg=AggregationType.SUM, expr="amount_eur")],
    dimensions=[PydanticDimension(name="status", type=DimensionType.CATEGORICAL,
                                  config=PydanticSemanticLayerElementConfig(meta=meta))],
)
metric = _simple_metric("total_amount", "amount")
metric.config = PydanticSemanticLayerElementConfig(meta=meta)

res = MSIToOSIConverter().convert(_manifest(semantic_models=[sm], metrics=[metric]))
osi_metric = res.output.semantic_model[0].metrics[0]
osi_field = [f for f in res.output.semantic_model[0].datasets[0].fields if f.name == "status"][0]

print("metric.custom_extensions:", osi_metric.custom_extensions)
print("field.custom_extensions :", osi_field.custom_extensions)
print("ISSUES:", res.issues)
```

Run from `converters/dbt/`.

Observed:

```
metric.custom_extensions: None
field.custom_extensions : None
ISSUES: []
```

Expected: the `meta` dict carried through as a `custom_extensions` entry — or, if that is deliberately out of scope, a `ConverterIssue` so the loss is visible.

**Why this matters**

`config.meta` is the only place dbt users can attach metadata that the MSI spec does not model — units, display formatting, grouping, ownership, links back to a source system. For anyone migrating an existing BI semantic layer into dbt it tends to be where all the presentation metadata ends up, precisely because the spec has nowhere else for it. Discarding it at the Ossie boundary means that content cannot reach any downstream consumer, even ones that would understand it.

It is also inconsistent with how this converter handles its other lossy paths: `CUMULATIVE` metrics emit `CUMULATIVE_SEMANTICS_LOSS`, conversion metrics emit `CONVERSION_METRIC_DROPPED`, natural entities emit `NATURAL_ENTITY_DROPPED`. This loss is silent.

**Suggested fix**

Map `config.meta` onto a single `custom_extensions` entry when non-empty, e.g.

```python
def _meta_extensions(config) -> Optional[List[OSICustomExtension]]:
    meta = getattr(config, "meta", None) if config else None
    if not meta:
        return None
    return [OSICustomExtension(vendor_name="DBT", data=json.dumps(meta))]
```

applied in `_convert_metrics`, `_convert_dimension`, `_convert_entity`, `_convert_measure` and `_convert_dataset`. `CustomExtension.data` is a JSON string, so the dict serialises directly.

Happy to send a PR if the approach sounds right. Two things worth deciding first:

1. **Vendor name.** `DBT` matches the schema's own example, but `COMMON` may be more appropriate if the intent is that any consumer may read it rather than dbt-specific tooling.
2. **Round-tripping.** `osi_to_msi` would need the inverse to make this lossless in both directions. Happy to include that in the same PR.

**Note:** the same `msi_to_osi` implementation is vendored in `dbt-labs/metricflow` as `metricflow/converters/msi_to_osi.py` (byte-identical apart from one word in a docstring) and ships inside dbt-core 1.12, where it generates `target/osi_document.json`. The same gap exists there.

Related: #111 covers `custom_extensions` being dropped again on the Snowflake side, so both would need addressing for metadata to survive a full dbt → Ossie → Snowflake conversion.

georgea-tk:
## Summary

`config.meta` is silently dropped when converting an MSI manifest to Ossie — dimensions, entities, measures, metrics, and semantic models can all carry arbitrary presentation/vendor metadata (e.g. a LookML `group_label` or `value_format`), and none of it survives the conversion. Both ends of the mapping already exist: MSI exposes `config.meta` as a dict, and Ossie's `Field`/`Metric`/`Dataset`/`SemanticModel` all define `custom_extensions` with an open `vendor_name` string.

This maps a non-empty `config.meta` onto one `custom_extensions` entry (`vendor_name="DBT"`, `data=json.dumps(meta)`) wherever it's currently dropped. Empty/absent `meta` still produces no `custom_extensions`.

Two things intentionally left out of scope, per the issue:
- Round-tripping through `ossie_to_msi` — open question in the issue on whether that's wanted.
- The equivalent drop in `converters/snowflake` (#111) — a separate PR, since fixing either alone doesn't get metadata all the way through.

`vendor_name="DBT"` follows the schema's own example; happy to change it if maintainers prefer `COMMON` or something else.

## Related Issues

Fixes #303

### Specification
- [ ] N/A — no spec change

### Ontology
- [ ] N/A

### Converters
- [x] New tests added under the converter's test directory

### Validation
- [ ] N/A

### Documentation
- [ ] N/A — internal mapping, no user-facing doc surface changed

### Examples
- [ ] N/A

### Tests
- [x] All existing tests pass — ran full `converters/dbt` suite locally, 106 passed
- [x] New functionality is covered by tests — dimension/entity/measure/metric/semantic-model config.meta, plus empty and absent meta

### Compliance
- [x] ASF license headers present — no new files, only edits to existing licensed files

### DIFF
--- a/converters/dbt/src/ossie_dbt/msi_to_ossie.py
+++ b/converters/dbt/src/ossie_dbt/msi_to_ossie.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import json
 import re
 from collections import defaultdict
 from dataclasses import dataclass
@@ -22,6 +23,7 @@ from itertools import combinations
 from typing import Dict, List, Optional, Sequence, Tuple
 
 from ossie import (
+    OssieCustomExtension,
     OssieDataset,
     OssieDialect,
     OssieDialectExpression,
@@ -32,6 +34,7 @@ from ossie import (
     OssieMetric,
     OssieRelationship,
     OssieSemanticModel,
+    OssieVendor,
 )
 from ossie_dbt.converter_issues import ConverterIssue, ConverterIssueType, ConverterResult
 from ossie_dbt.filter_utils import _collect_filter_sql, _merge_filter_sqls
@@ -44,6 +47,7 @@ from metricflow_semantic_interfaces.protocols.measure import (
     Measure,
     MeasureAggregationParameters,
 )
+from metricflow_semantic_interfaces.protocols.meta import SemanticLayerElementConfig
 from metricflow_semantic_interfaces.protocols.metric import Metric
 from metricflow_semantic_interfaces.protocols.semantic_model import SemanticModel
 from metricflow_semantic_interfaces.transformations.convert_count import ConvertCountMetricToSumRule
@@ -116,6 +120,7 @@ class MSIToOssieConverter:
                     name=metric.name,
                     expression=self._make_expression(expr),
                     description=metric.description,
+                    custom_extensions=self._custom_extensions_from_meta(metric.config),
                 )
             )
 
@@ -153,6 +158,7 @@ class MSIToOssieConverter:
             unique_keys=unique_keys if unique_keys else None,
             description=sm.description,
             fields=fields if fields else None,
+            custom_extensions=self._custom_extensions_from_meta(sm.config),
         )
 
     def _convert_dimension(self, dim: Dimension) -> OssieField:
@@ -165,6 +171,7 @@ class MSIToOssieConverter:
             dimension=OssieDimension(is_time=is_time),
             label=dim.label,
             description=dim.description,
+            custom_extensions=self._custom_extensions_from_meta(dim.config),
         )
 
     def _convert_entity(self, entity: Entity) -> OssieField:
@@ -175,6 +182,7 @@ class MSIToOssieConverter:
             expression=self._make_expression(expr),
             label=entity.label,
             description=entity.description,
+            custom_extensions=self._custom_extensions_from_meta(entity.config),
         )
 
     def _convert_measure(self, measure: Measure) -> OssieField:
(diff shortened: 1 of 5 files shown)


QMalcolm (commented):
Thank you for putting this together! What has been done thus far, looks good 🙂 The PR body mentions that
> Round-tripping through ossie_to_msi — open question in the issue on whether that's wanted

I think round-tripping would be desirable. I'd appreciate that being added before approval ❤️ If you disagree, let me know, and we can discuss it more and perhaps proceed as is.

ctumlin:
We run dbt Cloud, and 72 of our 89 MetricFlow metrics carry config.meta. We store an access tier there, and our semantic layer proxy reads it to enforce role-based access control over who can list and query each metric. Losing it on export means any downstream tool receives our metrics with no access classification, so it can't apply the same RBAC rules. I ran msi-to-ossie on current main against our real manifest and confirmed the output contains no custom_extensions at all. This PR fixes that for us. It looks like it now conflicts with the flat document migration in #396. I'm glad to test a rebased version against our manifest.

PR #375, opened 2026-09-10
fix(validation): enforce equal arity for relationship column arrays

christianeu-db:
## Summary

Enforce the spec's already-documented requirement that a relationship's `from_columns` and `to_columns` have the same number of columns.

`$defs/Relationship` constrains each array with `minItems: 1` but not their relative length, so a document with mismatched arity — describing an incomplete tuple join (e.g. `from_columns: [a, b]` with `to_columns: [c]`) — validated successfully. `core-spec/spec.md` ("Relationships") already states the two arrays must correspond positionally and have the same number of columns; JSON Schema cannot express this cross-array equality, so the rule is enforced in `validation/validate.py`:

- New `validate_relationship_column_arity` check, alongside the existing endpoint-validation rules (`validate_references`), emitting an `[Arity]` error per mismatched relationship. It skips documents that already fail schema validation.

Covered by tests in `validation/tests/test_validate.py` (equal-length simple and composite pass; both mismatch directions rejected; non-list `to_columns` skipped without crashing).

No spec or JSON Schema change is needed — this only closes the enforcement gap.

## Related Issues

Addresses Problem 6 (Relationship Column Array Arity) from https://github.com/apache/ossie/discussions/374

### Validation
- [x] Validation rules in `validation/` are updated
- [x] New validation cases are covered by tests

### Tests
- [x] New functionality is covered by tests

### Compliance
- [x] No new source files are added (existing ASF license headers unchanged)
- [x] No third-party dependencies are added

_Documentation and examples: no change — the spec already documents the constraint and the existing examples already conform._

PR #383, opened 2026-09-11
Define one semantic model per document without a wrapper

flyrain:
## Summary

Change the core JSON/YAML specification so each standalone document contains exactly one semantic model, with model properties directly at the root. Remove both the `semantic_model` array and its wrapper. For example:

```yaml
version: 0.2.0.dev0
name: sales_analytics
datasets:
  - name: orders
    source: sales.public.orders
```

A model still supports multiple datasets, relationships, and metrics. Bulk exchange uses separate documents; this PR does not introduce a bundle format.

**Breaking change to the unreleased, mutable `0.2.0.dev0` schema:** wrapped models (arrays or objects) are rejected. Migration moves a single model's properties to the root or splits multiple models into separate documents, preserving model contents and using the current specification version. Root-level `dialects` and `vendors` declarations remain removed, as established by #397.

## Changes

- Update the core JSON Schema, Markdown specification, YAML reference, and documentation overview.
- Flatten the canonical TPC-DS example without changing model contents.
- Update validator traversal and add focused checks for the new shape and rejection of legacy/malformed input.
- Preserve `$defs/SemanticModel` and the ontology map's embedded `semantic_model` property.

**Scope and merge coordination:** The Python SDK and converter migration is implemented in [PR #396](https://github.com/apache/ossie/pull/396), which also updates converter CI to run on shared schema/example changes. Review the spec and migration separately, but do not land this schema change until #396 is approved and its combined spec/consumer CI is green. Coordinate the two merges so the repository is not left with the new schema and old consumers. This PR alone does not migrate those consumers.

## Applicable checklist

- [x] Specification and reference examples updated in `core-spec/`.
- [x] Breaking change and migration documented.
- [x] Validator behavior and focused regression coverage updated.
- [x] Ontology embedded model definition preserved.
- [x] No dependencies added.
- [ ] Dev mailing list discussion (pending).
- [ ] Python SDK and converter adoption: [#396](https://github.com/apache/ossie/pull/396), required for coordinated landing.


validation/validate.py
```diff
@@ -340,8 +346,9 @@ def main():
     errors = []
     errors.extend(validate_schema(data, schema))
 
-    # Run semantic-model-specific checks only for semantic model payloads.
-    if data.get("semantic_model"):
+    # Semantic checks rely on valid structure; let schema validation report
+    # malformed inputs (including legacy arrays) without traversing them.
+    if not errors and isinstance(data, dict) and "datasets" in data:
```
khush-bhatia, line 351: Could you create an issue to fix this in a follow up ?

khush-bhatia (dismissed):

validation/validate.py
flyrain: This guard prevents semantic checks from crashing on structurally invalid input. We already have #270 tracking that issue and #271 addressing it. Does that cover the follow-up you had in mind, or were you referring to something else in this condition?

core-spec/spec.md
```diff
@@ -83,12 +83,24 @@ ontology specification's built-in value types; `Time`, `DateTimeTz`, and
...
+or archive. This specification does not define a bundle format or cross-model
+references.
 
 ### Schema
 
 | Field | Type | Required | Description |
 |-------|------|----------|-------------|
+| `version` | string | Yes | Apache Ossie specification version (`0.2.0.dev0`) |
+| `dialects` | array | No | Expression dialects used in the document |
+| `vendors` | array | No | Vendors with custom extensions in the document |
```
khush-bhatia: `vendors` and `dialects`  should be removed.

core-spec/spec.md
```diff
@@ -83,12 +83,24 @@ ontology specification's built-in value types; `Time`, `DateTimeTz`, and
 
 ## Semantic Model
 
-The top-level container that represents a complete semantic model, including datasets, relationships, and  metrics.
+Each JSON or YAML document represents exactly one semantic model. Model properties
+are defined directly at the document root alongside `version`, `dialects`, and
+`vendors`; there is no `semantic_model` wrapper. A model can contain multiple
+datasets, relationships, and metrics.
```
khush-bhatia: Please remove `dialects`, and `vendors` from here.

core-spec/spec.md
```diff
@@ -100,21 +112,56 @@ The top-level container that represents a complete semantic model, including dat
...
+declarations. Use `version: 0.2.0.dev0` for the migrated document. For multiple
+models, create one document per model, carrying over the applicable declarations
+and validating each result. An empty model array cannot produce a valid model
+document. Preserve model contents and custom extensions; never silently select
+only the first model or overwrite a file when splitting a document.
+
+The reusable `$defs/SemanticModel` schema still describes model contents without
+standalone document metadata. In particular, an ontology map continues to embed
+those contents under its `semantic_model` property. This standalone document
+change does not rename or flatten that ontology property.
```
khush-bhatia, line 154: Instead we should update the ontology spec to point to the root of this spec(perhaps "https://github.com/apache/ossie/core-spec/ossie-schema.json")  ? 

Can be done in a follow up PR though.

flyrain: Agreed, let’s handle that in a follow-up PR. Referencing the root schema would also require version on each embedded semantic model, so we should update the ontology examples, converters, and tests together. I’ll leave the current reference in place for this PR.

djwaldo:
Thanks for clarification.

khush-bhatia (dismissed):

kayemkim:
Merging this branch (6470653) with current main (fc6c9df): #375 landed yesterday, after the last push here, and its `validate_relationship_column_arity` still iterates `data.get("semantic_model", [])`. For a flat document that loop is empty, so the arity check never runs. A relationship with `from_columns: [customer_id, region_id]` and `to_columns: [id]` prints `Validation PASSED` on the merged tree, where the wrapper form fails on main with `[Arity]`. The two `test_arity_rejects_mismatched_length_columns` cases fail the same way, so Validation CI would go red on main after merge. There is no textual conflict and the PR's runs predate #375, which is why neither GitHub nor CI shows it.

Same fix this PR applies to the other three validators: drop the `for model in` loop and read `relationships` from the root. The other 37 pytest cases and the two remaining CI steps pass on the merged tree.

flyrain:
Thanks @kayemkim , fixed it in the new commit. I'd suggest to merge this to avoid further rebase.

khush-bhatia (approved):

Issue #361, opened 2026-09-04
Ontology expression language: semantics and interoperability considerations


FactEngineCommunity:
From the current ontology specification and the expression-language work underway, it appears that portable ontology-level expressions for derived_by and requires are intended to form part of Ossie's interchange capabilities.

Assuming that is the direction being taken, I think there are some consequences for the ontology specification that are worth identifying explicitly.

I am raising these partly from an implementation perspective. I am currently working with Ossie ontology interchange, including conversion to and from other conceptual modelling formats, and expressions are one area where independent implementations will need sufficiently precise semantics to produce equivalent results.

## Semantics of expression lists

The current ontology specification defines both derived_by and requires as lists of expressions.

For example, the ontology documentation includes:

derived_by:
  - "Person.parent_of(descendant)"
  - "Person.ancestor_of.parent_of(descendant)"

These are described as the base and recursive cases for deriving ancestor_of.

There are also examples such as:

derived_by:
  - "Person.files_single AND Person.earns <= 11925 AND TaxRate == 10.0"
  - "Person.files_married_joint AND Person.earns <= 23850 AND TaxRate == 10.0"

My reading of these examples is that individual entries in derived_by represent alternative derivation rules. In effect, the results of the individual expressions are combined:

expression-1
OR
expression-2
OR
expression-3

or, perhaps more precisely, the population produced by derived_by is the union of the populations produced by its individual expressions.

If that interpretation is correct, I think it would be useful for this to be stated normatively.

If it is not correct, that may itself demonstrate an interoperability issue: two reasonable implementations could currently interpret the same YAML differently.

The same question arises for requires.

My current reading is that multiple requires expressions are conjunctive: all requirements must hold.

If so, we effectively have:

derived_by list -> OR / union semantics
requires list   -> AND semantics

There is nothing inherently wrong with those semantics, but they should probably be explicit because the YAML representation itself does not convey the distinction.

Boolean composition within an expression

There is then a second level of composition within each expression.

Simple cases are straightforward:

A AND B AND C

Real derivations can, however, require expressions such as:

A AND (B OR C)

or:

(A OR B) AND NOT C

At this point an ontology expression language needs Boolean operators, grouping and defined operator precedence.

This is particularly relevant because Boolean composition may occur at two levels:

derived_by:
    expression 1
    expression 2

where the expressions themselves may contain:

AND
OR
NOT
(...)

An implementation therefore needs precise semantics for both the expression and the containing ontology construct.

[...]

PR #389, opened 2026-09-13
Add LinkML <-> Ossie ontology converters

Ostrzyciel:
## Summary

**Little bit of context:** Hi, I'm Piotr, maintainer of one of LinkML implementations: [LinkML-Scala](https://github.com/NeverBlink-OSS/linkml-scala). We've been tracking Ossie's progress, and saw a good opportunity to contribute converters for LinkML, as this was already requested by the Ossie community. We'll be more active in the community in the coming weeks :) we plan to start attending the WG meetings.

This PR adds converters for LinkML -> Ossie ontology, and Ossie ontology -> LinkML.

- The converters use an external dependency, [`neverblink-linkml`](https://pypi.org/project/neverblink-linkml/).
  - This is a natively-compiled version of LinkML-Scala. No, it does **not** use Java or the JVM :) modern Scala can also compile natively.
  - The library is licensed on Apache 2.0. It has no transitive dependencies. The binary size is ~2–3 MB, depending on the platform.
  - The PyPI package has wheels for all popular platforms, and an sdist, so it should work anywhere.
- Why is the dependency needed?
  - LinkML is an very complex framework. Doing anything in LinkML requires you to resolve imports, merge schemas, perform class & attribute derivation using some rather complex algorithms and so on. LinkML devs call this resolved form "SchemaView". You can find our implementation [here](https://github.com/NeverBlink-OSS/linkml-scala/tree/main/schemaview/src/eu/neverblink/linkml/schemaview).
  - To implement a converter from LinkML you must look at this resolved form, not the plain YAML syntax.
  - There are currently 2 SchemaView implementations – one is ours, another written in Python is maintained mainly by bioinformatics people. The Python implementation [pulls in many transitive deps](https://github.com/linkml/linkml/blob/main/pyproject.toml), some which are apparently GPL-ed (?).
- Coverage of the mapping is documented here: https://github.com/NeverBlink-OSS/linkml-scala/blob/main/docs/ossie_mapping.md 
- You can try out these new converters here: https://linkml.neverblink.eu/playground/
  - This is a 100% in-browser playground. LinkML-Scala also compiles to JavaScript.
- Tests.
  - The tests here are minimal, because the implementation is already well-tested in LinkML-Scala – currently 409 tests.
  - Tests: [unit](https://github.com/NeverBlink-OSS/linkml-scala/tree/main/generator/test/src/eu/neverblink/linkml/generator/ossie), [integration](https://github.com/NeverBlink-OSS/linkml-scala/tree/main/generator/test/src-jvm/eu/neverblink/linkml/generator/ossie) (cross-check with Ossie JSON Schema and the Ossie validator). This includes round-trip testing, testing over a large catalog of edge cases, and fuzzing the expression parser.
- Maintenance.
  - We would prefer to maintain the code in our repository and here only keep up the wrapper, minimal tests, and docs relevant to the Ossie community.
  - I've added links in the code and docs to direct bug reports to our repository.
  - We are also open to alternative arrangements! If you have any ideas, please shoot.

**EDIT:** Link to source code of the converter: https://github.com/NeverBlink-OSS/linkml-scala/tree/86a05dc5b031dbed1fb093f83132af6bce9d8211/generator/src/eu/neverblink/linkml/generator/ossie

### Converters

### Documentation
- [x] `docs/` is updated to reflect any user-facing changes
- [x] New features or behaviors are documented with examples where appropriate

### Tests

### Compliance

**^ needs PMC approval!** This should be easy, though, because the only dependency is licensed on Apache 2.0.


kayemkim (commented):
Thanks for bringing this over, and for the mapping doc, which made checking it much easier. Ran this on top of current main: the ontology converter CI steps pass on 3.11 through 3.14, and the flights fixture goes Ossie -> LinkML -> Ossie with both directions exiting 0 and the result validating against `ontology/ontology.json`. On the dependency, the wheel's metadata says `License-Expression: Apache-2.0` with the LICENSE file included, and it declares no `Requires-Dist`, which matches the description; the approval itself is for the PMC.

One suggestion before this lands. The docstrings, README and `test_roundtrip_drops_ontology_mappings` say the only loss is `ontology_mappings`, but a structural diff of the flights round trip shows more, all without a line on stderr:

- all four `derived_by` (the `average_*_delay` relationships on Airport and Route) come back as plain attributes with `range: Delay`
- the document-level `requires` (`COUNT[Airport] > 0`, `COUNT[Carrier] > 0`) are gone
- the ternary `Flight.registers_latitude_series` loses its `DegreesLatitude` role and comes back binary; the unary `Flight.canceled` gains a `String` role
- the five relationship-level `requires` (`Flight.departs_at < Flight.arrives_at`, `NOT Route.destination(Airport)`, ...) are gone
- two concept-level `requires` are gone because the parser does not read their form: `CancelationCode == 'A' OR ... == 'D'` (only `IN (...)` is read) and `1 <= DistanceGroup` (`DistanceGroup <= 10` survives, the literal-on-the-left one does not)

Your mapping doc does say derived_by, document requires and extra roles are not emitted, so the library is behaving as documented. But someone reading this repo only sees the `ontology_mappings` caveat, and the round-trip tests compare names, types and `identify_by`, so they pass through all of the above. Two small additions would close that gap: list these losses next to `ontology_mappings` in the module docstring and README, and add round-trip tests in the style of `test_roundtrip_drops_ontology_mappings` for at least `derived_by` and n-ary roles. `palantir_to_ossie` in the same package uses `warnings.warn` when it drops something, so a warning per dropped construct would also fit here if the library can report them.

The expression point connects to #361: which `requires` forms are read is exactly the kind of thing a grammar would settle, and the repo's own flights fixture uses both the readable and the unreadable spellings. Until then a line in the README saying which forms the importer reads would help.

Nit: `scripts/linkml_to_ossie.py` lines 26-27 repeat the "loader reports are written to stderr" sentence, the original line was left in when the Copilot fix was applied.

jochenchrist:
In the tests, I would love to see a full example for both YAML structures, and the properties match.

jochenchrist:
Are there any elements that cannot be transpiled?

kayemkim (commented):
Thanks for the update. The README paragraph and the docstrings now say what is and is not carried across, and the duplicate line is gone. Re-ran on current main (fc6c9df): the ontology converter CI steps pass on 3.11 through 3.14 with no conflicts. On the warnings, your point about the two specs not being formal enough to enumerate losses is fair; the mapping doc's tables already do the job for anyone who looks. From my side this is good to go.

For the question above about what cannot be transpiled: the new README paragraph and the linked mapping doc cover it, with the readable `requires` forms (`Ref IN (...)`, `Ref >= v`, `Ref <= v`) listed under Expressions.

Ostrzyciel:
@jochenchrist addressed your comment, there are now 2 new tests.

I've also added a link to the source code in the PR description and in the README.

Can I get an approval for this now?

PR #396, opened 2026-09-14
Migrate converters and Python SDK to flat semantic model documents

flyrain:
## Summary

Migrate the Python SDK and remaining affected converters to the single-model document format merged in #383. Ossie input/output now puts `name`, `datasets`, `relationships`, and `metrics` directly at the root alongside `version`; the legacy `semantic_model` wrapper is rejected instead of selecting its first entry.

```yaml
version: 0.2.0.dev0
name: sales_analytics
datasets:
  - name: orders
    source: sales.public.orders
relationships: []
metrics: []
```

**Rebased onto main at `bc7b2c0` after #383 and #407 merged.** Microsoft/Power BI migration and its stricter dataset validation are already on main through #407. This PR now contains only the SDK/converter migration and its CI/documentation updates. It also incorporates #397: root-level `dialects` and `vendors` are removed; expression dialects and vendor custom extensions remain supported.

## Behavior

- Flatten the shared `OssieDocument` API and update dbt, Sigma, Wisdom, Databricks, Honeydew, Omni, OrionBelt, Snowflake, GoodData, NVIDIA GSF, Salesforce, and Polaris readers/writers, fixtures, snapshots, and usage documentation.
- Remove obsolete root `dialects`/`vendors` fields and first-model/drop-extra-model behavior. Preserve vendor-native collections such as dbt's `semantic_models`.
- Preserve the ontology mapping's embedded `semantic_model` structure and exclude standalone document metadata from that embedded object.
- Add Polaris `import --output-dir DIR` to emit one document per nonempty namespace with deterministic, collision-safe filenames and no overwrites. Single-file/stdout imports require one nonempty namespace.
- Trigger converter CI on shared schema/example changes and run shared SDK tests in Validation CI.

This is a breaking SDK and converter format change for the mutable `0.2.0.dev0` specification. Legacy documents must be migrated first; see the spec's migration guidance.

## Checklist

- [x] Converter logic, fixtures, examples, and documentation reflect #383.
- [x] Breaking behavior and bulk export migration described.
- [x] Focused regression coverage added or updated.
- [x] Ontology embedding and vendor-specific model collections preserved.
- [x] No runtime dependencies added.

## Validation

- After rebasing onto #407: 43 focused Microsoft tests and 5 focused SDK tests passed. Microsoft lint and whitespace checks passed.
- Compared the rebased trees: the Microsoft converter matches main, and the remaining SDK/converter changes are preserved.
- Earlier migration validation included 199 focused SDK/converter tests, the Microsoft CLI import/export round-trip, and schema validation of all 10 checked-in standalone converter fixtures. Another 104 focused tests passed for the root-validation fixes.
- Full regression suites and optional Microsoft TOM/live-engine checks were not run locally; full regression validation is left to CI.


jbonofre:
I would appreciate to review this PR as I'm doing a bunch of work on converters.

kayemkim (commented):
Ran this against the validation and all thirteen converter workflows locally on the rebased head (366572c on bc7b2c0). Everything I could run is green and matches the checks here, and the two Java suites are green in CI. A few things beyond the test suites:

The repository's own documents now validate. On main, 10 of the 11 Ossie documents under `examples/` and `converters/**` fail against the schema because the fixtures still carry the wrapper. On this branch all pass except the salesforce example, which fails only on its four pre-existing `[SQL]` lines (bracketed identifiers), the same as before #383.

The wrapper is rejected explicitly rather than mis-read. Feeding the pre-#383 TPC-DS example to the SDK gives `semantic_model: Extra inputs are not permitted` alongside the missing `name`/`datasets`, and the sigma, snowflake, honeydew and nvidia CLIs all exit 1 with a message naming the wrapper. The flat example converts on all four. The five places I listed on the 15th are all handled, and 366572c gives every converter the same `Root dialects and vendors are not supported` check, which also covers Copilot's six inline notes on the earlier head.

The `core-spec/**` and `examples/**` triggers on the converter workflows are welcome. That gap is what let #297 sit with a failing orionbelt suite for a week and let #383 land with the Microsoft suite broken until #407.

For whoever merges, I merged each open converter-area PR onto this branch. New conflicts this PR introduces, each one file: #404 (salesforce README), #378 (dbt `msi_to_ossie.py`), #333 (databricks README), #338 (`python/tests/test_models.py`). Merging cleanly but still reading the wrapper, so they will break at runtime once this lands: the new converters in #289, #364, #352, #360, #381, #320, the interop harness in #348, and the Java converter in #333. Their authors will need a heads-up since neither the schema nor the SDK gate catches a parser that looks for `semantic_model` itself.

I have not gone through the Polaris `--output-dir` Java changes line by line. It follows from one model per document, and its tests pass. LGTM for the migration.

flyrain:
Hi @khush-bhatia @jbonofre, could you take a look at this PR? Don't want the main branch in a inconsistent status for too long. Thanks!

khush-bhatia (approved):
Thanks, this looks good.

khush-bhatia:
Merging this since this is affecting all converters and CI is red until this merges. cc @jbonofre and @flyrain

PR #400, opened 2026-09-15
Rename ontology uri property to iri

jochenchrist:
Follow-up to #332.

Renames the optional global identifier on concepts and relationships from `uri` to `iri`, as agreed in the discussion on #332 (IRI per [RFC 3987](https://www.rfc-editor.org/info/rfc3987/)). The `prefixes` map now uses the JSON Schema `iri` format, and the descriptions say IRI instead of URI.

Also documents the new fields in `ontology.md`: `prefixes` in the top-level table, `iri` in the concept and relationship tables, and a new "Global identifiers" section that explains IRIs, QName expansion against `prefixes`, and that IRIs do not change how local names are used in expressions and mappings. The example in that section validates against the updated schema.


ontology/ontology.md
```diff
@@ -379,6 +382,60 @@ ontology:
...
+Concept and relationship names are local to the ontology that declares them. To relate a concept
+or relationship to a definition outside the ontology, for instance a class or property in an
+existing RDF or OWL vocabulary, it can carry an optional `iri` field that holds a globally unique
+identifier. An IRI (Internationalized Resource Identifier, [RFC 3987](https://www.rfc-editor.org/info/rfc3987/))
+is a generalization of a URI that permits characters beyond ASCII. Any URI is also a valid IRI.
+
+The `iri` field accepts two forms:
+
+- A full IRI, e.g. `http://xmlns.com/foaf/0.1/Agent`.
+- A QName of the form `prefix:local`, e.g. `foaf:Agent`, where `prefix` is declared in the
```
Ostrzyciel, line 396: Not directly related to this PR, but... Please note that deciding which form to parse here is not really possible to do for all possible cases. For example `urn:isbn:whatever` is a valid IRI. What if you have a prefix `urn:`?

In LinkML there is the exact same issue with the `uriorcurie` datatype. See: https://github.com/linkml/linkml/issues/3906

Should there be a follow-up to define a clear algorithm for how `iri` should be parsed and serialized?

khush-bhatia (approved):

ontology/ontology.md
scotthenninger: URNs are valid IRIs (from RDC 3987) so they should be allowed. @jochenchrist perhaps the md can reflect this:

The ‘irk’ field accepts two forms:

 - A full IRI, e.g. `http://xmlns.com/foaf/0.1/Agent`, `urn:foaf:Agent`
 - A QName …
 
I’m also curious about @Ostrzyciel’s statements that it’s not possible to implement all cases. RDF processors have been doing this for years. Do I misunderstand your statement?

kayemkim (commented):
One data point for the QName versus full IRI thread: nothing in the repo enforces either form today. `validation/validate.py` builds `Draft202012Validator(schema)` without a format checker, so `format: iri` (like `format: uri` before it) is descriptive only, and jsonschema ships no `iri` checker anyway unless `rfc3987` is installed. The same goes for the "a QName whose prefix is not declared in `prefixes` is invalid" sentence; the validator accepts any string. So whichever resolution rule comes out of that discussion, it would need a semantic check next to the reference and arity ones, not a schema keyword. Not a blocker for this rename. Checked on current main that no example, fixture or converter in the repo uses the `uri` field, so the rename is complete as is.

ontology/ontology.md
jochenchrist: Good point. I think @Ostrzyciel is pointing that how can we decide if a value is a QName (to lookup the prefix) vs. an URN, as both have the form `schema:values`. 

However, URNs always start with `urn:...`.

An idea might be to not allow a prefix `urn`.

I will update the PR accordingly to mention that URNs are alos allowed as IRI.

Ostrzyciel: > I’m also curious about @Ostrzyciel’s statements that it’s not possible to implement all cases. RDF processors have been doing this for years. Do I misunderstand your statement?

In Turtle, these two have different syntaxes. Full IRIs are: `<urn:fullIri>`. And CURIES/QNames are: `urn:qname`. They are interpreted differently.

> An idea might be to not allow a prefix urn.

URNs are just one possible case where this will break :) IRIs have *a ton* of forms. You can get a sample here: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Example_URIs

As for the solution to this conundrum: I would recommend a solution along the lines of this one proposed in LinkML: https://github.com/linkml/linkml/issues/3906#issuecomment-5387767756

jbonofre: So, what do we do about this ambiguity?

I'm fine to merge as is and follow up but it's something we should address.

Concretely, my take on that: if `prefixes` declares `urn: http://example.org/ns#` and a concept sets `iri: urn:isbn:0451450523` (a valid RFC 3987 IRI), does that get expanded as a QName into `http://example.org/ns#isbn:0451450523` or preserved as the literal URN?

As written, an implementer following "a QName ... where prefix is declared in the ontology-level prefixes map" would expend it (silently producing the wrong global identifier).

Where should we move here? @jochenchrist ?

ontology/ontology.md
```diff
@@ -379,6 +382,60 @@ ontology:
...
+    extends: [String]
+  - concept: Organization
+    type: EntityType
+    extends: [Agent]
+    iri: http://www.w3.org/ns/org#Organization
+```
+
+Here `Agent` and `Agent.has_homepage` are identified by QNames that resolve against the `foaf`
+prefix, while `Organization` is identified by a full IRI. Both forms denote the same kind of
+identifier; the QName is merely shorthand. A QName whose prefix is not declared in `prefixes`
```
jbonofre: Nit: this states a validity rule, but nothing enforce it: `ontology/ontology.json` schema types `iri` as a plain string with no pattern tying it to the declared `prefixes` key, and `validation/validate.py` builds its `Draft202012Validator` with no custom format/keyword to check this cross-field constraint. Right now `iri: bogus:Thing` with no `bogus` prefix declared passes validation despite being "invalid" per this spec.

It has been mentioned before by @kayemkim . I believe either the schema/validator should enforce this, or the doc should say explicitly that this is a semantic constraint left to tooling outside `validate.py`, so implementers don't assume schema validation covers it.

jbonofre (approved):

PR #406, opened 2026-09-15
fix(sigma): record dropped keys when a relationship's column arrays have unequal length

### LINKED ISSUE #405: Sigma converter silently drops key columns when a relationship's from_columns/to_columns lengths differ
## Summary

`OSIToSigmaConverter._build_relationship` pairs a relationship's `from_columns` and `to_columns` with a bare `zip()`. The OSI schema only requires each array to have at least one entry independently (`$defs/Relationship` in `core-spec/ossie-schema.json`); it does not require them to be the same length. So a compound-key relationship with unequal-length arrays is legal input, and `zip()` silently stops at the shorter array, dropping the extra key column(s) from the exported Sigma spec with no warning.

## Steps to reproduce

```python
from ossie import OssieDataset, OssieDocument, OssieRelationship, OssieSemanticModel
from ossie_sigma.ossie_to_sigma import OssieToSigmaConverter

document = OssieDocument(
    semantic_model=[
        OssieSemanticModel(
            name="m",
            datasets=[
                OssieDataset(name="orders", source="db.public.orders"),
                OssieDataset(name="regions", source="db.public.regions"),
            ],
            relationships=[
                OssieRelationship(
                    name="OrderRegion", **{"from": "orders"}, to="regions",
                    from_columns=["region_id", "sub_id"], to_columns=["region_id"],
                ),
            ],
        )
    ]
)
result = OssieToSigmaConverter().convert(document)
print(result.output["pages"][0]["elements"][0]["relationships"][0]["keys"])
print(result.issues)
```

`keys` comes out with a single `region_id` pair; `sub_id` is gone and `issues` is empty, so nothing about the loss is visible to the caller.

## Environment

`apache/ossie` `main` @ `fc6c9df` (2026-09-15), `converters/sigma`.

AmirF194:
## Summary

`_build_relationship` paired `from_columns`/`to_columns` with a bare `zip()`, which stops at the shorter array. The OSI schema only requires each array to be non-empty independently, so a compound-key relationship with unequal lengths is legal input, and the extra key column(s) were dropped from the exported Sigma spec with no record of it.

Thread `issues` into `_build_relationship` and record a `RELATIONSHIP_COLUMN_ARITY_MISMATCH` issue when the lengths differ, mirroring the `RELATIONSHIP_COLUMN_UNRESOLVED` pattern already used for the reverse (Sigma-to-Ossie) direction. The conversion still degrades to the shorter pairing, since Sigma's `keys` array has no way to represent an unequal-length join, but the loss is now visible in `ConverterResult.issues` instead of silent.

## Related Issues

Fixes #405

### Converters
- [x] Converter logic in `converters/sigma` is updated
- [x] New tests cover the arity-mismatch case

### Tests
- [x] `uv run pytest` on Python 3.11, 3.12, 3.14 - 92 passed on each
- [x] Ran the new test on plain `main`: it errors there (the issue type doesn't exist yet)

### Compliance
- [x] No new source files, no new dependencies


converters/sigma/src/ossie_sigma/ossie_to_sigma.py
```diff
@@ -365,6 +366,18 @@ def _build_relationship(
         if raw_keys is not None:
             result["keys"] = raw_keys
         else:
+            if len(rel.from_columns) != len(rel.to_columns):
+                # zip() below stops at the shorter array; record what it drops.
+                issues.append(
+                    ConverterIssue(
+                        ConverterIssueType.RELATIONSHIP_COLUMN_ARITY_MISMATCH,
+                        rel.name,
```
jbonofre: This file already scopes relationship identity by `(dataset_name, rel.name)` two lines above (`_stable_id("relationship", dataset_name, rel.name)`) specifically because two unrelated relationships can share a name across different table pairs (see `test_relationship_ids_are_scoped_by_owning_dataset`). Using bare `rel.name` as `element_name` for the new issue breaks that convention: if two such relationships both hit an arity mismatch, their `ConverterIssue` entries become indistinguishable.

I suggest scoping it `f"{dataset_name}.{rel.name}"`.

jbonofre (changes requested):

AmirF194:
Good catch, fixed: the arity-mismatch issue now uses f"{dataset_name}.{rel.name}" for element_name, matching the relationship id's own scoping. Added test_relationship_arity_mismatch_element_names_are_scoped_by_owning_dataset (two same-named relationships on different table pairs, both arity-mismatched); against the prior code it fails with 1 == 2, both issues collapsing onto "Parent". Full 93-test suite passes in the repo's own verify command. Pushed 633ca26.

khush-bhatia:
@AmirF194 Please see this https://github.com/apache/ossie/pull/375 which was merged recently.
I believe the  converter code should be changed to fail this case instead.

converters/sigma/src/ossie_sigma/ossie_to_sigma.py
```diff
@@ -365,6 +366,18 @@ def _build_relationship(
         if raw_keys is not None:
             result["keys"] = raw_keys
         else:
+            if len(rel.from_columns) != len(rel.to_columns):
```
jbonofre, line 369: Now `core-spec/spec.md` (Relationship) states: "Both arrays must have the same number of columns". It's not something we should warn and proceed, it's a "real" issue.

The PR here handles the identical condition by recording a `ConverterIssue` and then converting via `zip()`, which silently truncates to the shorter array. That's inconsistent with the spec statement, and it means a document that skips `validate.py` gets corrupted output from the converter instead of being rejected.

We already have what we need: `converter_issues.py` defines `ConverterError`, raised when the input cannot converted at all, as opposed to partial, lossy conversion that a `ConverterIssue` an describe. This file already uses it for structurally invalid input (empty `semantic_model`).

An arity mismatch is a spec-invalid input, not a legal-but-lossy translation gap like `EXPRESSION_NOT_TRANSLATABLE`. 

I suggest raising `ConverterError` here instead of appending a `ConverterIssue` and continuing.

jbonofre (changes requested):

converters/sigma/src/ossie_sigma/ossie_to_sigma.py
jbonofre: @khush-bhatia FYI, see my request change here (about `ConverterError`).

AmirF194:
I'd keep the converter's own handling rather than switch to a hard fail: #375's arity check lives in validation/validate.py, a separate script the docs recommend running before conversion but nothing wires into ossie-sigma's CLI or model_validate() call. A document can reach OssieToSigmaConverter without ever going through that script, so this path is still reachable with real input, not just a defensive-programming guard against something the schema now forbids.

PR #407, opened 2026-09-16
Fix Microsoft converter for flat documents

eisber:
## Summary

- update `converters/microsoft` to read and emit the flat one-model-per-document Ossie shape introduced by #383
- reject legacy `semantic_model` wrappers and removed root metadata instead of ambiguously converting them
- preserve Power BI/TOM round-trip behavior for stash data, AI context, relationships, measures, and the installed CLI
- expand Microsoft CI path coverage to all core-spec and example changes

## Reproduction

On current `main` before this change:

- `test_all_tpcds_example_metrics_translate_to_dax` failed because the converter rejected the flat TPC-DS example as missing `semantic_model`
- `test_output_validates_against_core_spec_schema` failed because converter output still used the removed wrapper shape
- result: **2 failed, 381 passed, 1 skipped**

## Validation

- Microsoft suite with coverage: **399 passed, 0 skipped, 96.00% coverage**
- TOM unit and integration tests: **16 passed**
- wheel-installed `ossie-microsoft` import/export round trip: passed
- generated YAML validation against `core-spec/ossie-schema.json`: passed
- `ruff check converters/microsoft`: passed
- `git diff --check`: passed

`ruff format --check` reports 16 pre-existing unformatted files, so this PR does not introduce a repository-wide formatting rewrite.

## Overlap

The Microsoft-scoped implementation matches the corresponding changes in #396. This PR is intentionally self-contained and limited to restoring the Microsoft converter CI path.


converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
```diff
@@ -179,7 +179,7 @@ def build_ossie_document(bim_file):
         excluded_measures,
     )
 
-    return {"version": OSSIE_VERSION, "semantic_model": [semantic_model]}
+    return {"version": OSSIE_VERSION, **semantic_model}
```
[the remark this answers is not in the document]
eisber: Fixed in dbe344e. `build_ossie_document` now raises a clear `ValueError` after reporting exclusions when no table can become an Ossie dataset, so the CLI exits non-zero without creating schema-invalid YAML. Added regression coverage for malformed tables, all-excluded calculation/calculated tables, and the CLI no-output path. Full Microsoft suite: 400 passed with 96.01% coverage; TOM, Ruff, diff checks, and normal CLI schema validation pass.

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
```diff
@@ -163,16 +163,16 @@ def convert_ossie_to_semantic_model(
...
-            "and skipping the rest",
+    if "semantic_model" in document:
+        raise ValueError(
+            "Legacy 'semantic_model' wrappers are not supported; "
+            "place model properties at the document root"
         )
-    semantic_model = models[0]
+    if "dialects" in document or "vendors" in document:
+        raise ValueError("Root dialects and vendors are not supported by the Ossie spec")
+    if not document.get("name") or "datasets" not in document:
```
jbonofre: This only checks that `datasets` is present, not that it's a list of dicts. An unwrapped single dataset dict (a very plausible mistake given this PR removes a list wrapper) passes silently and produces a 0-table model with no warning, `_convert_datasets` iterates the dict's keys (strings), each failing `isinstance(dataset, dict)`. The old `isinstance(models[0], dict)` guard caught this, nothing replaces it. 

Also worth tightening while here: `datasets=None/{}/[]/"str"` all pass and silently yield an empty model, which violates core spec `minItems: 1` and is inconsistent with the new hard-fail added on the export side in this PR. 

I suggest validating `isinstance(document.get("datasets"), list) and len(...) > 0` and that each element is a dict.

converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
```diff
@@ -178,8 +178,12 @@ def build_ossie_document(bim_file):
         excluded_relationships,
         excluded_measures,
     )
+    if not tables:
```
jbonofre: This `raise ValueError` runs after `_stash_model` has already called `warn()` for every excluded table. Under the document strict-mode usage (`warnings.simplefilter("error")`, per the README), that means a `UserWarning` is raised here instead of the intended `ValueError`, callers catching `ValueError` per the documented contract get an unhandled exception of the wrong type.

The warning text is also misleading in this path, since it claims the table is "preserved in custom_extensions" when the whole result is about to be discarded.

I suggest to move the `if not tables:` check to right after `tables = [...]`, before any conversion/warning logic.

converters/microsoft/README.md
```diff
@@ -24,6 +24,13 @@ semantic model. Power BI output is available as either a TMSL `model.bim` mappin
 TMDL document. The conversion is offline and requires no Power BI or Fabric
 connection.
 
+Each Ossie JSON/YAML document contains one model, with `name`, `datasets`,
+`relationships`, and `metrics` at the root alongside `version`. Legacy
+`semantic_model` wrappers must be unwrapped before conversion; split documents
+containing multiple models into separate files. Because the Ossie schema requires at
+least one dataset, importing a Power BI model fails explicitly when every table is
```
jbonofre, line 31: This claim ("fails explicitly when every table is malformed or excluded") contradicts the strict-mode guidance documented later in this README: following that guidance and converting a model whose only table is a calculation group actually raises `UserWarning`, not `ValueError`.

Once that's fixed this should hold true. Until then it's inaccurate.

converters/microsoft/tests/conftest.py
```diff
@@ -46,4 +46,4 @@ def osi(bim):
 
 @pytest.fixture(scope="module")
 def model(osi):
-    return osi["semantic_model"][0]
+    return osi
```
jbonofre: The shared `model` fixture now returns the full flattened document, including a real top-level `"version"` key. Any test using `{"version": OSSIE_VERSION, **model}` (or a foreign version literal) to override the version will have that override silently discarded in favor of the fixture own `"version"`, since dict unpacking order means the fixture key wins.

Currently masked because both happen to equals `OSSIE_VERSION`, but it's a footgun for the next person who reuses this fixture to test a version mismatch.

We should consider having the fixture return the unwrapped model (pre-flattening) as before, with a separate fixture/helper for the full document.

jbonofre (changes requested):

converters/microsoft/src/ossie_microsoft/ossie_to_semantic_model.py
eisber: Fixed in 3fff22d. Flat input is now validated before conversion as a non-empty list of named dataset objects. `None`, mappings, empty lists, strings, mixed lists, and unnamed objects all raise `ValueError` instead of silently yielding an empty TMSL model. The now-unreachable no-table metric fallback was removed.

converters/microsoft/src/ossie_microsoft/semantic_model_to_ossie.py
eisber: Fixed in 3fff22d. The no-exportable-table check now runs immediately after table selection, before stash construction or exclusion warnings. Tests run both calculation-group-only and calculated-table-only inputs under `warnings.simplefilter("error")` and confirm the documented `ValueError` is raised.

converters/microsoft/README.md
eisber: Fixed in 3fff22d. All-excluded models now fail before any exclusion warning is emitted, including under the README strict-mode pattern, so the documented explicit-failure claim is accurate.

converters/microsoft/tests/conftest.py
eisber: Fixed in 3fff22d. The shared `osi` fixture remains the complete flat document, while `model` now returns a version-free model mapping. Existing `{"version": ..., **model}` callers can no longer have their requested version silently overwritten.

jbonofre:
@eisber thank you so much! I'm doing another pass.

jbonofre (approved):

PR #410, opened 2026-09-17
fix(orionbelt): carry field and metric datatype in both directions

### LINKED ISSUE #409: OrionBelt converter drops a field's logical datatype in both directions
## Summary

Neither conversion direction carries a field's logical type through the
canonical spec `datatype` field:

- **Ossie->OBML** ignores the declared `datatype` and derives OBML
  `abstractType` from a column-name heuristic, so a `String` field can be
  emitted as numeric and vice-versa (**wrong** type).
- **OBML->Ossie** never writes the spec `datatype` field at all - the type is
  stashed only inside an OrionBelt `custom_extensions` blob (**missing** type).

A round-trip appears lossless only because the converter is self-consistent
with its own non-spec key; any other spec-compliant tool sees wrong or absent
types.

## Root cause

**Ossie->OBML** (ossie_to_obml.py):

```python
ossie_type = field.get("data_type", "")               # (1) wrong key
if ossie_type and ossie_type in OSSIE_TO_OBML_TYPE:    # (2) wrong-cased map
    abstract_type = OSSIE_TO_OBML_TYPE[ossie_type]
else:
    abstract_type = self._infer_obml_type(field)
```

1. **Wrong key.** The spec field is `datatype`, no underscore, so the branch is dead
2. **Wrong-cased map.** The spec enum is capitalized, but
   `OSSIE_TO_OBML_TYPE` is keyed lowercase with different terms.
   Even after fixing the key, case-normalizing is not enough:
   `DateTime`/`DateTimeTz`/`Decimal`/`Float` have no key - so case-normalizing is not enough

**OBML->Ossie: canonical field never written** (`obml_to_ossie.py`).
The type goes only into a vendor extension, under the non-spec key
`data_type` with a lowercase value; `field["datatype"]` is not set.

## Illustrative Repro

```python
# Ossie -> OBML: declared datatype ignored
for f, want in [({"name": "total_amount", "datatype": "String"},  "string"),
                ({"name": "customer_id",  "datatype": "Integer"}, "int")]:
    _, col = OssietoOBML(ossie={})._convert_field(dict(f))
    got = col["abstractType"]
    print(f"fwd {f['datatype']:8} -> {got:8} (want {want}) [{'OK' if got==want else 'BUG'}]")

# OBML -> Ossie: canonical datatype never emitted
field = OBMLtoOssie(obml={})._convert_column("customer_id", {"code": "customer_id",
                    "abstractType": "int"}, "Orders", {})
print("rev canonical 'datatype' present?", "datatype" in field, "[BUG]" if "datatype" not in field else "")
```

## Why this survived

The existing fixtures (test_ossie_converter_vendors, test_ossie_metric_no_silent_loss)
use the same wrong key `data_type`, so the suite stays green with the bug present.

Drop-in test (currently fails). Suggested home:
`tests/test_ossie_metric_no_silent_loss.py`, which already guards this "no
silent loss on field conversion" path.

```python
def test_explicit_datatype_roundtrips():
    _, col = OssietoOBML(ossie={})._convert_field({"name": "total_amount", "datatype": "String"})
    assert col["abstractType"] == "string"          # fwd: currently 'float'
    field = OBMLtoOssie(obml={})._convert_column("total_amount",
                {"code": "total_amount", "abstractType": "string"}, "Orders", {})
    assert field.get("datatype") == "String"        # rev: currently missing
```

## Suggested fix

- **Ossie->OBML:** read `field.get("datatype")` (fix the comment); rebuild
  `OSSIE_TO_OBML_TYPE` on the spec enum
- **OBML->Ossie:** set the canonical `field["datatype"]` (capitalized enum value)
  in addition to any extension bookkeeping.

ralfbecher:
## Summary

The OrionBelt converter lost a field's logical type in both directions (#409): import read the non-spec `data_type` key against a lowercase map, so the spec `datatype` was ignored in favour of the name heuristic, and export never wrote `datatype` at all. This ports the fix that already shipped in the downstream `osi-orionbelt` package (ralforion/orionbelt-semantic-layer#246).

**Ossie -> OBML**
- Field `datatype` maps to OBML `abstractType`. Precedence: `datatype` > legacy `data_type` > name heuristic.
- `Decimal` narrows to `float` (OBML models exact decimals at the physical layer, not in `abstractType`); `Opaque` falls back to the heuristic.
- The `obml_abstract_type` stash written on export is now restored on import, so OBML -> Ossie -> OBML stays exact through the narrowing (`json`, `time_tz`).
- Metric `datatype` maps to the exact measure/metric `dataType` (`Decimal` -> `decimal(18, 2)`, `Integer` -> `integer`, `Float` -> `double`), unless one was already restored from the extension.

**OBML -> Ossie**
- Fields always emit `datatype` from `abstractType` (`json` -> `Opaque`, `timestamp_tz` -> `DateTimeTz`), in addition to the existing extension bookkeeping.
- Measures/metrics emit `datatype` only when an explicit `dataType` is declared, so plain measures stay unchanged and round trips remain idempotent.

| Ossie `datatype` | OBML field `abstractType` | OBML measure/metric `dataType` |
|---|---|---|
| `String` | `string` | `string` |
| `Integer` | `int` | `integer` |
| `Float` | `float` | `double` |
| `Decimal` | `float` | `decimal(18, 2)` |
| `Boolean` | `boolean` | `boolean` |
| `Date` / `Time` | `date` / `time` | `date` / `time` |
| `DateTime` | `timestamp` | `timestamp` |
| `DateTimeTz` | `timestamp_tz` | `timestamp` |
| `Opaque` | name heuristic | not set |

## Related Issues

Closes #409

## Notes for reviewers

- **Pre-existing failures on `main`:** 6 schema-validation tests in `converters/orionbelt` already fail on `main` since the flat document change (#383), e.g. `test_tpcds_fixture_passes_v02_validation`. They fail identically with and without this PR; the converter CI has not run since that change. #396 migrates the converter to the flat shape and should fix them.
- **Overlap with #396:** it touches the same files. The diff here is small and additive, so I'm happy to rebase onto #396 if that lands first.
- The existing fixtures still use `data_type`; they keep working via the legacy fallback and were left unchanged to keep this PR focused.

### Converters

### Documentation
- [x] Mapping analysis (`converters/orionbelt/ossie_obml_mapping_analysis.md`) updated

### Tests
- [ ] All existing tests pass (`pytest` / CI green): 158 pass; the 6 failures above are pre-existing on `main`
- [x] New functionality is covered by tests: `test_ossie_converter_datatype.py` (10 tests) plus the drop-in test from #409 in `test_ossie_metric_no_silent_loss.py`

### Compliance


converters/orionbelt/src/ossie_orionbelt/_common.py
```diff
@@ -91,3 +91,90 @@
...
+
+def obml_datatype_to_ossie(data_type: str | None) -> str | None:
+    """Map an explicit OBML measure/metric ``dataType`` to an Ossie ``DataType``.
+
+    Returns ``None`` when there is no mapping, so the caller emits nothing rather
+    than an unknown type. ``decimal(p, s)`` maps to ``Decimal``.
+    """
+    if not data_type:
+        return None
+    normalized = data_type.strip().lower()
```
jbonofre, line 180: `data_type.strip()` assumes `data_type` is a string, but nothing guarantees that if the source OBML document is malformed or hand-authored (the type hint is `str | None`, not enforced). A non-string `dataType` (e.g. `123`) crashes with `AttributeError: 'int' object has no attribute 'strip'` and aborts the whole conversion.

I think it's worth guarding with `isinstance(data_type, str)` up front (returning `None` otherwise) rather than relying on upstream schema validation that may not have run?

converters/orionbelt/src/ossie_orionbelt/_common.py
```diff
@@ -91,3 +91,90 @@
...
+    "timestamp": "DateTime",
+    "timestamp_tz": "DateTimeTz",
+    "boolean": "Boolean",
+}
+
+# Metric/measure `datatype`. Unlike fields, OBML measures/metrics carry an exact
+# `dataType` (physical vocabulary: `integer`/`double`/`decimal(p, s)`/...), which
+# is where `Decimal` genuinely belongs. So Ossie metric `datatype` maps to that
+# field, not the coarse `abstractType`.
+OBML_DECIMAL_DEFAULT = "decimal(18, 2)"  # mirrors OrionBelt's built-in default
```
jbonofre, line 138: `OBML_DECIMAL_DEFAULT = "decimal(18, 2)"` is applied unconditionally whenever an Ossie metric has `datatype: "Decimal"`. But OBML models can override the default numeric type via `settings.defaultNumericDataType`, this hardcodes past that override, so a model configured for e.g. `decimal(20, 6)` gets metrics silently emitted as `decimal(18, 2)` instead.

Should this read `obml_settings.defaultNumericDataType` (when present) before falling back to the `"decimal(18, 2)"` constant?

converters/orionbelt/src/ossie_orionbelt/ossie_to_obml.py
```diff
@@ -829,6 +843,15 @@ def _convert_metrics(self, ossie_metrics: list, ds_map: dict) -> tuple[dict, dic
             target = metrics.get(m["name"]) or measures.get(m["name"])
             if target is not None:
                 self._carry_foreign_extensions(m.get("custom_extensions"), target)
+                # Ossie metric `datatype` -> OBML exact `dataType` (its natural
+                # home; `Decimal` -> decimal(p, s)). Don't override a dataType
+                # already restored from an OBML-origin extension, and skip
+                # Opaque/unknown (absent from the map).
+                ossie_dt = m.get("datatype")
+                if ossie_dt and not target.get("dataType"):
```
jbonofre: This guard skips setting `dataType` whenever `target` already has one from `_apply_obml_measure_extras` (the `obml_data_type` custom_extensions stash). But that stash can be stale, a user edits a metric's `datatype` in Ossie after a prior OBML round-trip, and the fresh `datatype` is silently dropped in favor of the old cached value, with no warning.

Should an explicit `datatype` on the Ossie metric take precedence over the stashed extension, since it reflects the user's latest edit?

jbonofre (changes requested):

converters/orionbelt/src/ossie_orionbelt/_common.py
ralfbecher: Good catch, fixed in 7375c7e. `obml_datatype_to_ossie` now returns `None` for anything that is not a non-empty string, so a hand-authored `dataType: 123` is treated like an unknown type and the conversion carries on. I applied the same guard on the import side: a non-string field `datatype` or legacy `data_type` counts as absent, and a non-string metric `datatype` maps to nothing. Before, a list there raised `TypeError` on the dict lookup. Covered by `TestMalformedDatatype`.

converters/orionbelt/src/ossie_orionbelt/_common.py
ralfbecher: Agreed, fixed in 7375c7e. An Ossie `Decimal` metric now takes `settings.defaultNumericDataType` when the model has one, and falls back to `decimal(18, 2)` otherwise. The settings only exist on an OBML-origin model, in the stashed `obml_settings`, and the importer restored them after converting metrics. So they are now read before the metrics, and the restore itself stays where it was. OrionBelt rejects a `defaultNumericDataType` that is not a `decimal(p, s)`, so any other value falls back to the built-in default here too. Covered by `TestDecimalDefaultFromSettings`.

converters/orionbelt/src/ossie_orionbelt/ossie_to_obml.py
ralfbecher: Yes, the edit should win, fixed in 7375c7e for metrics and for fields, which had the same problem with `obml_abstract_type`. I did not make `datatype` win unconditionally, though. On a plain round trip the stash is more exact than the map: `decimal(20, 6)` and `bigint` would come back as `decimal(18, 2)` and `integer`. So the stash is kept while it agrees with `datatype`, meaning it maps back to the same Ossie type. When it maps to a different type, the `datatype` was edited after the export and replaces it. Covered by `TestEditedDatatypeBeatsStaleStash`, including a check that an agreeing stash stays exact.

PR #412, opened 2026-09-17
Feat(ontology): add Ossie -> RelationalAI converter with formula parsing and reasoner

vmihalovski:
### Summary

Adds an Ossie -> RelationalAI converter: an Ossie ontology converts into an executable PyRel model.

### What's added

**`expr/` — the formula layer (~2,550 lines).** A `ply` lexer and parser for Ossie's formula grammar, an AST, a validator, and visitors that collect handles, resolve unbound references, and emit PyRel. Parsing is on by default in `OssieParser`; the raw string factories remain as an opt-out for sources whose formula vocabulary this grammar doesn't model.

**`converter/ossie_to_relationalai/` (~655 lines).** Converts an `OssieOntology` into a PyRel model — concepts, relationships, entity and link mappings, derived rules. Table resolution is pluggable through a `table_provider`, so a caller can read from somewhere other than a warehouse without the converter knowing such alternatives exist.

**`vendor/relationalai/` (~2,280 lines).** `OntologyModel`, the binding types the converter emits against, and role helpers.

**`reasoner.py` (159 lines).** Answers which relationships identify a concept (`ref_scheme`), whether a relationship constructs one (`is_identifier_relationship`), and how concepts sit in the type hierarchy (`in_subtype_closure`). Only identifiers are indexed; everything else is read off the model on demand, so the answers don't depend on the order concepts were registered.

### What's changed

- `external/` renamed to `vendor/`, so vendor-specific code sits under one name.
- `spec_to_ossie`: a bare `expression:` mapping on an entity now takes its type from the entity's reference scheme rather than from the entity itself — a column of store numbers is `StoreNr`, not `Store`. This also removes a spurious conflict when one column was keyed through both mapping styles.
- `ossie_to_relationalai`: a bare-expression link node with children keys its child binding by the identifier value rather than the wrapped entity.
- `model.py`: `Concept.is_primitive` walks `extends` iteratively, so a malformed cyclic hierarchy returns an answer instead of `RecursionError`.
- Public API gains `OntologyReasoner`, `FormulaParserFactory`, `MappingFormulaParserFactory`.

### Packaging

`relationalai` is an opt-in extra (`pip install "apache-ossie-ontology[relationalai]"`). Parsing Ossie, converting Palantir and reading or writing the spec all work without it, and nothing reachable from `ossie_ontology/__init__.py` imports it. `tests/test_optional_extra.py` asserts that boundary in a subprocess with the SDK masked.

### Tests

199 tests, organised to mirror `src/`:

| package | tests | covers |
|---|---|---|
| `palantir/` | 54 | reading an export; what the converter admits |
| `spec/` | 37 | parsing, and the corpus pinned by snapshot |
| `common/` | 33 | graph sort, text helpers |
| `model/` | 26 | concepts, roles, verbalizations |
| `expr/` | 23 | formula grammar and validation |
| `relationalai/` | 12 | conversion, PyRel snapshots, executability |
| top level | 14 | suite and packaging guarantees |

**The suite runs offline and enforces it.** An autouse fixture refuses every socket connection, so a test that reaches for a warehouse fails immediately rather than succeeding on whichever machine has credentials. `tests/config/raiconfig.yaml` pins `relationalai` to an offline profile — required, since it will not construct a Model without a config and otherwise searches `~/.rai`, `~/.snowflake` and `~/.dbt`. `tests/test_offline.py` asserts the pin, that the config carries no reachable account, and that the guard itself works.

PyRel snapshots are `.py` and are executed by `test_generated_pyrel_is_executable`, so what's asserted is the code a user would otherwise have hand-written. pytest doesn't collect them (it only matches `test_*.py`) and `[tool.pyright]` excludes `tests/snapshots`.


converters/ontology/src/ossie_ontology/expr/formula/lexer.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def t_ID(self, t):
+        r'[a-zA-Z_0-9]+'
+        t.type = self.reserved.get(t.value, 'ID')
+        return t
+
+    # every string that doesn't match one of the previous tokens is considered an error
+    def t_error(self, t):
+        r'.'
+        return SyntaxError(f'Illegal character {t.value!r} at position {t.lexpos}')
```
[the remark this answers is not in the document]
vmihalovski: fixed

converters/ontology/src/ossie_ontology/expr/model.py
```diff
@@ -0,0 +1,77 @@
...
+@acceptor
+class BinOp(str, Node, Enum):
+    # comparison operators
+    EQ = "=="
+    NE = "!="
+    LT = "<"
+    LE = "<="
+    GT = ">"
+    GE = ">="
+    comparison_ops = {EQ, NE, LE, LE, GT, GE}
```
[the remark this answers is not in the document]
vmihalovski: fixed

converters/ontology/tests/palantir/converter/test_object_type_statuses.py
```diff
@@ -0,0 +1,78 @@
...
+
+from ossie_ontology.converter.palantir_to_ossie.converter import PalantirToOssieConverter
+from ossie_ontology.vendor.palantir.model import Status
+
+from tests.palantir.converter.builders import dataset, object_type, prop
+
+from tests.palantir.converter.helpers import _convert, _concept_names
+
+
+def test_default_policy_admits_active_endorsed_andintermediary(tmp_path: Path):
```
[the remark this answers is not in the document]
vmihalovski: fixed

converters/ontology/tests/palantir/converter/test_relation_statuses.py
```diff
@@ -0,0 +1,145 @@
...
+    prop,
+)
+
+from tests.palantir.converter.helpers import (
+    _convert,
+    _concept_names,
+    _relationship_names,
+    _mapped_relationship_names,
+    _linked_export,
+)
```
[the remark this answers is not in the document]
vmihalovski: fixed

vmihalovski:
> ### 🔵 Needs a closer look
> There are confirmed runtime-correctness issues in newly added code (e.g., `role_part_of` not handling `Reading` roots and an unsafe `Expression.__hash__`) that should be fixed before approval.
> 
> Review details
> ### Suppressed comments (2)
> **Previously missed (2)** — in code that hasn't changed since the last review.
> 
> **converters/ontology/src/ossie_ontology/expr/model.py:70**
> 
> * `Expression.__hash__` calls each argument’s `__hash__()` directly. For unhashable args this can raise a confusing `'NoneType' object is not callable` (because `__hash__` is `None`), and it bypasses Python’s normal `hash()` behavior. Hashing the args directly is both simpler and more correct.
>   **converters/ontology/tests/conftest.py:166**
> * The `spec_model` fixture docstring says formulas “stay as raw text” and that parsing would pull in the optional `relationalai` extra, but the fixture calls `OssieParser()` with default factories (which now parse/validate formulas) and formula parsing does not depend on the `relationalai` extra. This mismatch is likely to confuse future maintainers.
> 
> * **Files reviewed:** 87/105 changed files
> * **Comments generated:** 0 new
> * **Review effort level:** Lite

Fixed

converters/ontology/src/ossie_ontology/expr/formula/lexer.py
```diff
@@ -0,0 +1,126 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import ply.lex as lex
+
+class FormulaLexer:
```
Ostrzyciel, line 20: You state that this is a "lexer and parser for Ossie's formula grammar" – but where is the formula grammar defined? I can't see it here: https://github.com/apache/ossie/blob/main/ontology/ontology.md

I'd be happy to implement it as well, but I can't implement something that does not exist. What is the plan to make other converters compatible with this?

converters/ontology/pyproject.toml
```diff
@@ -40,9 +40,20 @@ keywords = [
...
 ]
 
+# Extras are opt-in: `pip install "apache-ossie-ontology[relationalai]"`.
+# `relationalai` is the vendor SDK the Ossie -> RelationalAI converter emits
+# against. Parsing Ossie, converting Palantir, and reading/writing the spec all
+# work without it, and nothing reachable from `ossie_ontology/__init__.py`
+# imports it — so it stays out of the base install.
+[project.optional-dependencies]
+relationalai = [
+    "relationalai==1.27.1",
```
Ostrzyciel, line 58: This package is non-free. This may require PMC approval (please see the PR checklist that you get when you open a new PR). 

In my opinion, at the very least it should be clearly stated in the README that this is a fully proprietary package and what is its license (the PyPI listing does not say that). That matters because the license can be arbitrarily restrictive, so people installing this may unknowingly get into trouble.

Disclaimer: I'm not a lawyer, nor a PMC member.

converters/ontology/src/ossie_ontology/expr/formula/lexer.py
vmihalovski: @Ostrzyciel, the Ontology working group was reviewing the formula language for a few months, and you can find examples of it in the [flights.yaml](https://github.com/apache/ossie/blob/main/examples/flights.yaml#L296) as well as in the [ontology.md](https://github.com/apache/ossie/blob/main/ontology/ontology.md) the grammar that defines this language can be found [here](https://docs.google.com/document/d/10fDtwWPmNREaUVmtQ-7aI3Gr2FwPnESjofkjISNBSuk/edit?tab=t.0).

You may not have noticed this because we were discussing it in the Ontology working group, and I see that you only recently joined this group.

The semantics of relational formulas are context-specific. A discussion of this is in sections 4 and 5 of the draft [Ossie Ontology Semantic document](https://docs.google.com/document/d/1Vx1qtw-PK6LPLBOOI905eb7sPJz-87-3BW2Hagt0NO4/edit?tab=t.0#heading=h.xk5hzctpziqs).

converters/ontology/pyproject.toml
vmihalovski: @Ostrzyciel, this is a good point. I will update the README file and add links and additional information about RelationalAI.

converters/ontology/src/ossie_ontology/expr/formula/lexer.py
Ostrzyciel: OK, thank you for the links. But, as these documents are not in the repository, I assume these are still drafts that were not yet accepted into the spec. Shall we perhaps first merge them into the repository? Then it would be clear which spec version is the parser targeting, and what is correct and what is not. This is not really possible to do with an unversioned Google Doc. The doc could be changed tomorrow and we wouldn't know.

PR #413, opened 2026-09-17
docs(core-spec): register Power BI vendor name

eisber:
## Summary

- register the existing `POWER_BI` token in the canonical Vendor Names table
- identify it as Microsoft Power BI-specific metadata

The schema, YAML spec, and Microsoft converter already use `POWER_BI`; this change brings the human-readable registry into sync.

## Validation

- `git diff --check`


jbonofre (approved):

eisber:
@jbonofre  can we merge?

jbonofre:
@eisber done! Thanks!

PR #417, opened 2026-09-18
Fix Sigma and NVIDIA GSF converters for flat semantic model documents

jbonofre:
## Summary

`main` currently has two red CI checks: **Converters Sigma CI** and **Converters NVIDIA GSF CI**. Both broke because #383 moved the Ossie document schema to a single semantic model defined directly at the document root (dropping the `semantic_model:` list wrapper), and #407 followed up to fix the Microsoft converter for that shape — but the Sigma and NVIDIA GSF converters were never updated. Their CI simply hadn't re-run since #383 landed until an unrelated change (#416) touched their paths and re-triggered it, surfacing the break.

- **NVIDIA GSF**: emitted/expected the old `semantic_model: [...]` wrapper; the official validator now rejects that shape (`'name' is a required property`, `'semantic_model' was unexpected`).
- **Sigma**: `OssieDocument.model_validate(...)` failed loading the (already-flattened) `examples/tpcds_semantic_model.yaml` fixture with `semantic_model: Field required`.

**Update:** #396 merged into `main` while this PR was open and independently migrated the whole codebase (all converters + the Python SDK) to the flat document shape, covering the same Sigma/NVIDIA GSF fix this PR set out to make. After rebasing onto `main`, this PR's diff is now just a handful of small cleanups (docstring wording, a leftover `OssieSemanticModel` → `OssieDocument` type alias, formatting) left over once the overlapping changes were absorbed by #396.

## What changed

- **`python/src/ossie/models.py`**: `OssieDocument` now extends `OssieSemanticModel` directly (`version` + the model's fields at the document root) instead of wrapping a `semantic_model: list[OssieSemanticModel]`, matching the current `core-spec/ossie-schema.json`. It also drops the removed root-level `dialects`/`vendors` fields.
- **Sigma & NVIDIA GSF converters**: updated to read/write the flat document shape; removed the now-impossible "multiple models per document" handling (`EXTRA_MODEL_DROPPED`, empty-model errors) since a document is always exactly one model now.
- **dbt & wisdom converters**: also updated. Both share the `ossie` Python package's `OssieDocument`/`OssieSemanticModel` models, so changing `OssieDocument`'s shape would have broken them the moment their (currently green, but stale) CI next ran. Fixed in the same PR to keep the shared package's consumers consistent, including regenerating their syrupy snapshots.
- Removed the `dialects=[...]` argument dbt passed into `OssieDocument` (that field no longer exists at the document root) and the test assertions that exercised it.

(Most of the above landed via #396 by the time this PR was rebased; see the Update note above.)

All four converters plus the shared `python/` package pass their full test suites locally with the exact commands CI uses (`uv sync && uv run pytest`): Sigma 91/91, NVIDIA GSF 83/83, dbt 106/106, wisdom 27/27, python 25/25.

## Follow-up

~~While fixing this I found the same latent break in several other converters that haven't had CI re-triggered since #383: `databricks`, `orionbelt`, `ontology` (+ the root `examples/flights.yaml`), `gooddata`, `omni`, `honeydew`, and the Java-based `salesforce`/`polaris`. None of those are touched by this PR — filing a separate issue to track them.~~

**Resolved by #396**, which migrated every one of those converters to the flat document shape in the same merge. Their CI is green on `main`, so no follow-up issue is needed.

## Test plan

- [x] `cd python && uv run --with pytest --with PyYAML pytest tests/test_models.py` — 25 passed
- [x] `cd converters/sigma && uv run pytest` — 91 passed
- [x] `cd converters/nvidia && uv run pytest` — 83 passed
- [x] `cd converters/dbt && uv run pytest` — 106 passed
- [x] `cd converters/wisdom && uv run pytest` — 27 passed


mattsenicksigma (approved):
Sigma converter changes look good to me!

Issue #418, opened 2026-09-18
Update remaining converters for the flat (one-model-per-document) Ossie schema


jbonofre:
## Background

#383 moved the Ossie document schema to a single semantic model defined directly at the document root, dropping the `semantic_model: [...]` list wrapper. Since then, each converter has been fixed one at a time, only after its own CI happened to re-trigger and turn red:

- #407 fixed the Microsoft converter.
- #417 fixes the Sigma and NVIDIA GSF converters, plus the dbt and wisdom converters (which share the `python/` package's `OssieDocument` model and would have broken the moment it was made to match the schema).

While working on #417, I found the same latent break in the remaining converters — none of them are currently red in CI, because their path-scoped workflows simply haven't re-run since #383 landed, but their fixtures/source still reference the old wrapper shape and will fail the moment something re-triggers their CI (the same way #416's license-header commit accidentally did for Sigma/NVIDIA GSF).

## Affected

Confirmed via `grep -rn "semantic_model:"` over fixtures/source, and via `gh run list` showing no CI run for any of these since #383 (2026-09-16):

- `converters/databricks` (fixtures)
- `converters/orionbelt` (fixtures)
- `converters/ontology` (fixtures, plus the root `examples/flights.yaml` it vendors a copy of)
- `converters/gooddata` (fixtures)
- `converters/omni` (fixtures)
- `converters/honeydew` (source + tests)
- `converters/snowflake` (source)
- `converters/salesforce` (Java/Maven — source + test resources)
- `converters/polaris` (Java/Maven — source + test resources)

## Suggested approach

Same pattern as #407 and #417: read/write the flat document shape (`name`/`datasets`/... at the root, no `semantic_model:` wrapper, no root-level `dialects`/`vendors`), update fixtures, and regenerate any snapshot tests. The Python converters can mostly follow the Sigma/NVIDIA GSF/dbt/wisdom diff in #417 as a template; the Java converters (`salesforce`, `polaris`) will need their own look since they don't share that tooling.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

eisber:
Thanks for cataloguing these. I mapped the affected list against the open work before starting parallel branches:

- #396 covers all nine converters listed here: Databricks, Orionbelt, Ontology, GoodData, Omni, Honeydew, Snowflake, Salesforce, and Polaris. It also covers Sigma, NVIDIA, dbt, and Wisdom.
[...]

PR #420, opened 2026-09-18
fix(honeydew): read and write flat semantic model documents

baruchoxman:
## Summary

Part of #418, for `converters/honeydew`. #383 moved the Ossie document schema to a single semantic model at the document root, and #397 dropped root-level `dialects`/`vendors`. This converter still required the `semantic_model: [...]` wrapper, so it could not read any current Ossie document — including `examples/tpcds_semantic_model.yaml`, which its own round-trip test loads. Its CI has not re-triggered since #383, so the break has been latent: `test_ossie_roundtrip_tpcds_example` is red on `main` today.

## What changed

- **Ossie → Honeydew**: read the model from the document root; removed the wrapper handling and the multi-model warning (a document now holds exactly one model). A document that still carries the legacy wrapper gets an explicit error telling the author to move the model's properties up to the root, rather than a confusing "missing name".
- **Honeydew → Ossie**: emit `version` plus the model's properties at the root, with no wrapper.
- **Dropped the `vendors` round-trip**: root-level `vendors` no longer exists in the schema, and the root is now `additionalProperties: false`, so emitting it made the output invalid. This also removes the `extra_vendors` plumbing through the `ossie` metadata section and the `test_vendors_roundtrip` case that covered it.
- **README**: the mapping tables now name `name` at the document root instead of `semantic_model.name`.
- **Tests**: helpers build and assert the flat shape; added a case asserting the legacy wrapper is rejected with the migration message.

## Verification

The TPC-DS round-trip now works end to end, and its output validates against the official schema — the converter's first output that the validator has ever accepted since #383:

```
$ uv run ossie-honeydew ossie-to-honeydew -i examples/tpcds_semantic_model.yaml -o /tmp/hd_ws
Wrote 20 file(s) to /tmp/hd_ws
$ uv run ossie-honeydew honeydew-to-ossie -i /tmp/hd_ws -o /tmp/roundtrip.yaml
$ python3 validation/validate.py /tmp/roundtrip.yaml
Validation PASSED: roundtrip.yaml
```

## Test plan

- [x] `cd converters/honeydew && uv sync && uv run pytest` — 126 passed (128 on `main`, of which 1 failed; net −3 `test_vendors_roundtrip` params, +1 legacy-wrapper case)
- [x] `test_ossie_roundtrip_tpcds_example` passes, having been red on `main`
- [x] `python3 validation/validate.py` accepts the round-tripped TPC-DS document


baruchoxman:
@jbonofre would appreciate you taking a look

jbonofre:
@baruchoxman absolutely. Thanks!

PR #428, opened 2026-09-18
fix(databricks): match cascade-drop references case-insensitively

### LINKED ISSUE #422: databricks: case-sensitive cascade-drop leaves metric views referencing dropped fields
### Summary
When a field is dropped (e.g. no DATABRICKS dialect), cascade-drop fails to detect a metric that references it in a different case. Databricks SQL identifiers are case-insensitive, but the cascade-drop regexes are case-sensitive, so a metric like `COUNT(DISTINCT REGION_NAME)` survives referencing a dropped `region_name`.

### Root cause
`converters/databricks/src/ossie_databricks/ossie_to_metric_view.py` (`_references_dropped`):
```python
if re.search(r"measure\(\s*" + re.escape(m) + r"\s*\)", expr):          # no re.IGNORECASE
    return m
if d != self_name and re.search(r"(?<![\w.])" + re.escape(d) + r"(?![\w.])", expr):  # no re.IGNORECASE
    return d
```
The converter normalizes identifier case elsewhere (L99-101, 158-164, 545-549) but not here.

### Repro (real converter, HEAD ab8fc1a — EXECUTED)
Model: `region_name` (T_SQL-only → dropped) + metric `region_count = COUNT(DISTINCT REGION_NAME)` (DATABRICKS). Actual output:
```yaml
measures:
- name: region_count
  expr: COUNT(DISTINCT REGION_NAME)   # dangling reference to a dropped field
```
Warning emitted: `[field 'region_name'] no DATABRICKS/ANSI_SQL dialect; dropping field`. The metric should have been cascade-dropped.

### Suggested fix
Add `re.IGNORECASE` to both patterns in `_references_dropped`.

christianeu-db:
## Summary

When a field is dropped (e.g. no DATABRICKS/ANSI_SQL dialect), the Databricks
converter cascade-drops any metric/dimension that references it, so the output
never carries a dangling reference. The reference match in `_references_dropped`
was case-sensitive, but Databricks SQL identifiers are case-insensitive. A metric
like `COUNT(DISTINCT REGION_NAME)` therefore failed to match a dropped
`region_name` and survived as a dangling reference in the emitted Metric View.

The fix adds `re.IGNORECASE` to both patterns in `_references_dropped` (the
`measure(<name>)` pattern and the bare-dimension token pattern). The word-boundary
guards are unchanged, so qualified `alias.col` references are still not
over-dropped.

## Related Issues

Fixes #422

### Converters

### Tests

New regression test `test_cascade_drop_matches_dropped_name_case_insensitively`
reproduces the reported case (a `T_SQL`-only `region_name` dropped, plus a
DATABRICKS metric `region_count = COUNT(DISTINCT REGION_NAME)`). It fails without
the fix (the measure survives) and passes with it. The full
`converters/databricks` suite passes (96 tests).

### Compliance
- [x] ASF license headers are present on all new source files (no new files added)

This pull request and its description were written by Isaac.


christianeu-db:
fyi @Haoranli503, this is the the python counterpart to your Java fix here: https://github.com/Haoranli503/ossie_databricks/commit/a991e686eea8b462fb6c98d49b2b1f456164e69c

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
```diff
@@ -590,17 +590,19 @@ def _references_dropped(expr, self_name, dropped_dims, dropped_measures):
...
+        if re.search(r"measure\(\s*" + re.escape(m) + r"\s*\)", expr, re.IGNORECASE):
             return m
     for d in dropped_dims:
         # Match only a bare, unqualified token: the negative look-behind/ahead for a
         # word char or `.` excludes both substrings of a larger identifier and
         # qualified paths (`alias.name` / `name.col`), so a join alias or joined
         # column sharing a dropped name is not falsely cascade-dropped.
         if d != self_name and re.search(
-                r"(?<![\w.])" + re.escape(d) + r"(?![\w.])", expr):
+                r"(?<![\w.])" + re.escape(d) + r"(?![\w.])", expr, re.IGNORECASE):
```
jbonofre: Making this match case insensitive means a dropped field bare name now collides with any surviving expression that happens to contain the same token in a different case, including common SQL keywords/function names (`COUNT`, `SUM`, `MIN`, `MAX`, `AVG`, `DATE`, ...).

Since these are pretty common as both column names and SQL functions, this looks like a regression to me more than a narrow fix.
Maybe worth to tightening the match (e.g. exclude known SQL keywords, or restrict to idenfiier-shaped tokens that are not immediately followed by ().

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
```diff
@@ -590,17 +590,19 @@ def _references_dropped(expr, self_name, dropped_dims, dropped_measures):
...
     for m in dropped_measures:
-        if re.search(r"measure\(\s*" + re.escape(m) + r"\s*\)", expr):
+        if re.search(r"measure\(\s*" + re.escape(m) + r"\s*\)", expr, re.IGNORECASE):
             return m
     for d in dropped_dims:
         # Match only a bare, unqualified token: the negative look-behind/ahead for a
         # word char or `.` excludes both substrings of a larger identifier and
         # qualified paths (`alias.name` / `name.col`), so a join alias or joined
         # column sharing a dropped name is not falsely cascade-dropped.
         if d != self_name and re.search(
```
jbonofre: This self reference guard is still case-sensitive here, but the regex it protects is case-insensitive (see below): so it no longer reliably excludes self-references when names differ only in case.

Should we `d.lower() != self_name.lower()` (or compare against a case-folded self_name) to match the new case-sensitive semantics?

jbonofre (changes requested):

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
christianeu-db: Agreed - the guard should be case-folded to match the regex. Changed to d.lower() != self_name.lower(), and mirrored the same case-folded self-guard. Added a regression test with a dropped REGION coexisting with a surviving region self-reference.

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
christianeu-db: Good catch - you're right that matching bare tokens case-insensitively widens a collision. Digging in, it's actually a pre-existing issue that IGNORECASE broadened rather than a brand-new one: even before this PR a dropped dimension named exactly COUNT already cascade-dropped a surviving COUNT(...) measure, because the bare-token match never distinguished a column reference from a function call. Case-insensitivity just extends that to case variants.

As shown in the valid snippet of DuckDB SQL below, some vendors support mixed-case keywords:

``` sql
WITH q as (select 1 as SUM)
SELECT sUM(SUM) from q
```

I went with the (?!\s*\() exclusion rather than a SQL-keyword blocklist deliberately: a blocklist would re-open #422 for any identifier that collided with a keyword; a dimension literally named date/count would then no longer cascade-drop and would survive as a dangling reference.

One caveat a bare keyword that is not a function call (a type in CAST(x AS DATE), a unit in EXTRACT(YEAR FROM d), CASE ... END, DISTINCT), or text inside a string literal, can still be over-dropped. Fully resolving column-vs-keyword needs a full parse, which is the ambiguity the function documents; I kept this fix to the function-call case you flagged but opted not to extend the scope of the PR, especially as the python interpreter will be sunset in favor of the Java interpreter, hopefully soon.

christianeu-db:
@jbonofre thank you kindly for the review. I've addressed the comments and added a few regression tests. I've flagged some cases that likely require a fully parser but wanted to hold off for a separate issue / PR to avoid scope creep.

cc @Haoranli503 for the Java converters

jbonofre:
@christianeu-db thank you sooooo much for your updates! I'm doing a new pass 😄

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
```diff
@@ -587,20 +587,32 @@ def _references_dropped(expr, self_name, dropped_dims, dropped_measures):
...
-        # column sharing a dropped name is not falsely cascade-dropped.
-        if d != self_name and re.search(
-                r"(?<![\w.])" + re.escape(d) + r"(?![\w.])", expr):
+        # Match only a bare, unqualified token that is not a function call: the
+        # negative look-behind/ahead for a word char or `.` excludes substrings of a
+        # larger identifier and qualified paths (`alias.name` / `name.col`), and the
+        # trailing `(?!\s*\()` excludes `NAME(...)` calls, so a dropped dim named e.g.
+        # `count` does not falsely match a surviving `COUNT(...)`.
+        if d.lower() != self_name.lower() and re.search(
+                r"(?<![\w.])" + re.escape(d) + r"(?![\w.])(?!\s*\()", expr, re.IGNORECASE):
```
jbonofre, line 620: I believe this flag makes the converter silently drop valid dimensions.

Let's take this:

```yaml
fields:
    - name: active # DATABRICKS
    - name: status # DATABRICKS: CASE WHEN active THEN 'a' ELSE 'b' END
    - name: end.    # T_SQL only -> dropped
```

I think `end` is dropped. The `END` of `CASE ... END` now matches the dropped `end`.  The new `(?!\s*\()` guard doesn't help: these keywords aren't calls. Same failure for a dropped `date` against `CAST(ts AS DATE)` and a dropped `year` against `EXTRACT(YEAR FROM ts)`.
I believe field names like a non-call SQL keyword (`end`, `date`, `year`, `interval`, `rows`, ...) are dropped.

Also, I believe the same happens for fields with name inside a string literal in another case (like `STATUS`).

I'm sorry I should have seen that during the first reivew.

The case insensitive matching is right here, but it only pays for itself once the matcher strips quoted literals and skips SQL reserved words first.

I suggest replacing the regex with a single tokenizing pass: strip literals, tokenize `[A-Za-z]\w*`, skip tokens adjacent to `.` or followed by `(`, filter reserved word set, lowercase into a set. It would address the dropped fields plus the `measure(` gaps (see my next comment on the same line 😄 ).

jbonofre, line 620: Oh also, considering this:

```yaml
datasets:
  - name: d
    fields:
      - name: id      # DATABRICKS
      - name: Region  # T_SQL only -> dropped
metrics:
  - name: region      # DATABRICKS: SUM(Region)
```

I believe the case-folding the self-guard suppresses a genuine cascade-drop, re-introducing the dangling reference this PR is meant to fix.

`'Region'.lower() == 'region'`, so the guard treats a reference to the dropped fiels as a self-reference and keeps the measure. Note this reaches the cascade check at all only because dropped field name never enter `seen_dims`, so the case-insensitive name collision that would normally reject this model never fires.

`test_cascade_drop_does_not_drop_self_reference_differing_only_in_case` does pass, but the case it pins (a survivor whose expression is its own bare name) is narrower than the guard it justified, the name guard also swallows references to a different, dropped column.

I think two things worth separating here:
* The `.lower()` on the self-guard isn't part of the case-insensitive fix and should come off.
* The guard compares against `self_name` for both loops, so a measure named like a dropped dimension is exempted from a check that has nothing to do with it (field `region` dropped + metric `region` = `COUNT(DISTINCT region)` emits a dangling ref). Passing a `kind` through from `_cascade_drop` so the self guard only applies within the matching dropped set would fix that.

jbonofre:
@christianeu-db I have a couple of comments/questions for you 😄

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
christianeu-db: Good catch, the cross-kind exemption is a real bug. Fixed: kind now flows from _cascade_drop into the guard, so it only exempts a field from referencing itself within its own kind. Your region measure now reads SUM(Region) as a reference to the dropped Region dimension and cascade-drops instead of slipping through. Pinned with test_cascade_drop_measure_named_like_dropped_dimension_is_dropped.
I kept the .lower() though. I think the case-folding was a red herring and kind was the real culprit. Once the guard is kind-scoped, .lower() only ever suppresses a genuine same-kind self-reference: a dimension referencing its own name in another case, which is what test_cascade_drop_does_not_drop_self_reference_differing_only_in_case covers (dropped REGION alongside a surviving region whose expr is region). Drop the .lower() and that survivor cascade-drops and the test goes red. So: kind-scoped guard, .lower() stays.

converters/databricks/src/ossie_databricks/ossie_to_metric_view.py
christianeu-db: The measure( gap, a measure named like a dropped dimension, is actually the same bug as your other comment, and the kind-scoped guard already closes it. No tokenizer needed there.
The rest, non-call keywords (end, date, year, interval, rows) and same-name tokens inside string literals, I looked at the tokenizing pass and don't think it belongs in this PR. It's a pre-existing quirk of the bare-token match, not something the case-insensitive change introduced (it over-drops case-sensitively today too), and when it fires it over-drops: a valid-but-incomplete view with a warning, not the dangling ref #422 is about.
The catch is that a reserved-word set re-opens #422. A dimension literally named date, end, or year gets filtered as a keyword and then survives as a dangling reference. I checked CAST(date AS STRING) with a dropped date, and the blocklist keeps it. So it trades the benign failure for the exact malignant one we're fixing.
Doing it right means grammar context, DATE the type versus date the column, which is a real parse and a much bigger lift than this fix. I'd rather capture the residual in a follow-up issue than bolt a parser on here since we're deprecating this impl in favor of the java one

christianeu-db:
> @christianeu-db thank you sooooo much for your updates! I'm doing a new pass 😄

Thanks for the second pass. Round 3 is up.

Cross-kind self-guard bug: fixed. kind now flows through, so a measure named like a dropped dimension cascade-drops instead of slipping past. Regression test added.

.lower(): kept it, reasoning inline. Short version: kind-scoping was the real culprit, and removing .lower() regresses the case-variant self-reference test.

Non-call keywords and string literals: held out of scope, reasoning inline. The behavior is pre-existing and benign, and a reserved-word set actually re-opens #422.

One question on scope. This started as a one-line case-insensitivity fix for a real dangling-reference bug, and with the kind-scope fix it now closes that bug cleanly. The remaining over-drop cases need a proper parser to resolve, which is a real chunk of work on an interpreter you've said is heading for deprecation in favor of the Java one. I'm not convinced the ROI is there. Do you want to land this as-is and track the residual in a follow-up for Java? I'm also happy to drop this PR entirely in favor of accelerating the Java impl

PR #429, opened 2026-09-18
Fix SDK version ordering and OrionBelt fallback validation

flyrain:
## Summary

SDK-generated documents currently place `version` after the inherited model fields. Serialize it first across the SDK's dictionary, JSON, and YAML output, matching the spec examples and other converters. This changes presentation only; field selection, aliases, and the generated serialization schema are preserved.

OrionBelt's `validate_ossie` can report a legacy wrapped document as valid when the schema file or `jsonschema` package is unavailable. Record explicit semantic errors for legacy `semantic_model` wrappers and non-object roots so those inputs remain invalid when schema validation is skipped.

Follow-up to #396

## Checklist

- [x] Added regression tests for both reported issues.
- [x] Updated SDK serialization documentation and affected snapshots.
- [x] Preserved ASF license headers.
- [x] No specification changes or new dependencies.


flyrain:
Hi @khush-bhatia, here is a followup for your offline feedback. PTAL.

converters/orionbelt/src/ossie_orionbelt/validation.py
```diff
@@ -207,13 +207,19 @@ def validate_ossie(ossie_dict: dict[str, Any], schema_path: Path | None = None)
...
-    # JSON Schema validation above already reports structural errors, so guard
-    # every level here rather than raising on malformed input.
+    # Guard every level rather than raising on malformed input, even when
+    # JSON Schema validation is unavailable.
     def _as_dict_list(value: Any) -> list[dict[str, Any]]:
         return [item for item in value if isinstance(item, dict)] if isinstance(value, list) else []
 
-    # Legacy wrappers are schema errors, not model contents to traverse.
-    if not isinstance(ossie_dict, dict) or "semantic_model" in ossie_dict:
+    if not isinstance(ossie_dict, dict):
```
jbonofre, line 230: This fallback only catches two specific malformations (`ossie_dict` not a dict, or a `semantic_model` wrapper).
When `jsonschema` is unavailable (or `schema_path` points at a missing schema), every other structurally invalid document silently comes back as `valid=True` with empty `schema_errors`/`semantic_errors`. For instance `datasets` as a string instead of a list, `datasets` missing entirely, `name` missing, or a dataset dict missing its required `source` field.

Since `_as_dict_list()` already silently drops non-dict items instead of erroring, and the downstream unique-name/reference checks only ever see whatever survives that filter, none of these malformations get flagged anywhere in this fallback.

Given this function docstring says it mirrors `validate.py` three layer validation, could we add guards at least the missing/malformed `datasets` and missing `name` cases here too? Or explicitly document that JSON-Schema-unavailable mode is a reduced-coverage fallback and only guarantees the two checks above?

I believe it's currently confusing because it reads as full validation but it isn't when `jsonschema` is absent.

jbonofre (changes requested):

converters/orionbelt/src/ossie_orionbelt/validation.py
flyrain: The issue wasn't introduced by this PR. Fixed it anyway in the new commit. PTAL, @jbonofre !

khush-bhatia (approved):

